PR #388, opened 2015-07-17
Separate base data from scenario data in ScenarioCalculationEnvironment

cjkent:
This PR implements a refactoring of scenarios in preparation for local scenarios.

`ScenarioCalculationEnvironment` now embeds an instance of `CalculationEnvironment` containing the base data, i.e. any data that is the same in all scenarios. When data is requested the scenario data is checked first and then the request is delegated to the base data. 

This separation of base and scenario data is a closer fit to the mental model of most users. It also means the market data containers will be much smaller for environments where only a subset of the data is perturbed in each scenario.

The downside of the changes are additional complexity when building market data but this is unavoidable.

### DIFF
--- a/modules/engine/src/main/java/com/opengamma/strata/engine/marketdata/ScenarioCalculationEnvironment.java
+++ b/modules/engine/src/main/java/com/opengamma/strata/engine/marketdata/ScenarioCalculationEnvironment.java
@@ -6,6 +6,7 @@
 package com.opengamma.strata.engine.marketdata;
 
 import java.time.LocalDate;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
@@ -56,6 +57,10 @@ import com.opengamma.strata.engine.calculations.NoMatchingRuleId;
 @BeanDefinition(builderScope = "private")
 public class ScenarioCalculationEnvironment implements ImmutableBean {
 
+  /** The market data values which are the same in every scenario. */
+  @PropertyDefinition(validate = "notNull")
+  private final CalculationEnvironment baseData;
+
   /** The number of scenarios. */
   @PropertyDefinition(validate = "ArgChecker.notNegativeOrZero")
   private final int scenarioCount;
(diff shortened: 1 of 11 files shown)


jonathansenior:
Be careful with the definition of "base data" in the description of this PR. It wouldn't normally mean "any data that is the same in all scenarios"; it would normally mean the set of market data corresponding to the base scenario. From the users' perspective, there may not be a scenario that can be identified as the "base" scenario.

modules/engine/src/main/java/com/opengamma/strata/engine/marketdata/ScenarioCalculationEnvironment.java
```diff
@@ -192,10 +174,15 @@ private void validate() {
...
 
-    if (values.isEmpty()) {
-      throw new IllegalArgumentException("No values available for market data ID " + id);
+    if (!values.isEmpty()) {
+      return (List<T>) values;
     }
-    return (List<T>) values;
+    T baseValue = baseData.getValue(id);
+
+    if (baseValue != null) {
```
jodastephen: My reading of the code is that `getValue()` does not return null.

modules/engine/src/main/java/com/opengamma/strata/engine/marketdata/ScenarioCalculationEnvironment.java
```diff
@@ -56,6 +57,10 @@
 @BeanDefinition(builderScope = "private")
 public class ScenarioCalculationEnvironment implements ImmutableBean {
 
+  /** The market data values which are the same in every scenario. */
+  @PropertyDefinition(validate = "notNull")
+  private final CalculationEnvironment baseData;
```
jodastephen: How about `commonData` or `sharedData`.

cjkent: Shared data is what I was thinking too.

cjkent:
Any other comments? I've addressed everything so far.

PR #395, opened 2015-07-20
Ensure command line tool works on Windows

jodastephen:
Resources always use forward-slash, not OS file separator
Change report tool to have a directory within zip for easy extracting

### DIFF
--- a/examples/src/main/java/com/opengamma/strata/examples/marketdata/MarketDataBuilder.java
+++ b/examples/src/main/java/com/opengamma/strata/examples/marketdata/MarketDataBuilder.java
@@ -122,29 +122,27 @@ public abstract class MarketDataBuilder {
    * @return the market data builder
    */
   public static MarketDataBuilder ofResource(String resourceRoot, ClassLoader classLoader) {
-    String qualifiedResourceRoot = resourceRoot.startsWith(File.separator) ? resourceRoot.substring(1) : resourceRoot;
-    if (!qualifiedResourceRoot.endsWith(File.separator)) {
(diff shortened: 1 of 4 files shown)


examples/src/main/java/com/opengamma/strata/examples/marketdata/MarketDataBuilder.java
```diff
@@ -122,29 +122,27 @@ public static MarketDataBuilder ofResource(String resourceRoot) {
...
   public static MarketDataBuilder ofResource(String resourceRoot, ClassLoader classLoader) {
-    String qualifiedResourceRoot = resourceRoot.startsWith(File.separator) ? resourceRoot.substring(1) : resourceRoot;
-    if (!qualifiedResourceRoot.endsWith(File.separator)) {
-      qualifiedResourceRoot += File.separator;
-    }
-    URL url = classLoader.getResource(qualifiedResourceRoot);
+    // classpath resources are forward-slash separated
+    String qualifiedRoot = resourceRoot;
+    qualifiedRoot = qualifiedRoot.startsWith("/") ? qualifiedRoot.substring(1) : qualifiedRoot;
+    qualifiedRoot = qualifiedRoot.startsWith("\\") ? qualifiedRoot.substring(1) : qualifiedRoot;
```
cjkent: This converts a prefix of `\\` to `\`. Is that what we want?

jodastephen: `\\` is just `\`, the extra slash is escape character.

PR #404, opened 2015-07-22
Add options to speed up travis

jodastephen:
Replace mvn install with mvn validate
Cache repository to avoid downloading the internet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: java
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch, this ought to speed up build times.

Whilst the above appears to work, it's not ideal, but probably a feature of the build system/environment opposed to anything we can alter whilst replicating what a user would typically do (have checked common `mvn` options and `-D` defines etc.).

Build passes and cache seems to be correctly maintained but an eye ought to be kept on it to make sure it isn't accidentally bloated or corrupted.

Thanks again. Merging.

PR #417, opened 2015-08-11
Ibor Future Option: Minor refactor of volatility data objects to use expiry time.

marc-opengamma:
Current version uses only expiry date. Also some "null" are passed as expiry time and zone in some places.

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivity.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivity.java
@@ -49,7 +49,7 @@ public final class IborFutureOptionSensitivity
    * The expiration date of the option.
    */
   @PropertyDefinition(validate = "notNull")
-  private final ZonedDateTime expiryDate;
+  private final ZonedDateTime expiry;
   /**
    * The underlying future last trading or fixing date.
    */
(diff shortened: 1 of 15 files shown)


modules/market/src/test/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivityTest.java
```diff
@@ -56,11 +60,11 @@ public void test_of_withCurrency() {
 
   //-------------------------------------------------------------------------
   public void test_withCurrency() {
-    IborFutureOptionSensitivity base = IborFutureOptionSensitivity.of(GBP_LIBOR_3M, date(2015, 8, 27),
+    IborFutureOptionSensitivity base = IborFutureOptionSensitivity.of(GBP_LIBOR_3M, DateUtils.getUTCDate(2015, 8, 27),
```
jodastephen: Please add a method `dateUtc()` to `TestHelper` to avoid using `DateUtils`.

modules/market/src/main/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivity.java
```diff
@@ -48,7 +49,7 @@
    * The expiration date of the option.
    */
   @PropertyDefinition(validate = "notNull")
-  private final LocalDate expiryDate;
+  private final ZonedDateTime expiryDate;
```
jodastephen: Looks like this should be renamed to `expiry`, as it is no longer just a date.

modules/market/src/main/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivity.java
```diff
@@ -115,7 +116,7 @@ public static IborFutureOptionSensitivity of(
    */
   public static IborFutureOptionSensitivity of(
       IborIndex index,
-      LocalDate expiryDate,
+      ZonedDateTime expiryDate,
```
yukiiwashita: use "expiry" to be agree with the field name in the class.

modules/market/src/main/java/com/opengamma/strata/market/sensitivity/IborFutureOptionSensitivity.java
```diff
@@ -91,7 +92,7 @@
...
```
yukiiwashita: use "expiry" to be agree with the field name in the class.

PR #424, opened 2015-08-14
Add XmlFile support

jodastephen:
Parse simple/common XML files into a DOM-like structure
Base on StAX, but hidden from regular use
Avoid exposing comments, PIs and mixed content
Enhance Guavate to check a stream contains a single element

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -8,6 +8,7 @@ package com.opengamma.strata.collect;
 import java.util.Comparator;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.BinaryOperator;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
@@ -87,6 +88,29 @@ public final class Guavate {
    */
   public static <R> Predicate<R> not(Predicate<R> predicate) {
     return predicate.negate();
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Reducer used in a stream to ensure there is no more than one matching element.
+   * <p>
+   * This method returns an operator that can be used with {@link Stream#reduce(BinaryOperator)}
+   * that returns either zero or one elements from the stream. Unlike {@link Stream#findFirst()}
+   * or {@link Stream#findAny()}, this approach ensures an exception is thrown if there
+   * is more than one element in the stream.
+   * <p>
+   * This would be used as follows:
+   * <pre>
+   *   stream.filter(...).reduce(Guavate.ensureOnlyOne()).get();
+   * </pre>
+   * 
+   * @param <T>  the type of element in the stream
+   * @return the operator
+   */
+  public static <T> BinaryOperator<T> ensureOnlyOne() {
+    return (a, b) -> {
+      throw new IllegalArgumentException(Messages.format("Duplicates found: {} and {}", a, b));
+    };
   }
 
   //-------------------------------------------------------------------------
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -109,7 +109,8 @@ public final class Guavate {
    */
   public static <T> BinaryOperator<T> ensureOnlyOne() {
     return (a, b) -> {
-      throw new IllegalArgumentException(Messages.format("Duplicates found: {} and {}", a, b));
+      throw new IllegalArgumentException(Messages.format(
+          "Multiple values found where only one was expected: {} and {}", a, b));
     };
   }
 
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
@@ -358,6 +358,7 @@ public final class XmlElement
     result = prime * result + name.hashCode();
     result = prime * result + content.hashCode();
     result = prime * result + attributes.hashCode();
+    result = prime * result + children.size();
     return result;
   }
 
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
@@ -1,0 +1,391 @@
+/**
+ * Copyright (C) 2015 - present by OpenGamma Inc. and the OpenGamma group of companies
+ * 
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.collect.io;
+
+import static com.opengamma.strata.collect.Guavate.ensureOnlyOne;
+import static com.opengamma.strata.collect.Guavate.toImmutableList;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import javax.xml.namespace.QName;
+
+import org.joda.beans.ImmutableBean;
+import org.joda.beans.MetaBean;
+import org.joda.beans.Property;
+import org.joda.beans.PropertyDefinition;
+import org.joda.beans.impl.light.LightMetaBean;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.opengamma.strata.collect.ArgChecker;
+import com.opengamma.strata.collect.Messages;
+
+/**
+ * A single element in the tree structure of XML.
+ * <p>
+ * This class is a minimal, lightweight representation of an element in the XML tree.
+ * The element has a name, attributes, and either content or children.
+ * <p>
+ * Note that this representation does not express all XML features.
+ * No support is provided for processing instructions, comments or mixed content.
+ * In addition, it is not possible to determine the difference between empty content and no children.
+ * <p>
+ * There is no explicit support for namespaces.
+ * When creating instances, the caller may choose to use a convention to represent namespaces.
+ * For example, element and attribute names may use prefixes or the standard {@link QName} format.
+ */
+public final class XmlElement
+    implements ImmutableBean {
+
+  /**
+   * The meta-bean.
+   * This is a manually coded bean.
+   */
+  private static MetaBean META_BEAN = LightMetaBean.of(XmlElement.class);
+
+  /**
+   * The element name.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final String name;
+  /**
+   * The attributes.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final ImmutableMap<String, String> attributes;
+  /**
+   * The element content.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final String content;
+  /**
+   * The child nodes.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final ImmutableList<XmlElement> children;
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance with content and no attributes.
+   * <p>
+   * Returns an element representing XML with content, but no children.
+   * 
+   * @param name  the element name, not empty
+   * @param content  the content, empty if the element has no content
+   * @return the element
+   */
+  public static XmlElement ofContent(String name, String content) {
+    return ofContent(name, ImmutableMap.of(), content);
+  }
+
+  /**
+   * Obtains an instance with content and attributes.
+   * <p>
+   * Returns an element representing XML with content and attributes but no children.
+   * 
+   * @param name  the element name, not empty
+   * @param attributes  the attributes, empty if the element has no attributes
+   * @param content  the content, empty if the element has no content
+   * @return the element
+   */
+  public static XmlElement ofContent(String name, Map<String, String> attributes, String content) {
+    return new XmlElement(name, ImmutableMap.copyOf(attributes), content, ImmutableList.of());
+  }
+
+  /**
+   * Obtains an instance with children and no attributes.
+   * <p>
+   * Returns an element representing XML with children, but no content.
+   * 
+   * @param name  the element name, not empty
+   * @param children  the children, empty if the element has no children
+   * @return the element
+   */
+  public static XmlElement ofChildren(String name, List<XmlElement> children) {
+    return ofChildren(name, ImmutableMap.of(), children);
+  }
+
+  /**
+   * Obtains an instance with children and attributes.
+   * <p>
+   * Returns an element representing XML with children and attributes, but no content.
+   * 
+   * @param name  the element name, not empty
+   * @param attributes  the attributes, empty if the element has no attributes
+   * @param children  the children, empty if the element has no children
+   * @return the element
+   */
+  public static XmlElement ofChildren(String name, Map<String, String> attributes, List<XmlElement> children) {
+    return new XmlElement(name, ImmutableMap.copyOf(attributes), "", ImmutableList.copyOf(children));
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Creates an instance.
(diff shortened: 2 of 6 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -91,6 +92,29 @@ private Guavate() {
...
+   * <pre>
+   *   stream.filter(...).reduce(Guavate.ensureOnlyOne()).get();
+   * </pre>
+   * 
+   * @param <T>  the type of element in the stream
+   * @return the operator
+   */
+  public static <T> BinaryOperator<T> ensureOnlyOne() {
+    return (a, b) -> {
+      throw new IllegalArgumentException(Messages.format("Duplicates found: {} and {}", a, b));
```
cjkent: The message is a bit misleading. `a` and `b` aren't really duplicates. Something like "Multiple values found:"?

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
```diff
@@ -0,0 +1,391 @@
...
+ * <p>
+ * Note that this representation does not express all XML features.
+ * No support is provided for processing instructions, comments or mixed content.
+ * In addition, it is not possible to determine the difference between empty content and no children.
+ * <p>
+ * There is no explicit support for namespaces.
+ * When creating instances, the caller may choose to use a convention to represent namespaces.
+ * For example, element and attribute names may use prefixes or the standard {@link QName} format.
+ */
+public final class XmlElement
```
cjkent: What's the advantage of defining our own proprietary XML model instead of using the DOM from the JDK?

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
```diff
@@ -0,0 +1,391 @@
...
+  /**
+   * Obtains an instance with children and no attributes.
+   * <p>
+   * Returns an element representing XML with children, but no content.
+   * 
+   * @param name  the element name, not empty
+   * @param children  the children, empty if the element has no children
+   * @return the element
+   */
+  public static XmlElement ofChildren(String name, List<XmlElement> children) {
```
cjkent: Is it worth having overloads for this method and the one below where the children are varags? That would provide a cleaner way to define an XML structure in code.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
```diff
@@ -0,0 +1,391 @@
...
+   * Gets the first matching child element with the specified name.
+   * <p>
+   * This returns the first child element with the specified name.
+   * An exception is thrown if the child does not exist.
+   * 
+   * @param childName  the name to match
+   * @return the child matching the name
+   * @throws IllegalArgumentException if there is no match
+   */
+  public XmlElement getChildFirst(String childName) {
```
cjkent: I'm not keen on this method name. I presume it's meant to be consistent with the other `getChild*` methods. But I think it would be better if it were called `getFirstChild`.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
```diff
@@ -0,0 +1,391 @@
...
+          attributes.equals(other.attributes) &&
+          children.equals(other.children);
+    }
+    return false;
+  }
+
+  /**
+   * Returns a suitable hash code.
+   * <p>
+   * Note that the has code excludes the children to ensure performance is reasonable in large trees.
```
cjkent: That means `hashCode` and `equals` aren't necessarily consistent. That could lead to some nasty performance characteristics if anyone ever uses `XmlElement` instances as map keys. 

If two elements have the same name, attributes and content they will be in the same bucket and they will be checked for equality which traverses the entire tree.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+   * @param refAttrName  the attribute name that should be parsed as a reference
+   * @return the parsed file
+   * @throws UncheckedIOException if an IO exception occurs
+   * @throws IllegalArgumentException if the file cannot be parsed
+   */
+  public static XmlFile of(ByteSource source, String refAttrName) {
+    ArgChecker.notNull(source, "source");
+    return Unchecked.wrap(() -> {
+      try (InputStream in = source.openBufferedStream()) {
+        XMLStreamReader xmlReader = xmlInputFactory().createXMLStreamReader(in);
```
cjkent: Should use the version of `createXMLStreamReader` that explicitly specifies the encoding.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+    return of(source, "");
+  }
+
+  /**
+   * Parses the specified source as an XML file to an in-memory DOM-like structure.
+   * <p>
+   * This parses the specified byte source expecting an XML file format.
+   * The resulting instance can be queried for the root element.
+   * <p>
+   * This supports capturing attribute references, such as an id/href pair.
```
cjkent: I'm not sure from the Javadoc what you mean by "capturing attribute references". More explanation or an example would be helpful.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+      String content = "";
+      int event = reader.next();
+      while (event != XMLStreamConstants.END_ELEMENT) {
+        switch (event) {
+          case XMLStreamConstants.START_ELEMENT:
+            childBuilder.add(parse(reader, refAttr, refs));
+            break;
+          case XMLStreamConstants.CHARACTERS:
+          case XMLStreamConstants.CDATA:
+            content += reader.getText();
```
cjkent: Use a `StringBuilder` here?

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  // find the start element and parses the name
+  private static String parseElementName(XMLStreamReader reader) throws XMLStreamException {
+    int event = reader.getEventType();
+    while (event != XMLStreamConstants.START_ELEMENT) {
+      event = reader.next();
+    }
+    String prefix = reader.getPrefix();
+    if (prefix.length() == 0) {
```
cjkent: `getPrefix` returns `null` if there is no prefix so this will throw an NPE.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+  // parses attributes into a map
+  private static ImmutableMap<String, String> parseAttributes(XMLStreamReader reader) {
+    ImmutableMap<String, String> attrs;
+    int attributeCount = reader.getAttributeCount() + reader.getNamespaceCount();
+    if (attributeCount == 0) {
+      attrs = ImmutableMap.of();
+    } else {
+      ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+      for (int i = 0; i < reader.getAttributeCount(); i++) {
+        String prefix = reader.getAttributePrefix(i);
```
cjkent: `getAttributePrefix` can return `null`.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+    return attrs;
+  }
+
+  //-------------------------------------------------------------------------
+  // creates the XML input factory., recreated each time to avoid JDK-8028111.
+  private static XMLInputFactory xmlInputFactory() {
+    XMLInputFactory factory = XMLInputFactory.newFactory();
+    factory.setProperty(XMLInputFactory.IS_COALESCING, true);
+    factory.setProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES, true);
```
cjkent: Does this make it vulnerable to an [XML bomb](https://en.wikipedia.org/wiki/Billion_laughs)?

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+    return factory;
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Restricted constructor.
+   */
+  private XmlFile(XmlElement root, Map<String, XmlElement> refs) {
+    this.root = ArgChecker.notNull(root, "root");
+    this.refs = ImmutableMap.copyOf(ArgChecker.notNull(refs, "refs"));
```
cjkent: The call to `notNull` is redundant, `ImmutableMap.copyOf` will blow up if `refs` is null.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+  public XmlElement getRoot() {
+    return root;
+  }
+
+  /**
+   * Gets the reference map of id to element.
+   * <p>
+   * This is used to decode references, such as an id/href pair.
+   * <p>
+   * For example, if one part of the XML has {@code <foo id="fooId">}, the map will
```
cjkent: Now I understand what this is all about. It would be useful to copy this example to the other places that talk about mapping references.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -0,0 +1,282 @@
...
+    return false;
+  }
+
+  /**
+   * Returns a suitable hash code for the file.
+   * 
+   * @return the hash code
+   */
+  @Override
+  public int hashCode() {
```
cjkent: `hashCode` and `equals` are inconsistent.

modules/collect/src/test/java/com/opengamma/strata/collect/GuavateTest.java
```diff
@@ -63,6 +64,13 @@ public void test_not_Predicate() {
   }
 
   //-------------------------------------------------------------------------
+  public void test_ensureOnlyOne() {
+    assertEquals(Arrays.asList().stream().reduce(Guavate.ensureOnlyOne()), Optional.empty());
```
cjkent: use `Stream.empty()` instead of `Arrays.asList().stream()`?

modules/collect/src/test/java/com/opengamma/strata/collect/GuavateTest.java
```diff
@@ -63,6 +64,13 @@ public void test_not_Predicate() {
...
+    assertEquals(Arrays.asList("a").stream().reduce(Guavate.ensureOnlyOne()), Optional.of("a"));
```
cjkent: Use `Stream.of("a")` instead of `Arrays.asList("a").stream()`?

modules/collect/src/test/java/com/opengamma/strata/collect/GuavateTest.java
```diff
@@ -63,6 +64,13 @@ public void test_not_Predicate() {
...
+    assertThrowsIllegalArg(() -> Arrays.asList("a", "b").stream().reduce(Guavate.ensureOnlyOne()));
```
cjkent: Use `Stream.of("a", "b")`?

cjkent:
That's everything from me.

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
cjkent: Maybe "Multiple values found where only one value was expected".

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
jodastephen: The JDK DOM API is a direct port of the standard DOM. Its a horrible API, very difficult to create and query. It also supports more than is really necessary. As shown here, one class is all that is really needed.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
jodastephen: Currently, this is only created by the parser, so no need to overcomplicate it.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
jodastephen: Having used the API, I felt they acted as variants of the core theme of getting a child. With the three methods grouped together it helped auto-completing. Having getOptionalChild, getSIngleChild and getFirstChild means you tend to have to know the key words (optional/single/first) earlier.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
jodastephen: I've added the size of the list of children, but I'm reluctant to include the whole tree. I suspect that these won't be used as hash keys anyway, but it would be mean to not support equals/hashCode. While partial hash codes are unusual, they are definitely part of the intent of hash codes.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
cjkent: I realise that, but I think `getChildOptional` is such an awkward name it would be better to be slightly inconsistent. Anyway, auto-completion shouldn't care. IntelliJ does partial matching so `getch` would match `getFirstChild`.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
jodastephen: Any evidence for this? My understanding is that XML documents contain the encoding on the first line and default to UTF-8 if not found. As such, I believe you are not supposed to set the encoding manually.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
jodastephen: Since `XMLStreamReader` has `IS_COALESCING=true` there should only be one call here.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
jodastephen: True, but the a start element event never returns null.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
jodastephen: No, because there is a limit of expansions internally. This is why we have to recreate it each time.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
jodastephen: Based on current use cases, I've removed `getChildFirst` and renamed `getChildSingle` to `getChild`, plus added an optional attribute method to even out the API.

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlElement.java
cjkent: I'm still not comfortable with breaking such a fundamental rule without a concrete reason. It feels like a premature optimisation.

If we don't think these will be used as map keys then what harm does it do to implement `hashCode` properly? Or not implement it at all? Either would be preferable to a surprising and non-standard implementation that isn't justified by any data.

PR #427, opened 2015-08-18
Add Ibor and Overnight market data ids

jodastephen:
Use market data system to build `IborIndexRates` and `OvernightIndexRates`.
Add ID, mapping and function.


modules/function/src/main/java/com/opengamma/strata/function/calculation/rate/swap/AbstractSwapFunction.java
```diff
@@ -97,6 +100,17 @@ public FunctionRequirements requirements(SwapTrade trade) {
         .build();
   }
 
+  // create a suitable key
+  private MarketDataKey<?> indexCurveKey(Index index) {
```
cjkent: Why have this logic here rather than in `MarketDataKeys`?

modules/function/src/main/java/com/opengamma/strata/function/marketdata/curve/IborIndexRatesMarketDataFunction.java
```diff
@@ -0,0 +1,112 @@
...
+          FailureReason.MISSING_DATA,
+          "No curve found: Index: {}, Group: {}, Feed: {}",
+          id.getIndex(),
+          id.getCurveGroupName(),
+          id.getMarketDataFeed());
+    }
+    Curve curve = marketData.getValue(curveId);
+
+    // create Ibor rates
+    return Result.wrap(() -> createIborIndexRates(id.getIndex(), marketData.getValuationDate(), timeSeries, curve));
```
cjkent: Why not just

    return createIborIndexRates(...);

Are there any exceptions thrown in that method that need to be caught by `wrap`? I can't see anything obvious.

jodastephen: This simply follows the established pattern for this kind of function. It doesn't seem like there is any particular harm caused. I suspect that not everything is checked, for example whether the x-value is a year fraction.

Issue #416, opened 2015-08-10
DoubleSurface should provide derivatives with respect to inputs


marc-opengamma:
The "Curve", which the 1D equivalent of the 2D surface, provide a "firstDerivative" method which correspond to the first order derivative of the output with respect to the input, i.e. d y / d x.
A similar method should be available on DoublesSurface. The method with return a double[] with 2 elements: d z / d x and d z / d y. The is required for several sensitivity analysis.

PR #436, opened 2015-08-27
Add surfaces

jodastephen:
Add a representation of a surface based on design for curves.

This uses the old interpolator, which will have to be replaced, but it is better to get this in as is.
Issue #416 will need adding later as well.


modules/market/src/main/java/com/opengamma/strata/market/surface/InterpolatedNodalSurface.java
```diff
@@ -0,0 +1,697 @@
...
+  /**
+   * The underlying data bundle.
+   */
+  private transient final Map<Double, Interpolator1DDataBundle> underlyingDataBundle;  // derived and cached, not a property
+  /**
+   * The underlying interpolator.
+   */
+  // TODO: put back to transient, add SurfaceInterpolator
+  @PropertyDefinition(validate = "notNull")
+  private final GridInterpolator2D interpolator;  // derived and cached, not a property
```
yukiiwashita: Is this derived?

PR #438, opened 2015-08-28
Remove JOOQ

jodastephen:
Library can be replaced by entrySet() streams and two helper methods.

Rename `ObjectDoublePair` to `ObjDoublePair`, retaining old one for Analytics for the moment.


jonathansenior:
Worthwhile change. As we're still on 0.x the deprecated code should be removed once the dependency on Analytics is removed.

PR #451, opened 2015-09-15
Enhance curves and surfaces

jodastephen:
Allow conversion from `Curve` to `NodalCurve` and same for `Surface`
Add `applyPerturbation()` to `Surface`


jonathansenior:
Reviewed post-merge.

PR #457, opened 2015-09-21
Enhance curve config

jodastephen:
Rename extrapolator properties to be consistent with other locations.
Allow curve to be used for multiple discount currencies.
Javadoc and tidy.
Add tests.


jonathansenior:
Reviewed post-merge.

PR #461, opened 2015-09-24
Prepare for curve calibration

jodastephen:
Add new methods that prepare the way for curve calibration.
Rename getXxx() methods that search to findXxx() as a better naming convention.


modules/market/src/main/java/com/opengamma/strata/market/curve/ParRates.java
```diff
@@ -57,6 +59,20 @@ public static ParRates of(Map<ObservableId, Double> rates, CurveMetadata metadat
...
   }
 
+  //-------------------------------------------------------------------------
+  /**
+   * Converts the par rates to a map by key.
+   *
+   * @return the equivalent map of par rate by key
+   */
+  public ImmutableMap<ObservableKey, Double> toRatesByKey() {
+    ImmutableMap.Builder<ObservableKey, Double> ratesByKey = ImmutableMap.builder();
```
cjkent: Could use:

    return rates.entrySet().stream()
        .map(e -> Pairs.of(e.getKey().toObservableKey(), e.getValue())
        .collect(pairsToImmutableMap());

cjkent:
Merging as the only comment wasn't a problem.

PR #474, opened 2015-10-02
Add strata-math, remove og-analytics

jodastephen:
This adds a new project strata-math and removes og-analytics.

The maths code is selected from the og-analytics math code. Some work has been done to minimize the necessary code, but no doubt it could be even slimmer. Some changes have been made to code format, such as removing _ from variable names, but not too much (there is limited value in doing so).

The remainder of the change is to alter existing code to use strata-math instead of og-analytics. This is mostly a package rename. However, a new class `StandardDataSets` was added to match the numbers from previous analytics code.

Given that this is mostly existing well-known code with a relatively high test coverage of 85%, it is not expected that every line will be reviewed in detail.


jonathansenior:
Good stuff :-) This is a big step forward.

PR #478, opened 2015-10-05
Store Jacobian matrix in curve metadata

jodastephen:
Add `CurveCalibrationInfo` and `JacobianCurveCalibrationInfo`. Store in curve metadata to ensure system stays unified.


modules/market/src/main/java/com/opengamma/strata/market/curve/CurveCalibrationInfo.java
```diff
@@ -0,0 +1,31 @@
...
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.market.curve;
+
+/**
+ * Information about curve calibration.
+ * <p>
+ * Implementations of this interface provide information that was obtained during curve calibration.
+ */
+public interface CurveCalibrationInfo {
```
cjkent: What's the thinking behind this interface? It seems a bit...unusual.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveMetadata.java
```diff
@@ -65,6 +65,15 @@
   public abstract Optional<DayCount> getDayCount();
 
   /**
+   * Gets the calibration information, optional.
+   * <p>
+   * When a curve is calibrated, useful information is obtained, which can be stored here.
+   * 
+   * @return the calibration information
+   */
+  public abstract Optional<CurveCalibrationInfo> getCalibrationInfo();
```
cjkent: Why is this on the metadata rather than the curve itself? It seems a bit of a stretch to describe the Jacobian matrix as metadata. Arguably it's part of the curve data.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveCalibrationInfo.java
jodastephen: The implementation is in a different package, so an interface is needed. Plus, it avoids locking in our specific way of doing curve calibration. The convert method gives a little more flexibility to us in the future, such as converting data on the fly between formats.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveMetadata.java
jodastephen: It is all pretty important stuff now, not just the calibration info. The x/y types must be set, as must the day count and name. There is potential to rename `CurveMetadata` to `CurveInfo`,

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveCalibrationInfo.java
cjkent: It smells like a marker interface that might or might not know how to convert itself to another type. But probably doesn't. Despite my distaste for marker interfaces, why not just use one rather than having the `convertTo` method? It would be more conventional.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveMetadata.java
cjkent: Why is it not part of the curve itself?

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveCalibrationInfo.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+package com.opengamma.strata.market.curve;
+
+/**
+ * Information about curve calibration.
+ * <p>
+ * Implementations of this marker interface provide information that was
+ * obtained during curve calibration, such as a Jacobian.
+ */
+public interface CurveCalibrationInfo {
```
cjkent: Why do we need this interface? Why not have:

    CurveInfoType<Jacobean> JACOBEAN = CurveInfoType.of("Jacobean");

What do we gain with the extra level of indirection where we have to cast `CurveCalibrationInfo` to `JacobeanCurveCalibrationInfo`?

jodastephen: It is used in `NodalCurveDefinition` to allow the calibration data to be added after/separately from building the curve. I tried to remove it, but it would have locked in a type in a location that would be undesirable, limiting flexibility.

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadata.java
```diff
@@ -137,26 +164,27 @@ public DefaultCurveMetadata withParameterMetadata(List<CurveParameterMetadata> p
...
       List<? extends CurveParameterMetadata> parameterMetadata) {
     JodaBeanUtils.notNull(curveName, "curveName");
     JodaBeanUtils.notNull(xValueType, "xValueType");
     JodaBeanUtils.notNull(yValueType, "yValueType");
+    JodaBeanUtils.notNull(info, "info");
     this.curveName = curveName;
     this.xValueType = xValueType;
     this.yValueType = yValueType;
-    this.dayCount = dayCount;
+    this.info = ImmutableMap.copyOf(info);
```
cjkent: Should probably validate that the types of keys and values are compatible. We're just assuming they are. If they're not the failure won't happen until someone tries to look up a value.

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadataBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Sets the day count.
+   * <p>
+   * This stores the day count in the additional information map using the
+   * key {@link CurveInfoType#DAY_COUNT}.
+   * 
+   * @param dayCount  the day count, may be null
+   * @return this, for chaining
+   */
+  public DefaultCurveMetadataBuilder dayCount(DayCount dayCount) {
```
cjkent: The only reason I can think of for allowing a null argument is to allow reuse of an existing builder where specifying a null day count would clear the previously set day count. 

That doesn't happen here; if the day count is null the previous day count is left in place and the method returns successfully. I think the behaviour should be either of the following when the day count is null

* Throw an `IllegalArgumentException`
* `info.remove(CurveInfoType.DAY_COUNT)`

I don't have a strong opinion on which one it is.

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadataBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Sets the calibration information.
+   * <p>
+   * This stores the calibration information in the additional information map
+   * using the key {@link CurveInfoType#JACOBIAN}.
+   * 
+   * @param jacobian  the calibration information, may be null
+   * @return this, for chaining
+   */
+  public DefaultCurveMetadataBuilder jacobian(JacobianCurveCalibration jacobian) {
```
cjkent: Same comment as above about handling of null values.

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadataBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+   * <p>
+   * This is stored in the additional information map using {@code Map.put} semantics
+   * 
+   * @param type  the type to store under
+   * @param instance  the instance to store, may be null
+   * @return this, for chaining
+   */
+  public <T> DefaultCurveMetadataBuilder addInfo(CurveInfoType<T> type, T instance) {
+    ArgChecker.notNull(type, "type");
+    if (instance != null) {
```
cjkent: Same as above.

modules/market/src/main/java/com/opengamma/strata/market/curve/JacobianCurveCalibration.java
```diff
@@ -29,49 +27,41 @@
...
+ * Jacobian matrix information produced during curve calibration.
  * <p>
- * This provides details of the names of the curves and the number of parameters they have.
- * This is used in curve calibration to create and split a single array of all parameters from all curves.
+ * The inverse Jacobian matrix produced using curve calibration is stored here.
+ * The information is used to calculate market quote sensitivity.
  */
 @BeanDefinition(builderScope = "private")
-public final class CurveBuildingBlock
+public final class JacobianCurveCalibration
```
cjkent: I'm not sure about the class name. A curve calibration isn't something you'd expect to be represented as an object. How about `CurveJacobian` or `CurveCalibrationJacobian`? Or, given we're in the `curve` package, just `Jacobian`?

modules/market/src/main/java/com/opengamma/strata/market/curve/JacobianCurveCalibration.java
```diff
@@ -29,49 +27,41 @@
...
     implements ImmutableBean, Serializable {
 
   /**
-   * The empty block.
+   * The order of the curves in the block that was calibrated together.
```
cjkent: s/was/were/

modules/market/src/main/java/com/opengamma/strata/market/curve/JacobianCurveCalibration.java
```diff
@@ -29,49 +27,41 @@
...
+   * This is the derivative of the curve parameters with respect to the market quotes.
    */
   @PropertyDefinition(validate = "notNull")
-  private final ImmutableList<CurveParameterSize> data;
+  private final DoubleMatrix2D jacobianMatrix;
 
   //-------------------------------------------------------------------------
   /**
-   * Obtains an empty block.
+   * Obtains a block from a map of curve name to curve block and Jacobian matrix.
```
cjkent: The comment still talks about a "block".

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadata.java
jodastephen: We'll defer that good idea to a later issue

modules/market/src/main/java/com/opengamma/strata/market/curve/DefaultCurveMetadataBuilder.java
jodastephen: Good point, changed to remove

modules/market/src/main/java/com/opengamma/strata/market/curve/JacobianCurveCalibration.java
jodastephen: `JacobianCalibrationMatrix`

cjkent: Much better.

modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/CurveCalibrator.java
```diff
@@ -143,99 +142,109 @@ private CurveCalibrator(
...
+
+      // use Jacobians to build output curves
+      providerCombined = providerGenerator.generate(calibratedGroupParams, jacobians);
     }
     // create the final block bundle and return
-    return Pair.of(combinedProvider, blockBundleBuilder.build());
+    return providerCombined;
   }
 
   // converts a definition to a block
```
cjkent: Comment is out of date

modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/CurveCalibrator.java
```diff
@@ -143,99 +142,109 @@ private CurveCalibrator(
...
-    return providerGenerator.generate(calibratedParams);
+    return rootFinder.getRoot(valueCalculator, derivativeCalculator, initGuessMatrix).getData();
   }
 
   //-------------------------------------------------------------------------
   // calculates the Jacobian and builds the result, called once per group
-  private CurveBuildingBlock updateBlockBundleForGroup(
-      List<Trade> trades,
+  // this uses, but does not alter, data from previous groups
+  private ImmutableMap<CurveName, JacobianCurveCalibration> updateBlockBundleForGroup(
```
cjkent: Method name is out of date.

PR #488, opened 2015-10-09
Enhance matrix classes

jodastephen:
Add new methods to make classes more generally useful.
Prepare for deletion of existing methods, but avoid the churn in this PR.

### DIFF
--- a/modules/math/src/main/java/com/opengamma/strata/math/impl/matrix/DoubleMatrix1D.java
+++ b/modules/math/src/main/java/com/opengamma/strata/math/impl/matrix/DoubleMatrix1D.java
@@ -5,60 +5,398 @@
  */
 package com.opengamma.strata.math.impl.matrix;
 
+import java.io.Serializable;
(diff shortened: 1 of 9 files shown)


modules/math/src/main/java/com/opengamma/strata/math/impl/matrix/DoubleMatrix1D.java
```diff
@@ -5,60 +5,398 @@
...
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance from an array of {@code double}.
+   * 
+   * @param array  the array to copy, cloned
+   */
+  private DoubleMatrix1D(double[] array, boolean trusted) {
+    this.array = array;
```
jonathansenior: `trusted` is undocumented and unused

jodastephen: This is in preparation for the next commit

PR #490, opened 2015-10-12
SABR model transfer

yukiiwashita:
This work closes #489


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/option/SABRInterestRateParameters.java
```diff
@@ -0,0 +1,576 @@
...
+  @PropertyDefinition(validate = "notNull")
+  private final NodalSurface rhoSurface;
+  /**
+   * The nu (volatility of volatility) surface. 
+   * <p>
+   * The first dimension is the expiration and the second the tenor.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final NodalSurface nuSurface;
+  /**
```
marc-opengamma: We should probably incorporate the "shift" in this version. I don't think there is value to have one model with SABR and one with "shifted SABR". It is enough to have a constructor that create a shift of a constant 0 for the simpler case.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/function/SABRHaganVolatilityFunctionProvider.java
```diff
@@ -0,0 +1,993 @@
...
+   * This returns an array with [0] the volatility, [1] Derivative w.r.t the forward, [2] the derivative w.r.t the strike, 
+   * [3] the derivative w.r.t. to alpha, [4] the derivative w.r.t. to beta, [5] the derivative w.r.t. to rho, and 
+   * [6] the derivative w.r.t. to nu.
+   * 
+   * @param option The option.
+   * @param forward The forward value of the underlying
+   * @param data The SABR data.
+   * @return the volatility and sensitivities
+   */
+  public double[] getVolatilityAdjoint(EuropeanVanillaOption option, double forward, SABRFormulaData data) {
```
marc-opengamma: For "Adjoint AD" version of functions, we should standardize the type of object which is returned. I will create a "DoubleDerivatives" or "ValueDerivatives" object as a return type.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/function/SABRHaganVolatilityFunctionProvider.java
```diff
@@ -0,0 +1,913 @@
...
+   * <p>
+   * This returns an array with alpha, beta, rho and nu sensitivities.
+   * 
+   * @param option  the option.
+   * @param forward  the forward value of the underlying
+   * @param data  the SABR data.
+   * @return the sensitivities
+   */
+  @Override
+  public DoubleArray getVolatilityModelAdjoint(EuropeanVanillaOption option, double forward, SABRFormulaData data) {
```
marc-opengamma: For adjoint versions, we should now use the "ValueDerivatives" as output object.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/function/SABRHaganVolatilityFunctionProvider.java
```diff
@@ -0,0 +1,913 @@
...
+  private static final double SMALL_Z = 1e-6;
+  private static final double LARGE_NEG_Z = -1e6;
+  private static final double LARGE_POS_Z = 1e8;
+  private static final double BETA_EPS = 1e-8;
+  private static final double RHO_EPS = 1e-5;
+  private static final double RHO_EPS_NEGATIVE = 1e-8;
+  private static final double ATM_EPS = 1e-7;
+
+  @Override
+  public double getVolatility(EuropeanVanillaOption option, double forward, SABRFormulaData data) {
```
marc-opengamma: Do we want a "EuropeanVanillaOption" object as input? In the object we don't use the Put/Call part. Moreover, in a lot of application we don't have such an object and it is created locally only for the call. This is the case for swaptions, where the strike is computed in the pricer.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/function/VolatilityFunctionProvider.java
```diff
@@ -0,0 +1,127 @@
...
+   * Computes the volatility sensitivity to the model parameters by means of central finite difference. 
+   * <p>
+   * This should be overridden where possible. 
+   * 
+   * @param option  the option
+   * @param forward  the forward value of the underlying
+   * @param data  the model data
+   * @return the sensitivities
+   */
+  public DoubleArray getVolatilityModelAdjoint(EuropeanVanillaOption option, double forward, T data) {
```
marc-opengamma: Why do we have two volatility + derivatives (adjoint) version. One should be enough.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/function/VolatilityFunctionProvider.java
```diff
@@ -0,0 +1,127 @@
...
+   * 
+   * @param option  the option
+   * @param forward  the forward value of the underlying
+   * @param data  the model data
+   * @return the volatility
+   */
+  public abstract double getVolatility(EuropeanVanillaOption option, double forward, T data);
+
+  /**
+   * Computes the volatility sensitivity to the model parameters by means of central finite difference.
```
marc-opengamma: Should indicate that a default central FD is implemented at this level but is likely overridden in the non-abstract implementations.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/fitting/SABRModelFitter.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Override
+  protected NonLinearParameterTransforms getTransform(DoubleArray start, BitSet fixed) {
+    return new UncoupledParameterTransforms(start, DEFAULT_TRANSFORMS, fixed);
+  }
+
+  @Override
+  protected DoubleArray getMaximumStep() {
+    return null;
```
marc-opengamma: Should this be null?

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/option/SABRInterestRateParameters.java
```diff
@@ -0,0 +1,713 @@
...
+   * Returns the volatility for given expiry, tenor, strike and forward rate.
+   * 
+   * @param expiryTime  time to expiry
+   * @param tenor  tenor
+   * @param strike  the strike
+   * @param forward  the forward
+   * @return the volatility
+   */
+  public double getVolatility(double expiryTime, double tenor, double strike, double forward) {
+    DoublesPair expirytTenor = DoublesPair.of(expiryTime, tenor);
```
marc-opengamma: Typo in name: "expiryTenor"

PR #496, opened 2015-10-14
 Leg Pricer: Added par spread for payment with FLAT compounding

marc-opengamma:
ParSpread and parSpreadSensitivity was supported only for payment period with only one accrual period. The method have been extended to support payments with compounding of accrual periods with compounding method “FLAT”.
Those methods are required for curve calibration with basis swaps. The FLAT compounding is the standard one for the liquid basis swaps.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/rate/swap/DiscountingKnownAmountPaymentPeriodPricer.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/rate/swap/DiscountingKnownAmountPaymentPeriodPricer.java
@@ -75,6 +75,11 @@ public class DiscountingKnownAmountPaymentPeriodPricer
     return fv * (partialDays / totalDays);
   }
 
+  @Override
+  public double pvbp(KnownAmountPaymentPeriod period, RatesProvider provider) {
+    throw new UnsupportedOperationException("Unable to calculate PVBP for KnownAmountPaymentPeriod");
+  }
+
   //-------------------------------------------------------------------------
   @Override
   public PointSensitivityBuilder presentValueSensitivity(KnownAmountPaymentPeriod period, RatesProvider provider) {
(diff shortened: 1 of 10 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/rate/swap/DiscountingKnownAmountPaymentPeriodPricer.java
```diff
@@ -75,6 +75,11 @@ public double accruedInterest(KnownAmountPaymentPeriod period, RatesProvider pro
     return fv * (partialDays / totalDays);
   }
 
+  @Override
+  public double pvbp(KnownAmountPaymentPeriod period, RatesProvider provider) {
+    return 0;
```
jodastephen: If the result cannot be calculated, better to throw UnsupportedOperationException

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/rate/swap/DiscountingRatePaymentPeriodPricer.java
```diff
@@ -404,5 +444,69 @@ private void explainPresentValue(
     builder.put(ExplainKey.PAY_OFF_RATE, accrualPeriod.getNegativeRateMethod().adjust(payOffRate));
     builder.put(ExplainKey.UNIT_AMOUNT, ua);
   }
+  
+  // Sensitivity to the spread for a payment period with FLAT compounding type
+  private double pvbpCompoundedFlat(RatePaymentPeriod paymentPeriod, RatesProvider provider) {
+    int nbCmp = paymentPeriod.getAccrualPeriods().size();
+    double[] rate = new double[nbCmp];
+    int i = 0;
```
jodastephen: Would prefer to see loops replaced with standard "for (int i = 0; ...)" style, rather than declaring 'i' and manually incrementing it. A stream would also work, using 'mapToDouble()' and 'toArray()'.

PR #501, opened 2015-10-16
Rename and move matrix classes

jodastephen:
Move out from strata-math implementation to strata-collect
Massive PR of a refactoring change.


jodastephen:
Merged directly so rename is in mainline of git history.

PR #505, opened 2015-10-20
Move trade interface to strata-basics

jodastephen:
Add new interface FinanceTrade in strata-finance


jodastephen:
Merged simple refactoring change.

PR #514, opened 2015-10-23
Enhance conventions

jodastephen:
Allow FRA and deposit conventions to be looked up by name.
Rename `Convention` to `TradeConvention` and rename `Template` to `TradeTemplate`.
Add new Term Deposit conventions.
Tidy up CDS conventions.
Further work is needed on swap and CDS conventions


jodastephen:
Merging for later review.

PR #516, opened 2015-10-26
Curve calibration with Ibor Futures

marc-opengamma:
Added the required Calibration Measure.
Corrected Ibor Future pricer to deal with trade price on trade date.
Added curve calibration example.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/TradeCalibrationMeasure.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/TradeCalibrationMeasure.java
@@ -13,6 +13,7 @@ import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.finance.rate.deposit.IborFixingDepositTrade;
 import com.opengamma.strata.finance.rate.deposit.TermDepositTrade;
 import com.opengamma.strata.finance.rate.fra.FraTrade;
+import com.opengamma.strata.finance.rate.future.IborFutureTrade;
 import com.opengamma.strata.finance.rate.swap.SwapTrade;
 import com.opengamma.strata.market.sensitivity.CurveCurrencyParameterSensitivities;
 import com.opengamma.strata.market.sensitivity.PointSensitivities;
@@ -20,6 +21,7 @@ import com.opengamma.strata.pricer.rate.ImmutableRatesProvider;
 import com.opengamma.strata.pricer.rate.deposit.DiscountingIborFixingDepositProductPricer;
 import com.opengamma.strata.pricer.rate.deposit.DiscountingTermDepositProductPricer;
 import com.opengamma.strata.pricer.rate.fra.DiscountingFraProductPricer;
+import com.opengamma.strata.pricer.rate.future.DiscountingIborFutureTradePricer;
 import com.opengamma.strata.pricer.rate.swap.DiscountingSwapProductPricer;
 
 /**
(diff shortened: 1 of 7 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/TradeCalibrationMeasure.java
```diff
@@ -43,6 +45,16 @@
           (trade, p) -> DiscountingFraProductPricer.DEFAULT.parSpreadSensitivity(trade.getProduct(), p));
 
   /**
+   * The calibrator for {@link IborFutureTrade} using par spread discounting.
+   */
+  public static final TradeCalibrationMeasure<IborFutureTrade> IBOR_FUT_PAR_SPREAD =
```
jodastephen: Should be FUTURE, not FUT

modules/pricer/src/main/java/com/opengamma/strata/pricer/rate/future/AbstractIborFutureTradePricer.java
```diff
@@ -48,5 +51,27 @@ public CurrencyAmount presentValue(IborFutureTrade trade, double currentPrice, d
...
+   * 
+   * @param trade  the trade to price
+   * @param valuationDate  the date for which the reference price should be calculated
+   * @param lastMarginPrice  the last price used in the margining 
+   * @return the reference price
+   */
+  public double referencePrice(IborFutureTrade trade, LocalDate valuationDate, double lastMarginPrice) {
+    ArgChecker.notNull(valuationDate, "valuation date");
+    Optional<LocalDate> tradeDate = trade.getTradeInfo().getTradeDate();
+    ArgChecker.isTrue(tradeDate.isPresent(), "trade date should be populated");
```
jodastephen: This can be better written as 

    LocalDate tradeDate = trade.getTradeInfo().getTradeDate()
        .orElseThrow(() -> new IllegalArgumentException("Trade date should be populated"));

modules/pricer/src/main/java/com/opengamma/strata/pricer/rate/future/AbstractIborFutureTradePricer.java
```diff
@@ -48,5 +51,27 @@ public CurrencyAmount presentValue(IborFutureTrade trade, double currentPrice, d
...
+    double referencePrice;
```
jodastephen: In cases like this, it is better to avoid the local variable and just return from each part of the if statement. Or use a ternary operator `(a ? b : c)`

modules/pricer/src/test/java/com/opengamma/strata/pricer/calibration/CalibrationZeroRateUsd2OisFuturesIrsTest.java
```diff
@@ -0,0 +1,393 @@
...
+  static {
+    DSC_NAMES.put(DSCON_CURVE_NAME, USD);
+    Set<Index> usdFedFundSet = new HashSet<>();
+    usdFedFundSet.add(USD_FED_FUND);
+    IDX_NAMES.put(DSCON_CURVE_NAME, usdFedFundSet);
+    Set<Index> usdLibor3Set = new HashSet<>();
+    usdLibor3Set.add(USD_LIBOR_3M);
+    IDX_NAMES.put(FWD3_CURVE_NAME, usdLibor3Set);
+    TS.put(USD_FED_FUND, TS_EMTPY);
+    TS.put(USD_LIBOR_3M, TS_EMTPY);
```
jodastephen: Time-series are now optional for rate providers, so these are probably not needed.

modules/pricer/src/test/java/com/opengamma/strata/pricer/calibration/CalibrationZeroRateUsd2OisFuturesIrsTest.java
```diff
@@ -0,0 +1,393 @@
...
+        (ov) -> CALIBRATOR.calibrate(CURVE_GROUP_CONFIG, VALUATION_DATE, ov, TS, FxMatrix.empty());
+    calibration_market_quote_sensitivity_check(f, shift);
+  }
+
+  private void calibration_market_quote_sensitivity_check(
+      Function<ImmutableObservableValues, ImmutableRatesProvider> calibrator,
+      double shift) {
+    double notional = 100_000_000.0;
+    double spread = 0.0050;
+    SwapTrade trade = FixedIborSwapConventions.USD_FIXED_1Y_LIBOR_3M.toTrade(VALUATION_DATE, Period.ofMonths(8),
```
jodastephen: Coding standards would break this line after `toTrade(`. We avoid having a line break in the middle of the parameter list like this.

PR #521, opened 2015-10-29
Remove group name from curve files

jodastephen:
Curve settings file and curve file should not contain the group name.
This is because the curve name must be unique across the system.
Beans have also been simplified and made package scoped.


examples/src/main/java/com/opengamma/strata/examples/marketdata/curve/RatesCurvesCsvLoader.java
```diff
@@ -54,7 +54,7 @@
  *      
  * <li> Settings - contains the settings that apply to each named curve across all valuation
  *      dates. This must contain the following header row:
- *      Group Name, Curve Name, Day Count, Interpolator, Left Extrapolator, Right Extrapolator, Value Type
+ *      Curve Name, Value Type, Day Count, Interpolator, Left Extrapolator, Right Extrapolator
  * </ul>
  * There may be one or more curves files with the following header row:
  * Valuation Date, Group Name, Curve Name, Date, Value, Label
```
yukiiwashita: remove Group Name

PR #522, opened 2015-10-30
Enhance swap conventions

jodastephen:
Allow conventions to be referred to by name.


jodastephen:
Merging for later review

PR #523, opened 2015-10-30
Enhance conventions

jodastephen:
Allow conventions to be obtained by name.


jodastephen:
Merging for later review.

PR #524, opened 2015-10-30
Enhance CDS conventions

jodastephen:
Match design of other trade conventions.
Replace `CdsStubType` with direct use of `StubConvention`.


jodastephen:
Merging for later review.

PR #526, opened 2015-10-30
Add fixings time-series loader

jodastephen:
Moved logic from strata-examples


modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -0,0 +1,80 @@
...
+  //-------------------------------------------------------------------------
+  /**
+   * Attempts to locate a rate index by reference name.
+   * <p>
+   * This utility searches both {@link IborIndex} and {@link OvernightIndex}.
+   * 
+   * @param reference  the reference name
+   * @return the resolved rate index
+   */
+  public static Index findIndex(String reference) {
```
cjkent: Should this return `RateIndex` if it only searches `IboxIndex` and `OvernightIndex`?

cjkent: I've just noticed the implementation can return other types - the Javadoc is wrong

PR #532, opened 2015-11-04
Modify market data API so one object can contain data for many scenarios

### LINKED ISSUE #431: Change market data API to always handle multiple values of market data
There is a mismatch between the calculation and market data APIs. The API of calculation functions handles multiple scenarios in a single call. For historical reasons a lot of the market data and scenario API deals in single market data values. The gap between these approaches is bridged by `CalculationMarketData` which returns lists of single values of market data, where the list represents all market data for a set of scenarios.

If the market data functions and perturbations were changed to handle data for multiple scenarios in one call it would clean up a number of APIs and enable optimisations that are currently difficult or impossible.

### LINKED ISSUE #434: Special case handling of double values in scenario market data
Double values in scenario data are stored in a multimap with every other type of market data. It might be worth treating them as a special case at storing them in primitive arrays. That saves the overhead of boxing which can be significant for very large data sets.

There is a question about how the data should be returned to functions:

* Immutable `List<Double>` backed by the array - avoids changing the API exposed to functions
* `double[]` that is defensively copied - gives the data to functions in a compact form but lots of copying
* `double[]` that holds the underlying data - efficient but dangerous

cjkent:
This PR changes the market data API to allow a single object to hold the data for multiple scenarios. This has a number of benefits:

* Reduced memory footprint; for example, curves can share the same set of X values across all scenarios or primitives can be stored in arrays, avoiding boxing overhead
* New optimisations are possible; if there are parts of a calculation which are the same for every scenario, the new design allows them to be reused rather than repeated every time.

Fixes #434 
Fixes #431


jodastephen:
* `ScenarioValuesList` should have `@param <T>` at class level
* No test for `ScenarioValuesList`
* Removing `ObservableIdTest` and `ObservableKeyTest` lowers test coverage
* Why `CalculationEnvironment.contanisValue()` less strict now?
* `CalculationEnvironmentBuilder.addValueUnsafe()` Javadoc refers to generic T type that is no longer used
* `CalculationEnvironmentBuilder.updateScenarioCount()` allows a scenario count of 1 to be replaced. So why not use 1 everywhere (or -0 or 1) instead of null? Maybe 1 should not be overwritten?
* Similar with `MarketEnvironmentBuilder`.
* `CalculationMarketData` more Javadoc for `getScenarioValue()` indicating possible conversions (not expected by being named "get".
* `DefaultMarketDataFactory` TODO suggests need for `ScenarioDefinition.empty()`
* `DefaultMarketDataFactory` needs to suppress warnings rawtypes in some places
* `MarketDataBox` (and some others) need `@throws` clauses in docs
* `PerturbationMapping` Javadoc reads wrong
* `PerturbationMapping.applyPerturbation()` refers to T and U in comments twice, but there is no U
* `ScenarioDefinition` all combinations has gone?!
* Do `SingleMarketDataBox` and `ScenarioMarketDataBox` need to be a public class?
* Would prefer to see instance methods after factory methods
* `ScenarioMarketDataBox` missing `@Override` on `apply()`
* `SingleMarketDataBox` missing `@Override` on `getValue()`
* `ScenarioPerturbation` inner class will need to be a package scoped bean
* `marketData.getValue(id1).getSingleValue()` looks unattractive. Maybe rename method on `CalculationEnvironment` to remove 'value'?
* `CurveGroupMarketDataFunction.ratesByKey()` should return `ObservableValues`
* `CurveGroupMarketDataFunction` and `ParRatesMarketDataFunction` are a lot bigger now, and would benefit from some // style method comments to help the eye navigate
* `CurveParallelShifts` should use `DoubleArray` not `double[]`

Generally though, this looks pretty good.

PR #533, opened 2015-11-04
Add curve calibration CSV loader

jodastephen:
Allows curve calibration to be defined in CSV.
The curve node parsing is included here for now in a restrictive way. A future change will be needed to make this more flexible/extensible.


modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -0,0 +1,368 @@
...
+  // create the curve node
+  private static CurveNode createCurveNode(
+      String typeStr,
+      String conventionStr,
+      String timeStr,
+      String label,
+      QuoteKey quoteKey,
+      double spread) {
+
+    if ("FIX".equals(typeStr) || "IborFixingDeposit".equals(typeStr)) {
```
cjkent: It might be nice to move these magic strings into constants.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -0,0 +1,368 @@
...
+  private static CurveNode curveFraCurveNode(
+      String conventionStr,
+      String timeStr,
+      String label,
+      QuoteKey quoteKey,
+      double spread) {
+
+    Matcher matcher = FRA_TIME_REGEX.matcher(timeStr.toUpperCase(Locale.ENGLISH));
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(Messages.format("Invalid time format for FRA: {}", timeStr));
```
cjkent: Is there an easy way to express what a valid time format would look like? It would be good to add that to the error message.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -0,0 +1,368 @@
...
+  private static CurveNode curveFixedOvernightCurveNode(
+      String conventionStr,
+      String timeStr,
+      String label,
+      QuoteKey quoteKey,
+      double spread) {
+
+    Matcher matcher = SIMPLE_TIME_REGEX.matcher(timeStr.toUpperCase(Locale.ENGLISH));
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(Messages.format("Invalid time format for OIS: {}", timeStr));
```
cjkent: Same as above

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -0,0 +1,368 @@
...
+  private static CurveNode curveFixedIborCurveNode(
+      String conventionStr,
+      String timeStr,
+      String label,
+      QuoteKey quoteKey,
+      double spread) {
+
+    Matcher matcher = SIMPLE_TIME_REGEX.matcher(timeStr.toUpperCase(Locale.ENGLISH));
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(Messages.format("Invalid time format for IRS: {}", timeStr));
```
cjkent: Same as above

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -0,0 +1,368 @@
...
+  private static CurveNode curveIborIborCurveNode(
+      String conventionStr,
+      String timeStr,
+      String label,
+      QuoteKey quoteKey,
+      double spread) {
+
+    Matcher matcher = SIMPLE_TIME_REGEX.matcher(timeStr.toUpperCase(Locale.ENGLISH));
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(Messages.format("Invalid time format for IRS: {}", timeStr));
```
cjkent: Same as above

modules/market/src/main/java/com/opengamma/strata/market/curve/definition/CurveGroupEntry.java
```diff
@@ -69,6 +70,24 @@
   @PropertyDefinition(validate = "notNull")
   private final ImmutableSet<OvernightIndex> overnightIndices;
 
+  //-------------------------------------------------------------------------
+  /**
+   * Merges the specified entry with this entry.
```
cjkent: Probably worth mentioning that a new value is returned and this value and the input aren't modified.

cjkent:
That's everything

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
jodastephen: I expect this code to be rewritten relatively soon (hence my comment in the PR description), so it doesn't seem worth doing this now.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
jodastephen: As previous comment, the rework to a better structure will fix this.

PR #547, opened 2015-11-11
Hull White model for IR futures

yukiiwashita:


modules/pricer/src/main/java/com/opengamma/strata/pricer/rate/future/HullWhiteOneFactorPiecewiseConstantConvexityFactorProvider.java
```diff
@@ -0,0 +1,421 @@
...
+import com.opengamma.strata.pricer.impl.rate.model.HullWhiteOneFactorPiecewiseConstantInterestRateModel;
+import com.opengamma.strata.pricer.impl.rate.model.HullWhiteOneFactorPiecewiseConstantParameters;
+
+/**
+ * Provider of convexity factor for futures based on Hull-White one factor model with piecewise constant volatility.
+ * <p>
+ * Reference: Henrard, M. "The Irony in the derivatives discounting Part II: the crisis", Wilmott Journal, 2010, 2, 301-316
+ */
+@BeanDefinition(builderScope = "private")
+public final class HullWhiteOneFactorPiecewiseConstantConvexityFactorProvider implements ImmutableBean, Serializable {
```
marc-opengamma: Should we have a provider of only the "convexity adjustment" or more generally of the "Hull-White parameters". I'm more in favor of the second. The same provider should be used for any instrument and any valuation.
Same question will appear for CMS: do we use the "SabrVolatilitySwaptionProvider" or do we create another provider. I'm more of the opinion that the provider provides the data/parameters and it is up to the pricer to use them wisely. In the present case, I would prefer to have a "HullWhiteOneFactorPiecewiseConstantParametersProvider".

PR #553, opened 2015-11-13
Joda-Beans v1.7

jodastephen:


jodastephen:
Self merging dependency upgrade

PR #554, opened 2015-11-13
Remove Result from MarketDataFunction.build

### LINKED ISSUE #486: Better error messages in market data building
When market data cannot be built, the engine still continues on to the calculation step. However, this should not happen if there is a misconfiguration of the market data functions, as otherwise the error is completely hidden.

cjkent:
This PR removes the `Result` from the return type of `MarketDataFunction.build`. This has three benefits:

* More consistent API
* Less surprising API; it now returns objects and throws exceptions like most Java APIs
* Removes boilerplate associated with `Result`

After this change user code should not be expected to create a `Result` and can indicate failures by throwing an exception.

This also fixes #486 so market data building fails with an exception if no market data function is available for an ID type. This is a system configuration error and should terminate the attempt to build data.


jodastephen:
That looks like a good simplification and a performance enhancement.

PR #558, opened 2015-11-16
Rename FuturesExpiry to YearMonth

jodastephen:
Make concept more general and match YearMonthCurveNodeMetadata name


modules/market/src/main/java/com/opengamma/strata/market/curve/YearMonthCurveNodeId.java
```diff
@@ -29,16 +29,18 @@
...
  */
 @BeanDefinition(builderScope = "private")
-public final class FuturesExpiryCurveNodeId
+public final class YearMonthCurveNodeId
     implements CurveNodeId, ImmutableBean, Serializable {
 
   /**
-   * The year and month of the futures expiry date.
+   * The year-month of the curve node.
+   * The is typically the futures expiry date.
```
cjkent: Should be "This is"

PR #559, opened 2015-11-16
Move Perturbation

jodastephen:
No need to be in strata-basics


modules/market/src/main/java/com/opengamma/strata/market/curve/perturb/Perturbation.java
```diff
@@ -3,7 +3,7 @@
  *
  * Please see distribution for license.
  */
-package com.opengamma.strata.basics.market;
+package com.opengamma.strata.market.curve.perturb;
```
cjkent: It feels like `Perturbation` is a general enough concept to be in the `market` package rather than `market.curve`

PR #560, opened 2015-11-16
Add MarketDataValue abstraction

jodastephen:
Add DiscountFactors.of()
Change PriceIndexValues to have valuation date instead of month

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/MarketDataValue.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/MarketDataValue.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2015 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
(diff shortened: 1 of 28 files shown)


modules/market/src/main/java/com/opengamma/strata/market/MarketDataValue.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public interface MarketDataValue<T extends MarketDataValue<T>> {
+
+  /**
+   * Gets the key used to obtain this market data.
+   * <p>
+   * A {@link MarketDataKey} is used to identify market data within a given context.
+   * A system may have multiple market data values with the same key, and {@link MarketDataId}
+   * is used to uniquely identify each.
+   * <p>
```
cjkent: It would be good to give a more concrete example of what a key represents vs an ID. Something like "there is a single USD discounting curve used in a calculation, but the system can contain multiple USD discounting curves used for different asset classes / couterparties / whatever".

PR #561, opened 2015-11-16
Replace ObservableValues with MarketData

### LINKED ISSUE #556: Use MarketDataLookup instead of CalculationEnvironment in the CalculationRunner API
`MarketDataLookup` is an interface and `CalculationEnvironment` is a bean which implements it. The `CalculationRunner` interface takes `CalculationEnvironment` but it would be better if it took an interface. Converting it to use `MarketDataLookup` is a trivial change.

As part of this it might be good to swap some names around. Arguably `CalculationEnvironment` is a better name which could be used for the interface and the implementing bean could be called something else. `CalculationMarketDataMap` maybe.

cjkent:
This PR removes `ObservableValues` and replaces it with `MarketData` which can contain non-observable data and time series in addition to observable data. This is in preparation for supporting calibration of curves using `FxSwapCurveNode`.

`MarketData` doesn't expose the valuation date. This is likely to be necessary in future but it not required yet so it hasn't been included.

Some of the existing market data containers are renamed and the calculation methods on `CalculationRunner` now expect a market data container interface instead of a bean. This fixes #556.


examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
```diff
@@ -203,7 +208,7 @@ public static void performance_calibration_pricing() {
       for (CurveNode node : nodes) {
         if (!(node instanceof IborFixingDepositCurveNode)) {
           // IborFixingDeposit is not a real trade, so there is no appropriate comparison
-          trades.add(node.trade(VALUATION_DATE, ObservableValues.ofIdMap(quotes)));
+          trades.add(node.trade(VALUATION_DATE, MarketData.builder().addValues(quotesByKey).build()));
```
jodastephen: Since `MarketDataBuilder` is a concrete class, it would be worth adding a `addValuesById()` method to simplify user code like this.

modules/basics/src/main/java/com/opengamma/strata/basics/market/MarketData.java
```diff
@@ -0,0 +1,60 @@
...
+  public abstract LocalDateDoubleTimeSeries getTimeSeries(ObservableKey key);
+
+  /**
+   * Returns a mutable builder for building instances of {@code MarketData}.
+   *
+   * @return a mutable builder for building instances of {@code MarketData}
+   */
+  public static MarketDataBuilder builder() {
+    return new MarketDataBuilder();
+  }
```
jodastephen: Would prefer to see static methods first in interfaces.

examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
cjkent: I did consider it but there are only 2 instances of this pattern in the whole codebase so I wasn't sure it was worth adding something to the API.

modules/basics/src/main/java/com/opengamma/strata/basics/market/MarketDataBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+    return this;
+  }
+
+  /**
+   * Adds multiple time series of market data values to the builder.
+   *
+   * @param timeSeriesMap  the time series
+   * @return this builder
+   */
+  public MarketDataBuilder addTimeSeries(Map<ObservableKey, LocalDateDoubleTimeSeries> timeSeriesMap) {
```
jodastephen: Should be `? extends ObservableKey`

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableMarketData.java
```diff
@@ -0,0 +1,400 @@
...
+import org.joda.beans.impl.direct.DirectMetaPropertyMap;
+
+import com.google.common.collect.ImmutableMap;
+import com.opengamma.strata.collect.timeseries.LocalDateDoubleTimeSeries;
+
+/**
+ * Immutable bean implementation of {@link MarketData}.
+ */
+@BeanDefinition
+public final class ImmutableMarketData implements MarketData, ImmutableBean {
```
jodastephen: Serializable (and test)

modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/CalculationEnvironmentBuilder.java
```diff
@@ -254,7 +254,19 @@ public CalculationEnvironmentBuilder valuationDate(MarketDataBox<LocalDate> valu
...
   public CalculationEnvironment build() {
-    return new CalculationEnvironment(valuationDate, scenarioCount, values, timeSeries, valueFailures, timeSeriesFailures);
+    return new CalculationMarketDataMap(valuationDate, scenarioCount, values, timeSeries, valueFailures, timeSeriesFailures);
+  }
+
+  /**
+   * Builds a set of market data from the data in this builder.
+   * <p>
+   * It is possible to continue to add more data to a builder after calling {@code build()}. Any
+   * {@code BaseMarketData} instances built previously will be unaffected.
```
jodastephen: Comment refers to a dead class

modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/CalculationMarketData.java
```diff
@@ -17,7 +17,7 @@
 /**
  * A source of market data provided to an engine function and used for a calculation across multiple scenarios.
  * <p>
- * The set of data provided by this interface is a subset of the set provided by {@link CalculationEnvironment}.
+ * The set of data provided by this interface is a subset of the set provided by {@link CalculationMarketDataMap}.
```
jodastephen: Is it better to refer to the interface, not the concrete type here?

modules/function/src/test/java/com/opengamma/strata/function/marketdata/curve/CurveEndToEndTest.java
```diff
@@ -135,10 +135,13 @@ public void roundTripFraAndFixedFloatSwap() {
...
         .build();
 
+    Map<ObservableKey, Double> parRatesByKey =
+        parRateData.entrySet().stream().collect(toImmutableMap(e -> e.getKey().toObservableKey(), e -> e.getValue()));
+
     LocalDate valuationDate = date(2011, 3, 8);
 
     // Build the trades from the node instruments
-    ObservableValues quotesMap = ObservableValues.ofIdMap(parRateData);
+    MarketData quotesMap = MarketData.builder().addValues(parRatesByKey).build();
```
jodastephen: Noting need for addValuesById again

modules/pricer/src/test/java/com/opengamma/strata/pricer/calibration/CalibrationZeroRateUsdEur2OisFxTest.java
```diff
@@ -298,10 +301,10 @@ private void calibration_market_quote_sensitivity_check(
     }
     double[] mqsEur2Computed = mqs.getSensitivity(EUR_DSC_CURVE_NAME, EUR).getSensitivity().toArray();
     for (int i = 0; i < EUR_DSC_NB_NODES; i++) {
-      Map<ObservableKey, Double> map = new HashMap<>(ALL_QUOTES.getValues());
+      Map<MarketDataKey<?>, Object> map = new HashMap<>(ALL_QUOTES.getValues());
       map.put(QuoteKey.of(StandardId.of(SCHEME, EUR_DSC_ID_VALUE[i])), EUR_DSC_MARKET_QUOTES[i] + shift);
-      ImmutableObservableValues ov = ImmutableObservableValues.of(map);
-      ImmutableRatesProvider rpShifted = calibrator.apply(ov);
+      ImmutableMarketData marketData = ImmutableMarketData.builder().values(map).build();
```
jodastephen: Looking at al the use cases that needed changing, adding a `MarketData.of(Map)` method looks worthwhile.

modules/basics/src/main/java/com/opengamma/strata/basics/market/MarketData.java
cjkent: It was a conscious decision to put it last. The core functionality of the interface is provided by the non-static methods so those are first.

examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
jodastephen: This is the kind of painful edge with ids and keys that we need to smooth over.

modules/pricer/src/test/java/com/opengamma/strata/pricer/calibration/CalibrationZeroRateUsdEur2OisFxTest.java
cjkent: I left it out because of time series. Do I need to provide all of these? I thought it might be better to just use the builder. 

    MarketData.of(Map<? extends MarketDataKey<?>, ?> values)
    MarketData.of(Map<? extends MarketDataKey<?>, ?> values, Map<? extends ObservableKey, LocalDateDoubleTimeSeries> timeSeries)
    MarketData.of(Map<? extends ObservableKey, LocalDateDoubleTimeSeries> timeSeries)

But there are enough cases of the first one that I've added it. We can add support for time series later if it looks like it's required.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableMarketData.java
cjkent: There's already a test. `ImmutableMarketDataTest`. Or am I misunderstanding?

examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
cjkent: As things stand `addValuesById` could only accept `ObservableId`. `ObservableId` can be converted to an `ObservableKey` but `MarketDataId` doesn't have a method (yet) to convert it to a `MarketDataKey`. It would be possible to add one but this is the only case I've seen so far where it would be useful. It's something to think about.

jodastephen: Yes, my thought was to use `ObservableId` here, maybe naming the method as such. Perhaps as ids and keys develop we may be able to improve, but I don't want to lose this method relative to what we had with `ObservableValues`

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableMarketData.java
jodastephen: I meant test serialiazation, as per `TestHelper.assertSerialization`

PR #565, opened 2015-11-17
Refactor ParRates and CurveNode to allow non-observable data

cjkent:
This PR refactors `ParRates` and `CurveNode` to use any type of market data instead of being limited to observable data. This is necessary to support calibration of curves using FX swaps as FX rates are not observable (cross rates are derived from observable data).


modules/basics/src/main/java/com/opengamma/strata/basics/market/ObservableKey.java
```diff
@@ -60,6 +60,10 @@
    * @param marketDataFeed  the market data feed that is the source of the observable market data
    * @return the identifier corresponding to this key
    */
-  ObservableId toObservableId(MarketDataFeed marketDataFeed);
+  public abstract ObservableId toObservableId(MarketDataFeed marketDataFeed);
 
+  @Override
+  public default MarketDataId<Double> toMarketDataId(MarketDataFeed marketDataFeed) {
```
jodastephen: Maybe this method could return `ObservableId`, avoiding the need for the other method.

cjkent: Good point

modules/market/src/main/java/com/opengamma/strata/market/curve/definition/FxSwapCurveNode.java
```diff
@@ -89,9 +92,10 @@ public DatedCurveParameterMetadata metadata(LocalDate valuationDate) {
 
   @Override
   public FxSwapTrade trade(LocalDate valuationDate, MarketData marketData) {
-    double fxNearRate = marketData.getValue(fxNearKey);
+    FxRate fxNearRate = marketData.getValue(fxNearKey);
+    double rate = fxNearRate.fxRate(fxNearKey.getPair());
```
jodastephen: This needs to use the convention's currency pair to extract the data to ensure that the rate is the right way round.

modules/market/src/main/java/com/opengamma/strata/market/curve/definition/XCcyIborIborSwapCurveNode.java
```diff
@@ -115,8 +118,9 @@ public DatedCurveParameterMetadata metadata(LocalDate valuationDate) {
   @Override
   public SwapTrade trade(LocalDate valuationDate, MarketData marketData) {
     double marketQuote = marketData.getValue(spreadKey) + additionalSpread;
-    double fxRate = marketData.getValue(fxKey);
-    return template.toTrade(valuationDate, BuySell.BUY, 1, fxRate, marketQuote);
+    FxRate fxRate = marketData.getValue(fxKey);
+    double rate = fxRate.fxRate(fxKey.getPair());
```
jodastephen: This needs to use the convention's currency pair to extract the data to ensure that the rate is the right way round.

modules/market/src/main/java/com/opengamma/strata/market/curve/definition/FxSwapCurveNode.java
cjkent: `FxRate` doesn't have a right and a wrong way. You have to pass in the currencies when you request the `double` rate in order to avoid any ambiguity.

It's designed that way to avoid getting in a terrible mess with scenarios. If there were two different rate objects representing the rate for a pair and its inverse then it would be very difficult to keep track. So there is only ever one rate object for a currency pair.

modules/pricer/src/main/java/com/opengamma/strata/pricer/calibration/MarketDataFxRateProvider.java
```diff
@@ -0,0 +1,39 @@
...
+import com.opengamma.strata.basics.currency.Currency;
+import com.opengamma.strata.basics.currency.FxRate;
+import com.opengamma.strata.basics.currency.FxRateProvider;
+import com.opengamma.strata.basics.market.FxRateKey;
+import com.opengamma.strata.basics.market.MarketData;
+
+/**
+ * A provider of FX rates that takes its data from a {@link MarketData} instance.
+ */
+class MarketDataFxRateProvider implements FxRateProvider {
```
jodastephen: `MarketData` could just implement `FxRateProvider`? While it makes it more specific, its not crazily more specific. It also allows other implementation possibilities, such as FX to be stored in an FxMatrix separate to the map of key to value.

cjkent: I'm not sure. The `MarketData` API is deliberately close to `CalculationEnvironment` and `CalculationMarketData` and is basically a map of key to value. I'm not sure it's worth adding it just to support this one case. If we find ourselves using this class all the time we can reconsider.

PR #567, opened 2015-11-17
Remove FX ticker from calibration CSV file

jodastephen:
No longer needed.


modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -86,17 +86,13 @@
  * <p>
  * The third file is the curve calibration nodes file.
  * This file has the following header row:<br />
- * {@code Curve Name,Label,Quote Symbology,Quote Ticker,Quote Field Name,FX Symbology,FX Ticker,FX Field Name,Type,Convention,Time,Spread}.
+ * {@code Curve Name,Label,Symbology,Ticker,Field Name,FX Symbology,FX Ticker,FX Field Name,Type,Convention,Time,Spread}.
```
cjkent: This line still refers to the FX columns.

PR #568, opened 2015-11-17
Basic support for building FX rates in the market data system

cjkent:
This PR adds basic support for building `FxRate` instances in the market data system using quoted rates. It currently only supports building rates directly from observable data, there is no support for cross rates yet.

It is not yet fully integrated. In order to use it the following steps are necessary:

* Pass an instance of `FxRateMarketDataFunction` to the constructor of `DefaultMarketDataFactory`
* Create `FxRatesConfig` mapping currency pairs to the `QuoteKey` identifying the rate and use it when creating `MarketDataConfig`
* Ensure the rate is available in the market data


jodastephen:
Merging, although could do with better docs for `defaultConfigs` in `MarketDataConfig`.

PR #572, opened 2015-11-18
Add functions and update examples

jodastephen:
Add functions for `IborFuture` and `DeliverableSwapFuture`.
Fix, test and update examples.


modules/function/src/main/java/com/opengamma/strata/function/calculation/index/AbstractIborFutureFunction.java
```diff
@@ -0,0 +1,96 @@
...
+   * 
+   * @return the pricer
+   */
+  protected DiscountingIborFutureTradePricer pricer() {
+    return DiscountingIborFutureTradePricer.DEFAULT;
+  }
+
+  @Override
+  public FunctionRequirements requirements(IborFutureTrade trade) {
+    IborFuture fra = trade.getProduct();
```
cjkent: `fra` doesn't seem like the right name for the product var

cjkent:
That's everything

PR #584, opened 2015-11-24
Current cash, currency exposure

yukiiwashita:


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/swap/DiscountingKnownAmountPaymentPeriodPricer.java
```diff
@@ -120,4 +121,17 @@ public void explainPresentValue(KnownAmountPaymentPeriod period, RatesProvider p
...
   }
 
+  //-------------------------------------------------------------------------
+  @Override
+  public MultiCurrencyAmount currencyExposure(KnownAmountPaymentPeriod period, RatesProvider provider) {
+    return MultiCurrencyAmount.of(CurrencyAmount.of(period.getCurrency(), presentValue(period, provider)));
+  }
+
+  @Override
+  public double currentCash(KnownAmountPaymentPeriod period, RatesProvider provider) {
```
marc-opengamma: As the amount is known, we don't need the full provider, the valuation date is enough.

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/swap/DiscountingNotionalExchangePricer.java
```diff
@@ -83,4 +84,18 @@ public void explainPresentValue(NotionalExchange event, RatesProvider provider,
...
   }
 
+  //-------------------------------------------------------------------------
+  @Override
+  public MultiCurrencyAmount currencyExposure(NotionalExchange event, RatesProvider provider) {
+    return MultiCurrencyAmount.of(CurrencyAmount.of(event.getCurrency(), presentValue(event, provider)));
+  }
+
+  @Override
+  public double currentCash(NotionalExchange event, RatesProvider provider) {
```
marc-opengamma: Same remark as above: As the amount is known, we don't need the full provider, the valuation date is enough.

modules/pricer/src/test/java/com/opengamma/strata/pricer/impl/swap/DiscountingFxResetNotionalExchangePricerTest.java
```diff
@@ -186,6 +189,60 @@ public void test_explainPresentValue_paymentDateInPast() {
   }
 
   //-------------------------------------------------------------------------
+  public void test_currencyExposure() {
```
marc-opengamma: Test currency exposure by finite difference: compare change of value of the product with change of value of the MultiCurrencyAmount currency exposure when the underlying exchange rate in FX_MATRIX is changed by a small amount.
The test should be done for a FxResetNotional which as not fixed and one which has fixed (i.e. the valuation date is between fixing and payment).

modules/pricer/src/test/java/com/opengamma/strata/pricer/impl/swap/DiscountingRatePaymentPeriodPricerTest.java
```diff
@@ -797,6 +799,52 @@ public void test_explainPresentValue_single_gearingSpread() {
   }
 
   //-------------------------------------------------------------------------
+  public void test_currencyExposure_fx() {
```
marc-opengamma: Same comment as for DiscountingFxResetNotionalExchangePricerTest:
Test currency exposure by finite difference: compare change of value of the product with change of value of the MultiCurrencyAmount currency exposure when the underlying exchange rate in FX_MATRIX is changed by a small amount.
The test should be done for a Payment which as not fixed and one which has fixed (i.e. the valuation date is between fixing and payment).

marc-opengamma:
I finished the first review. 
The only thing is for the payments with FX reset (notional and rate), I would like to see a "finite difference" test, both for the case when the fixing has not taken place and when he has taken place (i.e. the valuation date is between fixing and payment).

PR #593, opened 2015-11-25
Add label to curve node

jodastephen:
Ensure user-supplied label is used


modules/market/src/main/java/com/opengamma/strata/market/curve/node/FixedIborSwapCurveNode.java
```diff
@@ -56,6 +57,12 @@
    */
   @PropertyDefinition
   private final double additionalSpread;
+  /**
+   * The label to use for the node.
+   * If absent an appropriate default label will be used.
+   */
+  @PropertyDefinition(get = "optional")
```
cjkent: Why is label optional on the node but defaulted on the metadata? Is there any reason not to use a default value on the node? Is there any reason someone will want to query a node later to find out whether a label was specified?

modules/market/src/main/java/com/opengamma/strata/market/curve/node/FixedIborSwapCurveNode.java
```diff
@@ -80,7 +87,26 @@ public static FixedIborSwapCurveNode of(FixedIborSwapTemplate template, Observab
...
+  }
+
+  /**
+   * Returns a curve node for a Fixed-Ibor interest rate swap using the
+   * specified instrument template, rate key, spread and label.
+   *
+   * @param template  the template defining the node instrument
+   * @param rateKey  the key identifying the market data providing the rate for the node instrument
+   * @param additionalSpread  the additional spread amount added to the rate
+   * @param label  the label to use for the node, if null or empty an appropriate default label will be used
```
cjkent: Allowing a null value for the label feels a bit un-Strata. Particularly when there's an overload of the method that doesn't include the label. If you don't want to specify the label you can use the version with no label parameter, if you use the version with the label you should have to specify a valid label.

modules/market/src/main/java/com/opengamma/strata/market/curve/node/IborFutureCurveNode.java
```diff
@@ -56,6 +56,14 @@
    */
   @PropertyDefinition
   private final double additionalSpread;
+  /**
+   * The label to use for the node, may be empty.
+   * <p>
+   * If empty, a default label will be created when the metadata is built.
+   * The default label depends on the valuation date, so cannot be created in the node.
+   */
+  @PropertyDefinition(get = "private", validate = "notNull")
```
cjkent: Is this supposed to be private?

PR #595, opened 2015-11-25
Handle duplicate curve input keys when building a curve group

cjkent:
This fixes a bug where an exception was thrown when two curves in the same group used the same market data.

It also tightens up construction of `MarketEnvironment` and requires the valuation date to be specified.

`CurveCalibrator` has been converted to an interface to allow the curve group function to be tested without providing data to allow a real curve to be calibrated.

### DIFF
--- a/modules/function/src/main/java/com/opengamma/strata/function/marketdata/curve/CurveGroupMarketDataFunction.java
+++ b/modules/function/src/main/java/com/opengamma/strata/function/marketdata/curve/CurveGroupMarketDataFunction.java
@@ -11,6 +11,7 @@ import static java.util.stream.Collectors.toMap;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BinaryOperator;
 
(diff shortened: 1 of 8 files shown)


modules/function/src/main/java/com/opengamma/strata/function/marketdata/curve/CurveGroupMarketDataFunction.java
```diff
@@ -188,6 +203,29 @@ private CurveGroup buildGroup(
...
+   */
+  private static Object checkValuesEqual(Object value1, Object value2) {
+    if (value1.equals(value2)) {
+      return value1;
+    }
+    throw new IllegalArgumentException(
+        Messages.format(
+            "Values with the same key must be equal but found unequal values: {} and {}",
+            value1,
+            value2));
```
jodastephen: This error message does not contain the key, which will be painful to debug. While it is a pain, I think the calling code should be changed to not use streams in order to make the key available.

PR #605, opened 2015-11-26
Change interpolator/extrapolator methods

jodastephen:
Use `double`, not `Double`.
Remove some usages of `final`.
Some evidence of performance improvement following this change, perhaps 5%.

Mostly boring/automatic change.


modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/Interpolator1D.java
```diff
@@ -20,30 +20,26 @@
  * A base class for interpolation in one dimension.
  */
 public abstract class Interpolator1D
-    implements Interpolator<Interpolator1DDataBundle, Double>, Serializable {
+    implements Serializable {
```
cjkent: Does an interpolator really need to implement `Serializable`? Particularly given it's not a bean.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/Interpolator2D.java
```diff
@@ -13,7 +13,7 @@
 /**
  * A base class for two-dimensional interpolation.
  */
-public abstract class Interpolator2D implements Interpolator<Map<Double, Interpolator1DDataBundle>, DoublesPair> {
+public abstract class Interpolator2D {
```
cjkent: If the `Interpolator1D` actually needs to be serialisable then presumably this does too.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/data/ArrayInterpolator1DDataBundle.java
```diff
@@ -85,35 +86,32 @@ private static void swap(final double[] keys, final double[] values, final int f
...
   }
 
   @Override
-  public Double get(final Double key) {
-    final int index = Arrays.binarySearch(_keys, key);
+  public double get(double key) {
+    int index = Arrays.binarySearch(_keys, key);
     if (index < 0) {
-      return null;
+      return Double.NaN;
```
cjkent: Do we definitely want to be returning `NaN`? What's the expected behaviour in this case? Is it possible to produce a useful answer after returning `NaN`? If not then maybe it should throw an exception instead. `NaN` has a nasty habit of quietly spreading and infecting things and it can be hard to track down where it came from.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/data/InterpolationBoundedValues.java
```diff
@@ -13,14 +13,22 @@
...
 public final class InterpolationBoundedValues implements Serializable {
+
   private static final long serialVersionUID = 1L;
+
   private final int _lowerBoundIndex;
-  private final Double _lowerBoundKey;
-  private final Double _lowerBoundValue;
+  private final double _lowerBoundKey;
+  private final double _lowerBoundValue;
   private final Double _higherBoundKey;
```
cjkent: Why are two of the fields now `double` but the other two are still `Double`?

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/data/InterpolationBoundedValues.java
```diff
@@ -13,14 +13,22 @@
  * 
  */
 public final class InterpolationBoundedValues implements Serializable {
```
cjkent: `Serializable`?

cjkent:
That's everything

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/Interpolator1D.java
jodastephen: This PR leaves `Serializable` unchanged from before. I'll take a bigger look in the next commit.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/data/InterpolationBoundedValues.java
jodastephen: Those fields are optional. I didn't want to break too much, so left them as is. Its not the best design anyway...

jodastephen:
Updated

PR #610, opened 2015-11-27
Enhanced interpolation

jodastephen:
Move interpolator/extrapolator interface to strata-market
Separate from strata-math interpolator/extrapolator
Remove serializable where possible
Design for potential future enhancements


modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/Interpolator1DFactory.java
```diff
@@ -232,16 +229,16 @@ public static String getInterpolatorName(final Interpolator1D interpolator) {
     return s_instanceNames.get(interpolator.getClass());
   }
 
-  public static CurveInterpolator findCurveInterpolator(final String interpolatorName) {
+  public static Interpolator1D findCurveInterpolator(final String interpolatorName) {
```
cjkent: The name of this method clashes with the convention that `find*` should return an `Optional` and `get*` should throw an exception if the data isn't found.

modules/market/src/main/java/com/opengamma/strata/market/interpolator/ImmutableCurveInterpolator.java
```diff
@@ -0,0 +1,118 @@
...
+
+  //-------------------------------------------------------------------------
+  @Override
+  public BoundCurveInterpolator bind(
+      DoubleArray xValues,
+      DoubleArray yValues,
+      CurveExtrapolator extrapolatorLeft,
+      CurveExtrapolator extrapolatorRight) {
+
+    if (!(extrapolatorLeft instanceof ImmutableCurveExtrapolator)) {
```
cjkent: Is this `instanceof` check intended to be temporary?

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/DoubleQuadraticInterpolator1D.java
```diff
@@ -29,13 +27,10 @@
 @BeanDefinition(style = "light", constructorScope = "public")
 public final class DoubleQuadraticInterpolator1D
     extends Interpolator1D
-    implements CurveInterpolator, ImmutableBean, Serializable {
```
cjkent: Shouldn't this be `Serializable` given that it's a bean?

cjkent:
That's everything

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/DoubleQuadraticInterpolator1D.java
jodastephen: Not all beans need to be serializable. In this case, bean is being used just for convenience. All interpolator implementations no longer need to be serializable.

PR #614, opened 2015-11-30
Class to efficiently hold multiple amounts for multiple currencies

### LINKED ISSUE #586: Create MultiCurrencyValuesArray
This is an efficient way to store currency amounts in mixed currencies for multiple scenarios. It is likely to be very similar to `CurrencyValuesArray`.

`MultiCurrencyAmount` is a reasonably heavy object and it has been implicated in performance problems in a few cases. This should help.

cjkent:
This PR adds `MultiCurrencyValuesArray` which is the equivalent of `CurrencyValuesArray` for multiple currencies. It efficiently holds multiple currency amounts for large numbers of scenarios.

Fixes #586


modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
```diff
@@ -63,9 +64,42 @@
...
+   * Constructor that doesn't clone the input array to avoid unnecessary copies. The factory methods copy
+   * the input array where necessary.
+   *
+   * @param currency  the currency of the values
+   * @param values  the currency values
+   */
+  @ImmutableConstructor
+  private CurrencyValuesArray(
+      Currency currency,
+      double[] values) {
```
jodastephen: So throughout the rest of Strata, we now use `DoubleArray`, not `double[]` in beans. This provides a consistent safe wrapper, that limits unsafe method spread. While I understand that this class and the multi version are for performance, I still think `DoubleArray` is appropriate.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/MultiCurrencyValuesArray.java
```diff
@@ -0,0 +1,590 @@
...
+      throw new IllegalArgumentException(
+          Messages.format(
+              "Number of rates ({}) must be 1 or the same as the number of values ({})",
+              rateCount,
+              size));
+    }
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
jodastephen: Manual equals/hashCode are avoided by using `DoubleArray`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
jodastephen: BTW, the logic of the `convertedTo` method looks broken, as `rateCount` is allowed to be 1, but the stream is over a size that can be greater than one. (A `ScenarioMDBox` can have size one)

cjkent: A market data box with one value can be used with any number of scenarios. Think of that one value as the unperturbed value which is used in all scenarios. That's why there is a special case for 1.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
```diff
@@ -73,35 +84,35 @@ public CurrencyValuesArray convertedTo(Currency reportingCurrency, CalculationMa
     }
     MarketDataBox<FxRate> rates = marketData.getValue(FxRateKey.of(currency, reportingCurrency));
     checkNumberOfRates(rates.getScenarioCount());
-    double[] convertedValues = IntStream.range(0, values.length)
-        .mapToDouble(i -> rates.getValue(i).convert(values[i], currency, reportingCurrency))
+    double[] convertedValues = IntStream.range(0, values.size())
+        .mapToDouble(i -> rates.getValue(i).convert(values.get(i), currency, reportingCurrency))
```
jodastephen: you can use `values.mapWithIndex()` here, which avoids the unsafe below

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
```diff
@@ -63,6 +63,17 @@
    * @return an instance with the specified currency and values
    */
   public static CurrencyValuesArray of(Currency currency, double[] values) {
+    return new CurrencyValuesArray(currency, DoubleArray.copyOf(values));
+  }
```
jodastephen: I'd prefer to remove this factory method, which is what happened on other classes like curves. (Caller creates `DoubleArray`)

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/MultiCurrencyValuesArray.java
```diff
@@ -0,0 +1,537 @@
...
+  private static double[] currencyValuesArray(HashMap<Currency, double[]> mutableMap, Currency currency, int size) {
+    double[] currencyValues = mutableMap.get(currency);
+
+    if (currencyValues != null) {
+      return currencyValues;
+    }
+    double[] newArray = new double[size];
+    mutableMap.put(currency, newArray);
+    return newArray;
+  }
```
jodastephen: `Map.computeIfAbsent()` ?

PR #619, opened 2015-12-02
Refactor INI configuration file loading

jodastephen:
Simpler design, Avoid problems in Uber-Jars, Better logging of errors.

Logic for INI file merging centralized in new class `ResourceConfig`, which manages the files on the classpath. Chaining is de-emphasised, although still possible to do.


jonathansenior:
One use-case which we've discussed before is how additional conventions could be supplied dynamically - e.g. from a database, or loaded from files. The application using Strata would create the objects representing these additional conventions, but would need a way to inject them. The ini file mechanism has the ability to reference code, but without any arguments it seems limited to static extensions. Is there a way that extensions could be registered dynamically?

jodastephen:
I'm considering a CSV format for some of the conventions that would allow files to define new instances. It all depends on whether a static constant is needed for developers (because in that case you need to write code anyway)

The database case is more complex, particularly as Strata doesn't do databases in general. A provider can be written to do it, but a general purpose one probably won't work.

jonathansenior:
I agree that if a static constant was needed then you'd be writing code. I'm thinking of the case where you want to register some new conventions so that they can be used by name elsewhere (e.g. in a trade or curve definition file). I don't think Strata needs to care where these additional conventions come from - the application that wants to supply them might pull them out of a database or load them from some CSV format - but this application would need a simple mechanism to register them with Strata.

cjkent:
My current use case requires holidays to be loaded in response to a request by the user. I envisaged creating a `NamedLookup` implementation to store them. But both my code and the framework code need access to it. There are two obvious options:

1. The extend enum code creates the `NamedLookup` and exposes it to application code
2. The application code creates the `NamedLookup` and adds it to the extended enum mechanism

Are either of those currently possible? Or is there another way to achieve what I want?

Option 2 feels like the better one. If the `NamedLookup` requires a lot of configuration (e.g. for database access) it would be better to leave it to the application to configure it in whatever way it chooses. Otherwise we need to create a configuration mechanism to cover every eventuality.

jonathansenior:
Option 2 would solve the use case I am describing as well. In fact, it's the same use case.

jodastephen:
Firstly, this discussion is separate to this PR, whose primary goal to to ensure config files don't clash in uber-jars.

Secondly, while extended enums can in theory be changed at runtime, that has never been their primary use case. Way back when the concept was setup, the limitation was also made clear, that the design is primarily for things that get loaded when the system starts. All the setup is done in static initializers.

Providing a `NamedLookup` instance that queries a CSV file on statup is easy. Users would add their additional config to the CSV file and start/restart the system.

Providing a `NamedLookup` instance that queries a database is not too hard (apart from configuring the database). However, it needs to be a one-off query run from the static initializer. Running a database query every time a lookup is performed would be too slow. (As such, I think CSV is better than a database).

Providing a `NamedLookup` instance that has a mutable static hash map is also possible, allowing items to be changed on the fly. It obviously has to be written carefully to watch concurrency and performance.

The key however is that the set of `NamedLookup` instances is fixed when the static initializer runs.

(The original design before extended enums had mutable static state, but got vetoed as a design).

PR #621, opened 2015-12-02
Collector for creating MultiCurrencyValuesArray from MultiCurrencyAmount

### LINKED ISSUE #620: CurrencyValuesArray.get should return CurrencyAmount
`CurrencyValuesArray.get` returns a double representing the currency amount. 

If you have a reference to the `CurrencyValuesArray` you can get the currency. But the individual values are extracted using `get` when running calculations for a single scenario and the currency is lost.

cjkent:
This PR adds a collector implementation to `FunctionUtils` which creates `MultiCurrencyValuesArray` from streams of `MultiCurrencyAmount`. This is much more compact than having a `MultiCurrencyAmount` for every scenario and allows the amounts to be accessed as primitive double arrays for greater performance.

Every engine function which calculates `MultiCurrencyAmount` has been modified to use the new collector and to return `MultiCurrencyValuesArray`.

Fixes #620


modules/function/src/main/java/com/opengamma/strata/function/calculation/fx/FxNdfMultCurrencyAmountFunction.java
```diff
@@ -0,0 +1,34 @@
...
+import com.opengamma.strata.calc.runner.DefaultSingleCalculationMarketData;
+import com.opengamma.strata.calc.runner.function.result.ScenarioResult;
+import com.opengamma.strata.function.marketdata.MarketDataRatesProvider;
+import com.opengamma.strata.product.fx.ExpandedFxNdf;
+import com.opengamma.strata.product.fx.FxNdfTrade;
+
+/**
+ * Function which calculates values of a measure for an {@link FxNdfTrade} and returns a {@link MultiCurrencyAmount}.
+ */
+public abstract class FxNdfMultCurrencyAmountFunction extends AbstractFxNdfFunction<MultiCurrencyAmount> {
```
jodastephen: Class name is wrong. It also has a different naming order to `MultiCurrencyAmountFxSingleFunction`, not sure which way around works best.

PR #624, opened 2015-12-02
ScenarioMarketDataValue backed by a DoubleArray

### LINKED ISSUE #618: ScenarioMarketDataValue implementation backed by double array
This is one of the most fundamental types of scenario data - one double value per scenario. We need an efficient way to store it. The only option at the moment is `List<Double>` which is inadequate.

cjkent:
Adds an implementation of `ScenarioMarketDataValue` for storing quotes for multiple scenarios. It is backed by a `DoubleArray` so it is space efficient and provides access to the underlying `double[]` for maximum performance.

The only potentially controversial change was moving `ScenarioMarketDataKey` and `MarketDataBox` into the `strata-basics` module. This was necessary because the `strata-calc` module refers to `ScenarioMarketDataKey` in `CalculationMarketData` but key implementations need to be in `strata-market`. The only module accessible to both calc and market is basics. `MarketDataBox` needed to move too because `ScenarioMarketDataKey` refers to it.

`MarketDataKey` is already in basics so putting `ScenarioMarketDataKey` there seems like the right thing to do. Moving `MarketDataBox` feels like less of a sure thing, but I don't have a problem with it. It's a pretty low-level concept.

Fixes #618


modules/market/src/main/java/com/opengamma/strata/market/key/QuotesArrayKey.java
```diff
@@ -0,0 +1,420 @@
...
+
+  @Override
+  public QuotesArray createScenarioValue(MarketDataBox<Double> marketDataBox) {
+    // TODO This suggests MarketDataBox needs a stream() method or at least a helper to create a stream from a box
+    double[] quotes = new double[marketDataBox.getScenarioCount()];
+
+    for (int i = 0; i < marketDataBox.getScenarioCount(); i++) {
+      quotes[i] = marketDataBox.getValue(i);
+    }
+    return QuotesArray.of(DoubleArray.ofUnsafe(quotes));
```
jodastephen: return QuotesArray.of(DoubleArray.of(marketDataBox.getScenarioCount(), i -> marketDataBox.getValue(i)));

modules/market/src/main/java/com/opengamma/strata/market/key/QuotesArrayKey.java
```diff
@@ -0,0 +1,420 @@
...
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final StandardId standardId;
+
+  /**
+   * The field name in the market data record that is required.
+   * For example, {@link FieldName#MARKET_VALUE}.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final FieldName fieldName;
```
jodastephen: Think it would make more sense if it held a `QuoteKey` instead of the two parts of a quote key. After all, its the same concept, just a wrapper for a different storage.

modules/market/src/main/java/com/opengamma/strata/market/key/QuotesArrayKey.java
```diff
@@ -0,0 +1,420 @@
...
+import com.opengamma.strata.basics.market.MarketDataBox;
+import com.opengamma.strata.basics.market.ScenarioMarketDataKey;
+import com.opengamma.strata.collect.array.DoubleArray;
+import com.opengamma.strata.collect.id.StandardId;
+import com.opengamma.strata.market.value.QuotesArray;
+
+/**
+ * A key identifying a {@link QuotesArray} containing values for a piece of quoted market data in multiple scenarios.
+ */
+@BeanDefinition
```
jodastephen: `builderScope = "private"`

modules/market/src/main/java/com/opengamma/strata/market/value/QuotesArray.java
```diff
@@ -0,0 +1,328 @@
...
+import org.joda.beans.impl.direct.DirectMetaProperty;
+import org.joda.beans.impl.direct.DirectMetaPropertyMap;
+
+import com.opengamma.strata.basics.market.ScenarioMarketDataValue;
+import com.opengamma.strata.collect.array.DoubleArray;
+
+/**
+ * Values for an item of quoted market data in multiple scenarios.
+ */
+@BeanDefinition
```
jodastephen: `builderScope = "private"`

cjkent:
BTW, the only real changes are the four new classes. Everything else is automated refactoring from the moves.

modules/market/src/main/java/com/opengamma/strata/market/key/QuotesArrayKey.java
cjkent: Nice, I clearly need to get more familiar with `DoubleArray`

modules/calc/src/test/java/com/opengamma/strata/calc/marketdata/scenario/SingleMarketDataBoxTest.java
```diff
@@ -10,6 +10,8 @@
 
 import org.testng.annotations.Test;
 
+import com.opengamma.strata.basics.market.MarketDataBox;
```
jodastephen: Looks like some tests need to move to basics

modules/market/src/main/java/com/opengamma/strata/market/value/QuotesArray.java
```diff
@@ -0,0 +1,328 @@
...
+  @PropertyDefinition(validate = "notNull")
+  private final DoubleArray values;
+
+  /**
+   * Returns an instance wrapping a set of quotes.
+   *
+   * @return an instance wrapping a set of quotes
+   */
+  public static QuotesArray of(DoubleArray quotes) {
+    return new QuotesArray(quotes);
```
jodastephen: quotes or values?

jodastephen: Also missing param Javadoc

modules/market/src/main/java/com/opengamma/strata/market/key/QuotesArrayKey.java
```diff
@@ -0,0 +1,420 @@
...
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final FieldName fieldName;
+
+  /**
+   * Returns a key identifying the market data with the specified ID and field name.
+   *
+   * @return a key identifying the market data with the specified ID and field name
+   */
+  public static QuotesArrayKey of(StandardId id, FieldName fieldName) {
```
jodastephen: Missing param Javadoc on both factory methods

PR #628, opened 2015-12-03
Load index data from CSV file

jodastephen:
Allows indices to be added without code.
Add normalize() method to DaysAdjustment.


modules/basics/src/main/java/com/opengamma/strata/basics/index/FxIndexCsvLookup.java
```diff
@@ -0,0 +1,101 @@
...
+    Map<String, FxIndex> map = new HashMap<>();
+    for (ResourceLocator resource : resources) {
+      try {
+        CsvFile csv = CsvFile.of(resource.getCharSource(), true);
+        for (int i = 0; i < csv.rowCount(); i++) {
+          FxIndex parsed = parseFxIndex(csv, i);
+          map.put(parsed.getName(), parsed);
+        }
+      } catch (RuntimeException ex) {
+        log.severe("Error processing resource as FX Index CSV file: " + resource + ex.getMessage());
```
cjkent: This will produce an odd error message. Could use `log(Level.SEVERE, "..." + resource, ex);`

cjkent: That's it form me - there are 3 or 4 instances of the same logging pattern.

PR #632, opened 2015-12-08
Avoid locale dependent behaviour

jodastephen:


modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
```diff
@@ -76,6 +76,11 @@
  */
 public class TestHelper {
 
+  /**
+   * UTF-8 encoding name.
+   */
+  private static final String UTF_8 = "UTF-8";
```
cjkent: Wouldn't it be better to use `StandardCharsets.UTF_8.name()`?

PR #635, opened 2015-12-08
Add holiday data file

jodastephen:
A flexible mechanism to add new holiday calendars, which could also be used to replace built in ones. No change is made to the existing algorithmic calendar generation, as it is effective and minimal.

INI and property files now allow a shorthand format where `key = value` can be replaced by `key`, which sets the value to an empty string.

`TestHelper` was enhanced to allow logging to be captured.


modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
```diff
@@ -372,7 +376,7 @@ public static synchronized String caputureSystemOut(Runnable runner) {
    * <p>
    * For example:
    * <pre>
-   *  String sysErr = caputureSystemerr(() -> myCode);
+   *  String sysErr = caputureSystemErr(() -> myCode);
```
cjkent: Typo

modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
```diff
@@ -395,6 +399,55 @@ public static synchronized String caputureSystemErr(Runnable runner) {
...
+   * For example:
+   * <pre>
+   *  String log = caputureLog(Foo.class, () -> myCode);
+   * </pre>
+   * 
+   * @param loggerClass  the class defining the logger to trap
+   * @param runner  the lambda containing the code to test
+   * @return the captured output
+   */
+  public static synchronized List<LogRecord> caputureLog(Class<?> loggerClass, Runnable runner) {
```
cjkent: Typo in method name.

cjkent:
The typo is the only problem

jodastephen:
Fixed typo

PR #642, opened 2015-12-10
Refactor CurveGroupDefinition to make curve definitions optional

cjkent:
Remove `NodalCurveDefinition` from `CurveGroupEntry` and add a map of curve definitions to `CurveGroupDefinition`. This has the effect of making curve definitions optional and allows curve group definitions to be used with curves that are created outside Strata.

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinition.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinition.java
@@ -166,7 +166,7 @@ public final class CurveGroupDefinition
    * @param curveName  the name of the curve
    * @return the definition for the curve with the specified name
    */
-  public Optional<NodalCurveDefinition> findDefinition(CurveName curveName) {
+  public Optional<NodalCurveDefinition> findCurveDefinition(CurveName curveName) {
     return Optional.ofNullable(curveDefinitionsByName.get(curveName));
   }
 
--- a/modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinition.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinition.java
@@ -22,6 +22,7 @@ import org.joda.beans.BeanBuilder;
 import org.joda.beans.BeanDefinition;
 import org.joda.beans.ImmutableBean;
 import org.joda.beans.ImmutableConstructor;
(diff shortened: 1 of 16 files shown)


modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinition.java
```diff
@@ -100,27 +119,57 @@ public static CurveGroupDefinition of(CurveGroupName name, Collection<CurveGroup
...
 
+  /**
+   * Finds the definition for the curve with the specified name.
+   * <p>
+   * If the curve is not found, optional empty is returned.
+   *
+   * @param curveName  the name of the curve
+   * @return the definition for the curve with the specified name
+   */
+  public Optional<NodalCurveDefinition> findDefinition(CurveName curveName) {
```
jodastephen: Field is `curveDefinitions` but this method is just `definition`. Should stick to one name.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupDefinitionBuilder.java
```diff
@@ -65,7 +68,32 @@ public CurveGroupDefinitionBuilder addDiscountCurve(
...
+   * <p>
+   * A curve added with this method cannot be calibrated by the market data system as it does not include
+   * a curve definition. It is intended to be used with curves which are supplied by the user.
+   *
+   * @param curveName  the name of the curve
+   * @param otherCurrencies  additional currencies for which the curve can provide discount factors
+   * @param currency  the currency for which the curve provides discount rates
+   * @return this builder
+   */
+  public CurveGroupDefinitionBuilder addDiscountCurve(
```
jodastephen: These new methods feel wrong named `addXxxCurve`. Perhaps `addXxxCurveName` or `addXxxCurveUsage`

cjkent: I can see your point, but I think both of those alternative names actually make it more confusing. I think `addXxxCurve` communicates what's happening pretty well. You're telling the curve group that it contains a curve with a specified name. That feels like an `add` operation to me.

If the other methods were actually adding `Curve` objects I'd see the need to change it, but they are only adding a curve definition so I don't think the distinction is that great.

I think the the current names aren't perfect but I think they're the lesser of two evils.

PR #646, opened 2015-12-11
Improve performance of linear interpolation

jodastephen:
Rename getIndex() to getValue() and add getKey()
Avoid InterpolationBoundedValues object construction

### DIFF
--- a/modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/LinearInterpolator1D.java
+++ b/modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/LinearInterpolator1D.java
@@ -14,10 +14,8 @@ import org.joda.beans.MetaBean;
 import org.joda.beans.Property;
 import org.joda.beans.impl.light.LightMetaBean;
 
-import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.math.impl.MathException;
 import com.opengamma.strata.math.impl.interpolation.data.ArrayInterpolator1DDataBundle;
-import com.opengamma.strata.math.impl.interpolation.data.InterpolationBoundedValues;
 import com.opengamma.strata.math.impl.interpolation.data.Interpolator1DDataBundle;
 
 /**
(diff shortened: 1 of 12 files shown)


modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/LinearInterpolator1D.java
```diff
@@ -34,56 +32,51 @@
...
   @Override
   public double firstDerivative(Interpolator1DDataBundle data, double value) {
-    JodaBeanUtils.notNull(data, "data");
-    InterpolationBoundedValues boundedValues = data.getBoundedValues(value);
-    double x1 = boundedValues.getLowerBoundKey();
-    double y1 = boundedValues.getLowerBoundValue();
-    if (data.getLowerBoundIndex(value) == data.size() - 1) {
+    int lowerIndex = data.getLowerBoundIndex(value);
+    if (lowerIndex == data.size() - 1) {
+      if (data.size() == 1) {
```
cjkent: Couldn't this `if` block move before the other `if` block? If the data has a size of 1 then the lower index will always be `data.size() - 1`.

jodastephen: Yes, but it would be slower. Given a typical curve has more than one node, and a typical lookup is not after the last node, it is better to catch both cases with a single if, rather than two.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/LinearInterpolator1D.java
```diff
@@ -34,56 +32,51 @@
...
+      }
       if (value > data.lastKey()) {
-        throw new MathException("Value of " + value + " after last key. Use exstrapolator");
+        throw new MathException("Value of " + value + " after last key. Use extrapolator");
       }
-      double[] x = data.getKeys();
-      double[] y = data.getValues();
-      int n = x.length;
-      return n == 1 ? 0.0 : (y[n - 1] - y[n - 2]) / (x[n - 1] - x[n - 2]);
+      lowerIndex--;
```
cjkent: A quick comment explaining the logic would save a lot of head scratching here.

cjkent:
That's everything

PR #651, opened 2015-12-16
Created a new type of curve based on the sum of a fixed curve and a spread curve

marc-opengamma:
This type of curve is used in particular for interest rate curves when one wants to keep a curve shape but compute sensitivities to a different set of points or with a different interpolator.


modules/market/src/main/java/com/opengamma/strata/market/curve/AddFixedCurve.java
```diff
@@ -0,0 +1,425 @@
...
+
+/**
+ * A curve described as the sum of two other curves.
+ * <p>
+ * Of the two curves composition this curve, one is considered as fixed. It impacts the shape of the total curve
+ * but not its number of parameters. The sensitivity and perturbation of the total curve is coming from change
+ * on the non-fixed curve, also called spread curve.
+ */
+
+@BeanDefinition
```
cjkent: Can have `builderScope = "private"` as the `of` method is sufficient.

modules/market/src/main/java/com/opengamma/strata/market/curve/AddFixedCurve.java
```diff
@@ -0,0 +1,425 @@
...
+  
+  /**
+   * The spread curve. Also called the variable curve.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final Curve spreadCurve;
+
+  //-------------------------------------------------------------------------
+  /**
+   * Creates a curve ass the sum of a fixed curve and a spread curve.
```
cjkent: s/ass/as/

PR #653, opened 2015-12-16
Allow duplicate curve names in CurveGroup.ofCurves

cjkent:
This change allows `CurveGroup.ofCurves` to accept a list of curves which contains duplicates. Previously an exception was thrown if there were multiple curves in the list with the same name.

It turns out a common use case for this method is passing in a list created from the curves in another curve group. If the same curve is used for discounting and a rates index it can appear twice in the list.

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCurvesCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCurvesCsvLoader.java
@@ -126,10 +126,9 @@ public final class RatesCurvesCsvLoader {
 
     // Ensure curve names are unique
     for (Curve curve : curves) {
-      if (curveNames.contains(curve.getName())) {
+      if (!curveNames.add(curve.getName())) {
         throw new IllegalArgumentException("Multiple curves with the same name: " + curve.getName());
       }
-      curveNames.add(curve.getName());
     }
     return curveGroups.stream().map(groupDef -> CurveGroup.ofCurves(groupDef, curves)).collect(toImmutableList());
   }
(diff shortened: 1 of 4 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCurvesCsvLoader.java
```diff
@@ -120,6 +122,15 @@
     List<CurveGroupDefinition> curveGroups = CurveGroupDefinitionCsvLoader.loadCurveGroups(groupsResource);
     Multimap<LocalDate, Curve> allCurves = loadCurves(settingsResource, curvesResources, marketDataDate);
     Collection<Curve> curves = allCurves.get(marketDataDate);
+    Set<CurveName> curveNames = new HashSet<>();
+
+    // Ensure curve names are unique
+    for (Curve curve : curves) {
+      if (curveNames.contains(curve.getName())) {
```
jodastephen: `if (!curveNames.add(curve.getName())) {`

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroup.java
```diff
@@ -80,7 +80,10 @@ public static CurveGroup of(CurveGroupName name, Map<Currency, Curve> discountCu
   /**
    * Creates a curve group using a curve group definition and some existing curves.
    * <p>
-   * The curves must include all curves named in the curve group definition.
+   * If there are curves named in the definition which are not present in the curves the group is built using
+   * whatever curves are available.
+   * <p>
+   * If there are multiple curves with the same name in the curves it is undefined which one is used.
```
jodastephen:
```
   * If there are curves named in the definition that are not present in the input curves,
   * the group is built using whatever curves are available.
   * <p>
   * If there are multiple curves with the same name in the set of input curves, it is undefined which one is used.
```

I can't say I like the "undefined" part of this, but unless we pass a `Map<CurveName, Curve>` in (and validate it) things could get messy.

Issue #545, opened 2015-11-10
CMS coupon and cap/floor - pricer by replication


marc-opengamma:
Create a pricer by replication (from swaption) for CMS related product. 
The underlying swaption pricer should have a control on the volatility wings (high strikes) as this has an impact for CMS pricing. 
See "CapFloorCMSSABRExtrapolationRightReplicationMethod" in Platform 2.x code and documentation OG 15 and OG 11 for the quant description.

PR #657, opened 2015-12-18
Swap index

yukiiwashita:
* This work is required by #545.


modules/product/src/main/java/com/opengamma/strata/product/swap/SwapIndex.java
```diff
@@ -0,0 +1,354 @@
...
+import org.joda.beans.impl.direct.DirectMetaPropertyMap;
+
+import com.opengamma.strata.collect.named.Named;
+import com.opengamma.strata.product.swap.type.FixedIborSwapTemplate;
+
+/**
+ * Class describing a swap index.
+ */
+@BeanDefinition(builderScope = "private")
+public final class SwapIndex
```
jodastephen: I think `SwapIndex` should follow the same pattern as `IborIndex`, with an interface, implementation, use of extended enums and a CSV loader.

yukiiwashita:
Closed by #661

PR #660, opened 2015-12-18
Refactor interpolators

jodastephen:
Port three interpolator implementations.
Remove `Exponential` interpolator as it is the same as `LogLinear`

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/interpolator/DoubleQuadraticCurveInterpolator.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/interpolator/DoubleQuadraticCurveInterpolator.java
@@ -1,0 +1,255 @@
+/**
+ * Copyright (C) 2015 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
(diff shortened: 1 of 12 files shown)


modules/market/src/main/java/com/opengamma/strata/market/interpolator/DoubleQuadraticCurveInterpolator.java
```diff
@@ -0,0 +1,255 @@
...
+  static class Bound extends AbstractBoundCurveInterpolator {
+    private final double[] xValues;
+    private final double[] yValues;
+    private final int intervalCount;
+    private final RealPolynomialFunction1D[] quadratics;
+    private final RealPolynomialFunction1D[] quadraticsFirstDerivative;
+
+    Bound(DoubleArray xValues, DoubleArray yValues) {
+      super(xValues, yValues);
+      this.xValues = xValues.toArrayUnsafe();
```
cjkent: Are you sure you want `toArrayUnsafe` here? They are passed in from the `bind` method and come from the outside world without any cloning.

PR #663, opened 2015-12-22
Tidy swaption pricers

jodastephen:
Add LongShort.sign()
Use methods on PutCall and PayReceive
Make volatilities factory methods consistent with Normal

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/LongShort.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/LongShort.java
@@ -23,11 +23,20 @@ public enum LongShort {
   /**
    * Long.
    */
-  LONG,
+  LONG(1),
   /**
    * Short.
    */
-  SHORT;
+  SHORT(-1);
+
+  /**
+   * True if long, used to avoid a branch.
+   */
+  private final boolean isLong;
+  /**
+   * The sign, used to avoid a branch.
+   */
+  private final int sign;
 
   //-------------------------------------------------------------------------
   /**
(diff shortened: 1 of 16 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/LongShort.java
```diff
@@ -23,11 +23,20 @@
...
+  SHORT(-1);
+
+  /**
+   * True if long, used to avoid a branch.
+   */
+  private final boolean isLong;
+  /**
+   * The sign, used to avoid a branch.
+   */
+  private final int sign;
```
jonathansenior: Would `multiplier` be more appropriate?

modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/NormalSwaptionPhysicalProductPricer.java
```diff
@@ -180,13 +177,13 @@ public double impliedVolatilityFromPresentValue(
       double presentValue) {
 
     ExpandedSwaption expanded = swaption.expand();
-    double sign = expanded.getLongShort() == LongShort.LONG ? 1d : -1d;
+    double sign = expanded.getLongShort().sign();
     ArgChecker.isTrue(presentValue * sign > 0, "Present value sign must be in ligne with the option Long/Short flag ");
```
jonathansenior: Someone switched into French part-way through this error message.

modules/basics/src/main/java/com/opengamma/strata/basics/LongShort.java
jonathansenior: On second thoughts, looking at where this is used, `sign` is fine.

PR #664, opened 2015-12-22
Create new calibration example with 3 leg basis swap and discount factors

marc-opengamma:
New example of curve calibration.
Uses basis swaps with 3 legs (standard in EUR) to calibrate one of the Ibor Index curve.
The curve interpolation is done on discount factor with 'LogNaturalCubicDiscountFactor' interpolator.


examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationEur3CheckExample.java
```diff
@@ -0,0 +1,259 @@
...
+public class CalibrationEur3CheckExample {
+
+  /**
+   * The valuation date.
+   */
+  private static final LocalDate VAL_DATE = LocalDate.of(2015, 11, 20);
+  /**
+   * The empty time-series.
+   */
+//  private static final LocalDateDoubleTimeSeries TS_EMTPY = LocalDateDoubleTimeSeries.empty();
```
yukiiwashita: Delete rather than comment out.

examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationEur3CheckExample.java
```diff
@@ -0,0 +1,259 @@
...
+    if (args.length > 0) {
+      if (args[0].equals("-p")) {
+        performance_calibration_pricing();
+      }
+    }
+    System.out.println("Checked PV for all instruments used in the calibration set are near to zero");
+  }
+
+  // Example of performance: loading data from file, calibration and PV
+  public static void performance_calibration_pricing() {
```
yukiiwashita: Add java doc or set this private.

PR #675, opened 2016-01-04
Modified IborFixingDeposit pricer to work when the day fixing is present in the provider.

marc-opengamma:
The IborFixingDeposit pricer has been modified recently, causing problem in curve calibration when the fixing of the day is present in the time series. The test on that feature unfortunately contained an error, passing the wrong date for the fixing.
This PR should reinstalled the initial feature, correct the test in the pricer, extend the tests to other methods and add a calibration test passing a fixing time series with the fixing of the day.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/deposit/DiscountingIborFixingDepositProductPricer.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/deposit/DiscountingIborFixingDepositProductPricer.java
@@ -8,9 +8,12 @@ package com.opengamma.strata.pricer.deposit;
 import com.opengamma.strata.basics.currency.Currency;
 import com.opengamma.strata.basics.currency.CurrencyAmount;
 import com.opengamma.strata.basics.index.IborIndex;
+import com.opengamma.strata.collect.timeseries.LocalDateDoubleTimeSeries;
+import com.opengamma.strata.market.curve.Curve;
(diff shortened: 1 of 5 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/deposit/DiscountingIborFixingDepositProductPricer.java
```diff
@@ -125,14 +128,18 @@ public PointSensitivities parSpreadSensitivity(IborFixingDepositProduct product,
   // query the forward rate
   private double forwardRate(ExpandedIborFixingDeposit product, ImmutableRatesProvider provider) {
     IborIndex index = product.getFloatingRate().getIndex();
-    IborIndexRates rates = provider.iborIndexRates(index);
+    Curve curve = provider.getIndexCurves().get(index);
+    IborIndexRates rates = DiscountIborIndexRates.of(index, LocalDateDoubleTimeSeries.empty(),
```
jonathansenior: After a bit of thinking, I can see what this change does - i.e. by passing in an empty fixing series, the rate will always be sourced from the forward curve, whereas previously it implicitly uses the fixing, if available, in preference to the forward curve. This is quite subtle, so I think it would be worth a comment here to explain this to others.

marc-opengamma:
After more review, this does not seems the best way to achieve the objective. I will close this PR and reopen a new one with a slightly different approach (and more explanations on why it is done that way).

PR #676, opened 2016-01-04
Allow threads to be closed in runner factory

### LINKED ISSUE #674: CalculationRunnerFactory leaks threads
The methods `CalculationRunnerFactory.ofSingleThreaded` and `ofMultiThreaded` create a fixed size thread pool. The factory interface doesn't have a method to dispose of the factory and it doesn't expose the thread pool. 

This means the threads in the pool can never be shut down and the threads will exist until the JVM shuts down.

jodastephen:
Use lambda to capture common case
Fixes #674


examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
```diff
@@ -202,9 +202,11 @@ private static void performance_calibration_pricing() {
         .build();
 
     // calculate the results
-    CalculationRunner runner = CalculationRunnerFactory.ofSingleThreaded()
-        .createWithMarketDataBuilder(trades, columns, rules, marketDataFactory(), marketDataConfig);
-    return Pair.of(trades, runner.calculateSingleScenario(marketSnapshot));
+    return CalculationRunnerFactory.multiThreaded(factory -> {
```
jonathansenior: It feels to me like this call tries to do too much. Something like the following would be more explicit and potentially clearer:

```
try (CalculationRunner runner = CalculationRunnerFactory.createMultiThreaded()) {
  runner.calculateSingleScenario(...)
}
```

This implies the `CalculationRunner` itself being the thing that's `AutoCloseable`, which is what I would more naturally expect than the factory.

cjkent: The factory needs to be closable because it's the factory that owns the thread pool. If the runners are closable it leaves us with the same problem with have already - the thread pool in the factory is never shut down.

jodastephen:
Updated following discussion

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRunner.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  public static CalculationRunner of(ExecutorService executor) {
+    return DefaultCalculationRunner.of(executor);
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Performs calculations for a single set of market data.
+   * <p>
+   * This returns a grid of results based on the specified tasks and market data.
```
cjkent: The comment still refers to tasks but there are no task parameters any more.

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRunner.java
```diff
@@ -0,0 +1,174 @@
...
+  public abstract Results calculateSingleScenario(
+      List<? extends CalculationTarget> targets,
+      List<Column> columns,
+      CalculationRules calculationRules,
+      CalculationEnvironment marketData);
+
+  /**
+   * Performs calculations for multiple scenarios, each with a different set of market data.
+   * <p>
+   * This returns a grid of results based on the specified tasks and market data.
```
cjkent: Tasks

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRunner.java
```diff
@@ -0,0 +1,174 @@
...
+   * 
+   * @param targets  the targets for which values of the measures will be calculated
+   * @param columns  the configuration for the columns that will be calculated, including the measure and
+   *   any column-specific overrides
+   * @param calculationRules  the rules defining how the calculation is performed
+   * @param marketData  the market data used in the calculations
+   * @return the results of running the calculations in the view for every item in the portfolio and every scenario
```
cjkent: Comment includes reference to "the view".

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRunner.java
```diff
@@ -0,0 +1,174 @@
...
+      CalculationEnvironment marketData,
+      CalculationListener listener);
+
+  //-------------------------------------------------------------------------
+  /**
+   * Gets the underlying task runner.
+   * <p>
+   * In most cases, this runner will be implemented using an instance of {@link CalculationTaskRunner}.
+   * That interface provides a lower-level API, with the ability optimize if similar calculations
+   * are being made repeatedly. In addition, the lower level API provides the ability to determine
```
cjkent: The lower level API isn't actually necessary to determine the requirements, you only need to create an instance of `CalculationTasks` and call `getRequirements`.

modules/calc/src/main/java/com/opengamma/strata/calc/DefaultCalculationRunner.java
```diff
@@ -0,0 +1,129 @@
...
+import com.opengamma.strata.calc.runner.CalculationListener;
+import com.opengamma.strata.calc.runner.CalculationTaskRunner;
+import com.opengamma.strata.calc.runner.CalculationTasks;
+import com.opengamma.strata.calc.runner.Results;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * The default calculation runner.
+ * <p>
+ * This uses a single instance of {@link ExecutorService}.
```
cjkent: This isn't true any more, it uses a `CalculationTaskRunner` instead.

modules/calc/src/main/java/com/opengamma/strata/calc/DefaultCalculationRunner.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+class DefaultCalculationRunner implements CalculationRunner {
+
+  /**
+   * Executes the tasks that perform the individual calculations.
+   * This will typically be multi-threaded, but single or direct executors also work.
```
cjkent: This comment still appears to refer to the executor.

modules/calc/src/main/java/com/opengamma/strata/calc/DefaultCalculationRunner.java
```diff
@@ -0,0 +1,129 @@
...
+   * It is recommended to use try-with-resources to manage the runner:
+   * <pre>
+   *  try (DefaultCalculationRunner runner = DefaultCalculationRunner.ofMultiThreaded()) {
+   *    // use the runner
+   *  }
+   * </pre>
+   * 
+   * @return the calculation runner
+   */
+  public static DefaultCalculationRunner ofMultiThreaded() {
```
cjkent: The factory methods can be package scoped like the class.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTaskRunner.java
```diff
@@ -0,0 +1,166 @@
...
+import com.opengamma.strata.calc.marketdata.CalculationEnvironment;
+
+/**
+ * Component that provides the ability to run calculation tasks.
+ * <p>
+ * This interface is the lower-level counterpart to {@link CalculationRunner}.
+ * It provides the ability to calculate results based on {@link CalculationTasks}.
+ * Unless you need to optimize, the {@code CalculationRunner} is a simpler entry point.
+ * <p>
+ * Each {@code CalculationTasks} instance contains one {@code CalculationTask} for each
```
cjkent: It might be clearer to describe a calculation task as corresponding to a single cell in a grid where the rows are defined by targets and the columns by measures.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTaskRunner.java
```diff
@@ -0,0 +1,166 @@
...
+   * to occur once, which could be a performance optimization if many different calculations
+   * are performed with the same set of trades and measures.
+   *
+   * @param targets  the targets for which values of the measures will be calculated
+   * @param columns  the configuration for the columns that will be calculated, including the measure and
+   *   any column-specific overrides
+   * @param calculationRules  the rules defining how the calculation is performed
+   * @return the tasks that perform the calculations
+   */
+  public abstract CalculationTasks createTasks(
```
cjkent: I don't think this method adds anything over just calling `CalculationTasks.of(targets, columns, calculationRules)`.

We put an equivalent method on the old `CalculationRunner` just in case we ever needed some state from the runner but we never did in all the time it was there. So I think it's safer (and simpler) to get rid of it.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTaskRunner.java
```diff
@@ -0,0 +1,166 @@
...
+      List<? extends CalculationTarget> targets,
+      List<Column> columns,
+      CalculationRules calculationRules);
+
+  //-------------------------------------------------------------------------
+  /**
+   * Performs calculations for a single set of market data.
+   * <p>
+   * This returns a grid of results based on the specified tasks and market data.
```
cjkent: More detail about the results would be good here given it's such a key API. Rows defined by targets, columns by measures etc.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/DefaultCalculationTaskRunner.java
```diff
@@ -103,33 +150,48 @@ public Results calculateSingleScenario(CalculationEnvironment marketData) {
...
+  public Results calculateMultipleScenarios(CalculationTasks tasks, CalculationEnvironment marketData) {
     Listener listener = new Listener(tasks.getColumns());
-    calculateMultipleScenariosAsync(marketData, listener);
+    calculateMultipleScenariosAsync(tasks, marketData, listener);
     return listener.result();
   }
 
   @Override
-  public void calculateSingleScenarioAsync(CalculationEnvironment marketData, CalculationListener listener) {
+  public void
```
cjkent: Newline before the method name

modules/function/src/test/java/com/opengamma/strata/function/calculation/swap/SwapPricingTest.java
```diff
@@ -146,11 +148,16 @@ public void presentValueVanillaFixedVsLibor1mSwap() {
     List<Column> columns = ImmutableList.of(Column.of(Measure.PRESENT_VALUE));
     ReportingRules reportingRules = ReportingRules.fixedCurrency(USD);
     CalculationRules rules = CalculationRules.of(pricingRules, marketDataRules, reportingRules);
-    CalculationRunner runner = CalculationRunnerFactory.ofSingleThreaded()
-        .createWithMarketDataBuilder(trades, columns, rules, marketDataFactory(), MarketDataConfig.empty());
 
     // calculate results using the runner
-    Results results = runner.calculateSingleScenario(suppliedData);
+    Results results = null;
+    try (CalculationTaskRunner runner = CalculationTaskRunner.of(MoreExecutors.newDirectExecutorService())) {
```
cjkent: No need to wrap it in a try-with-resources block if we're using a direct executor.

modules/function/src/test/java/com/opengamma/strata/function/marketdata/curve/CurveEndToEndTest.java
```diff
@@ -177,19 +179,22 @@ public void roundTripFraAndFixedFloatSwap() {
...
 
+    List<Column> columns = ImmutableList.of(Column.of(Measure.PRESENT_VALUE));
     MarketEnvironment knownMarketData = MarketEnvironment.builder()
         .valuationDate(date(2011, 3, 8))
         .addValues(parRateData)
         .build();
-    Results results = runner.calculateSingleScenario(knownMarketData);
+
+    Results results = null;
+    try (CalculationTaskRunner runner = CalculationTaskRunner.of(MoreExecutors.newDirectExecutorService())) {
```
cjkent: No need for try-with-resources

examples/src/test/java/com/opengamma/strata/examples/finance/credit/harness/TestCalculator.java
```diff
@@ -80,8 +80,9 @@ public Results calculateResults(LocalDate valuationDate, TradeSource tradeSource
 
     // create the engine and calculate the results
     ImmutableList<Trade> trades = ImmutableList.of(tradeSource.apply());
-    CalculationRunner runner = CalculationRunnerFactory.ofSingleThreaded().create(trades, columns, rules);
-    return runner.calculateSingleScenario(marketSnapshot);
+    try (CalculationRunner runner = CalculationRunner.of(MoreExecutors.newDirectExecutorService())) {
```
cjkent: No need for try-with-resources

examples/src/test/java/com/opengamma/strata/examples/regression/SwapReportRegressionTest.java
```diff
@@ -81,9 +83,13 @@ public void testResults() {
     LocalDate valuationDate = LocalDate.of(2009, 7, 31);
     MarketEnvironment marketSnapshot = marketDataBuilder.buildSnapshot(valuationDate);
 
-    CalculationRunner runner = CalculationRunnerFactory.ofSingleThreaded()
-        .createWithMarketDataBuilder(trades, columns, rules, marketDataFactory(), MarketDataConfig.empty());
-    Results results = runner.calculateSingleScenario(marketSnapshot);
+    Results results = null;
+    try (CalculationTaskRunner runner = CalculationTaskRunner.of(MoreExecutors.newDirectExecutorService())) {
```
cjkent: No need for try-with-resources

cjkent:
That's everything

PR #678, opened 2016-01-04
Added forward rate method in IborIndexRates. Changed IborFixingDeposit.

marc-opengamma:
The IborFixingDeposit pricer has been modified recently, causing problem in curve calibration when the fixing of the day is present in the time series. The test on that feature unfortunately contained an error, passing the wrong date for the fixing.
This PR create a new method in the IborIndexRates to access the forward rates directly.
It also reinstalled the fixing feature in the pricer, correct the test in the pricer, extend the tests to other methods and add a calibration test passing a fixing time series with the fixing of the day.
A new curve calibration test is created to check the curve calibration when the calibration date is an holiday (25-Dec in the test).

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/value/IborIndexRates.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/value/IborIndexRates.java
@@ -14,7 +14,6 @@ import com.opengamma.strata.market.Perturbation;
 import com.opengamma.strata.market.curve.Curve;
 import com.opengamma.strata.market.curve.CurveCurrencyParameterSensitivities;
 import com.opengamma.strata.market.curve.CurveName;
-import com.opengamma.strata.market.curve.CurveUnitParameterSensitivities;
 import com.opengamma.strata.market.key.IborIndexRatesKey;
 import com.opengamma.strata.market.sensitivity.IborRateSensitivity;
 import com.opengamma.strata.market.sensitivity.PointSensitivityBuilder;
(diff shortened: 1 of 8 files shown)


modules/market/src/main/java/com/opengamma/strata/market/value/IborIndexRates.java
```diff
@@ -90,6 +89,20 @@ public default IborIndexRatesKey getKey() {
...
+   * This method bypasses the potential historical time series associated to the index. 
+   * The standard method to access the Ibor rate is {@link IborIndexRates#rate}; this method should be used only 
+   * for specific purposes, when the actual time series are not relevant. 
+   * <p>
+   * In case there is a doubt on using this method, it probably means that it should not be used.
+   * 
+   * @param fixingDate  the fixing date to query the rate for
+   * @return the rate of the index as given by the forward curve
+   */
+  public abstract double forwardRate(LocalDate fixingDate);
```
jodastephen: This name might be too appealing for callers. Perhaps `rateForwardOnly()` ? Or `rateIgnoringTimeSeries` ?

modules/market/src/main/java/com/opengamma/strata/market/value/IborIndexRates.java
```diff
@@ -90,6 +89,20 @@ public default IborIndexRatesKey getKey() {
...
+   * Gets the forward rate at the specified fixing date.
+   * <p>
+   * This method bypasses the potential historical time series associated to the index. 
+   * The standard method to access the Ibor rate is {@link IborIndexRates#rate}; this method should be used only 
+   * for specific purposes, when the actual time series are not relevant. 
+   * <p>
+   * In case there is a doubt on using this method, it probably means that it should not be used.
+   * 
+   * @param fixingDate  the fixing date to query the rate for
+   * @return the rate of the index as given by the forward curve
```
jodastephen: This reverses the warning to put it first:

```
   * This method should only be used in rare and special cases.
   * In most cases callers should use {@link IborIndexRates#rate(LocalDate) rate(LocalDate)}.
   * <p>
   * An instance of {@code IborIndexRates} is typically based on a forward curve and a historic time-series.
   * The {@code rate(LocalDate)} method uses either the curve or time-series, depending on whether the
   * fixing date is before or after the valuation date. This method only queries the forward curve,
   * totally ignoring the time-series, which is needed for rare and special cases only.
```

marc-opengamma:
Javadoc and naming improvements following previous review.

PR #679, opened 2016-01-05
Create FxRatesArray

### LINKED ISSUE #582: Create ScenarioMarketDataValue implementation for FX rates
Storing an `FxRate` instance for every scenario is very inefficient, it would be trivial to create a `ScenarioMarketDataValue` implementation containing the `CurrencyPair` and the rates as a primitive double array.

cjkent:
This PR adds the `FxRatesArray` class which is an efficient container for multiple FX rates for the same currency pair for use in multiple scenarios.

Currently an `FxRate` instance containing a currency pair and rate must be created for each scenario. `FxRatesArray` has a single reference to the currency pair and stores the rates without boxing.

Fixes #582

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
@@ -120,9 +120,6 @@ public final class FxRatesArray implements ScenarioMarketDataValue<FxRate>, Immu
    * @throws IllegalArgumentException if no FX rate could be found
    */
   public double fxRate(Currency baseCurrency, Currency counterCurrency, int scenarioIndex) {
-    if (baseCurrency.equals(counterCurrency)) {
-      return 1d;
-    }
     if (baseCurrency.equals(pair.getBase()) && counterCurrency.equals(pair.getCounter())) {
       return rates.get(scenarioIndex);
     }
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
@@ -1,0 +1,473 @@
+/**
+ * Copyright (C) 2016 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.basics.currency;
+
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+import org.joda.beans.Bean;
+import org.joda.beans.BeanDefinition;
+import org.joda.beans.ImmutableBean;
+import org.joda.beans.ImmutableValidator;
+import org.joda.beans.JodaBeanUtils;
+import org.joda.beans.MetaProperty;
+import org.joda.beans.Property;
+import org.joda.beans.PropertyDefinition;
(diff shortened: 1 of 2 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
```diff
@@ -0,0 +1,473 @@
...
+ * elements - the base currency 'EUR', the counter currency 'USD' and the rate '1.25'.
+ * When performing a conversion a rate of '1.25' means that '1 EUR = 1.25 USD'.
+ * <p>
+ * The {@link FxRate} class represents a single rate for a currency pair. This class is
+ * intended as an efficient way of storing multiple rates for the same currency pair
+ * for use in multiple scenarios.
+ *
+ * @see FxRate
+ */
+@BeanDefinition
```
jodastephen: builderScope of private

modules/basics/src/test/java/com/opengamma/strata/basics/currency/FxRatesArrayTest.java
```diff
@@ -0,0 +1,45 @@
...
+    assertThat(rates.fxRate(Currency.USD, Currency.EUR, 0)).isEqualTo(1 / 1.07);
+    assertThat(rates.fxRate(Currency.USD, Currency.EUR, 1)).isEqualTo(1 / 1.08);
+    assertThat(rates.fxRate(Currency.USD, Currency.EUR, 2)).isEqualTo(1 / 1.09);
+  }
+
+  public void identicalCurrenciesHaveRateOfOne() {
+    assertThrowsIllegalArg(
+        () -> FxRatesArray.of(Currency.EUR, Currency.EUR, DoubleArray.of(1.07, 1.08, 1.09)),
+        "Conversion rate between identical currencies must be one");
+  }
```
jodastephen: When testing beans, it is helpful to have a method called "coverage" that calls `coverImmutableBean()` like this one:
https://github.com/OpenGamma/Strata/blob/master/modules/product/src/test/java/com/opengamma/strata/product/fra/FraTest.java#L232

This ensures that we can rapidly find those parts of the system that are low on test coverage. Note that this one also calls `coverBeanEquals()` with two beans where every property differs to complete the coverage.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRatesArray.java
```diff
@@ -0,0 +1,473 @@
...
+      return 1d;
+    }
+    if (baseCurrency.equals(pair.getBase()) && counterCurrency.equals(pair.getCounter())) {
+      return rates.get(scenarioIndex);
+    }
+    if (counterCurrency.equals(pair.getBase()) && baseCurrency.equals(pair.getCounter())) {
+      return 1d / rates.get(scenarioIndex);
+    }
+    throw new IllegalArgumentException("Unknown rate: " + baseCurrency + "/" + counterCurrency);
+  }
```
jodastephen: I imagine there will be bulk conversion methods eventually, but best to add them as needed.

cjkent: I started adding them and gave up when I realised they wouldn't be as simple and clear-cut as I first thought. The `fxRate` method gives an efficient, unboxed way to get at the rate so I figured that would do for now.

PR #684, opened 2016-01-06
Remove market data views from market data builder

jodastephen:
Remove key and ID classes for views.
Rename `MarketDataValue` to `MarketDataView`.
Split `RateIndex` key/ID to `IborIndex` and `OvernightIndex`.
Enhance example data to build missing IDs.
Simplify examples to use standard calculation runner.


modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
```diff
@@ -99,15 +102,18 @@ public double fxRate(Currency baseCurrency, Currency counterCurrency) {
   //-------------------------------------------------------------------------
   @Override
   public DiscountFactors discountFactors(Currency currency) {
-    return marketData.getValue(DiscountFactorsKey.of(currency));
+    Curve curve = marketData.getValue(DiscountCurveKey.of(currency));
```
cjkent: `MarketData.getValue` never returns null. It throws `IllegalArgumentException` for an unknown key.

modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
```diff
@@ -122,19 +128,34 @@ public FxForwardRates fxForwardRates(CurrencyPair currencyPair) {
   //-------------------------------------------------------------------------
   @Override
   public IborIndexRates iborIndexRates(IborIndex index) {
-    return marketData.getValue(IborIndexRatesKey.of(index));
+    Curve curve = marketData.getValue(IborIndexCurveKey.of(index));
+    if (curve == null) {
```
cjkent: MarketData.getValue never returns null

modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
```diff
@@ -122,19 +128,34 @@ public FxForwardRates fxForwardRates(CurrencyPair currencyPair) {
...
+    }
+    return IborIndexRates.of(index, getValuationDate(), curve, timeSeries(index));
   }
 
   //-------------------------------------------------------------------------
   @Override
   public OvernightIndexRates overnightIndexRates(OvernightIndex index) {
-    return marketData.getValue(OvernightIndexRatesKey.of(index));
+    Curve curve = marketData.getValue(OvernightIndexCurveKey.of(index));
+    if (curve == null) {
```
cjkent: MarketData.getValue never returns null

modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
```diff
@@ -122,19 +128,34 @@ public FxForwardRates fxForwardRates(CurrencyPair currencyPair) {
...
+    }
+    return OvernightIndexRates.of(index, getValuationDate(), curve, timeSeries(index));
   }
 
   //-------------------------------------------------------------------------
   @Override
   public PriceIndexValues priceIndexValues(PriceIndex index) {
-    return marketData.getValue(PriceIndexValuesKey.of(index));
+    Curve curve = marketData.getValue(PriceIndexCurveKey.of(index));
+    if (curve == null) {
```
cjkent: MarketData.getValue never returns null

modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
```diff
@@ -122,19 +128,34 @@ public FxForwardRates fxForwardRates(CurrencyPair currencyPair) {
...
+      throw new IllegalArgumentException("Unable to find forward curve: " + index);
+    }
+    if (!(curve instanceof InterpolatedNodalCurve)) {
```
cjkent: This smells bad. Doesn't it suggest that `PriceIndexCurveKey` should explicitly refer to `InterpolatedNodalCurve` instead of `Curve`?

modules/market/src/main/java/com/opengamma/strata/market/id/IndexRateId.java
```diff
@@ -33,7 +33,7 @@
 /**
  * A market data ID identifying the current and historical values for an {@link Index}.
  * <p>
- * The forward curve of the index is identified with an {@link RateIndexCurveId}.
+ * The forward curve of the index is identified with a separate ID, such as n {@link IborIndexCurveId}.
```
cjkent: Typo - "n" should be "an"

modules/function/src/main/java/com/opengamma/strata/function/marketdata/MarketDataRatesProvider.java
jodastephen: Doing so has knock on effects as the key implements `CurveKey`. There may be other implementations of the interface that do accept `Curve`, so lets leave this for now.

cjkent: Does it need to implement `CurveKey` or could that be changed? Or does `CurveKey` need to have a type parameter `<T extends Curve>`?

jodastephen: See #690

PR #711, opened 2016-01-13
Fix measure name evaluation in reportin tool

### LINKED ISSUE #686: Report tool no longer suggests valid measure names
`PricingRules` has a method `configuredMeasures` which allows an application to list the available measures. This was used when parsing a trade report template to provide a helpful list of valid measure names in the failure message if the user only enters `Measures.` (almost like auto-complete) or the measure name they have chosen is invalid. This allowed report templates to be written more "interactively" through repeated running of the command-line tool. Unfortunately this behaviour has been lost in the refactoring.

jodastephen:
Ensure the list of available measures is returned in case of error
Fixes #686

### DIFF
--- a/modules/report/src/main/java/com/opengamma/strata/report/framework/expression/RootEvaluator.java
+++ b/modules/report/src/main/java/com/opengamma/strata/report/framework/expression/RootEvaluator.java
@@ -8,6 +8,7 @@ package com.opengamma.strata.report.framework.expression;
 import java.util.List;
 import java.util.Set;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
 
 /**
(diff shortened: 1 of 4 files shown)


modules/report/src/main/java/com/opengamma/strata/report/framework/expression/RootEvaluator.java
```diff
@@ -56,4 +52,17 @@ public EvaluationResult evaluate(ResultsRow resultsRow, String firstToken, List<
     }
   }
 
+  // find the result starting from a measure
+  private EvaluationResult evaluateMeasures(ResultsRow resultsRow, List<String> remainingTokens) {
+    // if no measures, return list of valid measures
+    if (remainingTokens.isEmpty() || Strings.nullToEmpty(remainingTokens.get(0)).trim().isEmpty()) {
+      List<String> measureNames = ResultsRow.measureNames(resultsRow.getTrade());
+      return EvaluationResult.failure("No measure specified. Use one of: {}", measureNames);
```
jonathansenior: Is this tested?

PR #715, opened 2016-01-15
Make CalculationListener a functional interface

cjkent:


cjkent:
I've realised the use case I wanted this for actually needs to implement both methods in the interface.

PR #717, opened 2016-01-15
Add MapStream for easier streaming over maps

cjkent:


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -0,0 +1,391 @@
...
+  private final Stream<Map.Entry<K, V>> underlying;
+
+  /**
+   * Returns a stream over the entries in the map.
+   *
+   * @param map a map
+   * @return a stream over the entries in the map
+   */
+  public static <K, V> MapStream<K, V> of(Map<K, V> map) {
+    return new MapStream<>(map.entrySet().stream());
```
jodastephen: Missing `empty()` factory.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -0,0 +1,391 @@
...
+  /**
+   * Transforms the entries in the stream by applying a mapper function to each key and value.
+   *
+   * @param mapper  a mapper function whose return values are included in the new stream
+   * @param <R>  the type of elements in the new stream
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <R> Stream<R> map(BiFunction<? super K, ? super V, ? extends R> mapper) {
+    return underlying.map(e -> mapper.apply(e.getKey(), e.getValue()));
+  }
```
jodastephen: Add `mapToValues()` and `mapToKeys()`. These would be useful if you need to filter by key but then only need the values.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -0,0 +1,391 @@
...
+   * Returns an immutable map built from the entries in the stream.
+   * <p>
+   * The keys must be unique or an exception will be thrown. Duplicate keys can be handled by using
+   * {@link #collect(Collector)} and {@code Collectors.toMap}.
+   *
+   * @return an immutable map built from the entries in the stream
+   */
+  public ImmutableMap<K, V> toImmutableMap() {
+    return underlying.collect(Guavate.toImmutableMap(e -> e.getKey(), e -> e.getValue()));
+  }
```
jodastephen: Could we just have `toMap()`? Would it be clear enough?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -0,0 +1,391 @@
...
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * A stream implementation which adds methods for manipulating keys and values when streaming over map entries.
+ */
+public class MapStream<K, V> implements Stream<Map.Entry<K, V>> {
```
jodastephen: Class should be `final`?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -0,0 +1,391 @@
...
+
+  @Override
+  public void close() {
+    underlying.close();
+  }
+
+  //--------------------------------------------------------------------------------------------------
+
+  private static <K, V> Map.Entry<K, V> entry(K k, V v) {
+    return new AbstractMap.SimpleEntry<>(k, v);
```
jodastephen: `SimpleImmutableEntry` not `SimpleEntry`?

modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
```diff
@@ -0,0 +1,81 @@
...
+    assertThat(result).isEqualTo(expected);
+  }
+
+  public void toMapDuplicateKeys() {
+    assertThrowsIllegalArg(() -> MapStream.of(map).mapKeys(k -> "key").toMap());
+  }
+
+  public void toImmutableMapDuplicateKeys() {
+    assertThrowsIllegalArg(() -> MapStream.of(map).mapKeys(k -> "key").toImmutableMap());
+  }
```
jodastephen: Test coverage is probably quite low, but other methods are not that interesting. Worth adding a `coverage` test just to deal with them.

jodastephen:
Looks good overall. Will be interested to see how it works in practice.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
cjkent: I'm not sure I know what you mean. What signature are you thinking of?

cjkent:
Added `MapStream.forEach` after reviewing usages of `Map.entrySet().stream()` in Strata.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
jodastephen: `Stream<V> mapToValues()`

cjkent: What parameters?

jodastephen: None. It is the equivalent of calling `values()` on a `Map`.

cjkent: OK, in which case I don't think the method name should include "map". That's what confused me. Map implies applying a function to the elements. I think `keys()` and `values()` would be better.

But you could achieve the same thing with the existing API like this:

    stream.map((k, v) -> v)...

I'm not sure those 2 methods add enough to justify inclusion at the moment, I've only added the ones I think we really need.

PR #718, opened 2016-01-15
Add ObjArray and Array interface

jodastephen:
Provide boxed view for DoubleArray


modules/collect/src/main/java/com/opengamma/strata/collect/array/Array.java
```diff
@@ -0,0 +1,59 @@
...
+   * @return true if empty
+   */
+  public default boolean isEmpty() {
+    return size() == 0;
+  }
+
+  /**
+   * Gets the element at the specified index in this array.
+   * <p>
+   * The index must be valid, between zero and {@code size()}.
```
cjkent: I read this as meaning `0 <= index <= size` which is obviously wrong. "between zero and `size() - 1`"?

modules/collect/src/main/java/com/opengamma/strata/collect/array/BoxedDoubleArray.java
```diff
@@ -0,0 +1,224 @@
...
+import com.opengamma.strata.collect.ArgChecker;
+import com.opengamma.strata.collect.DoubleArrayMath;
+
+/**
+ * An immutable array of {@code double} values, boxed to {@code Double}.
+ * <p>
+ * This is used to wrap a {@code double[]} to meet the API of {@link Array}.
+ * An instance is created using {@link DoubleArray#boxed()}.
+ */
+final class BoxedDoubleArray
```
cjkent: What's the intended use case for this class?

Also, it's an `ImmutableBean` but I can't see `AUTOGENERATED` anywhere. Was it hand-written?

cjkent: Is the intention to replace `ScenarioResult` as the return type of calculation functions?

jodastephen: Yes, that is my thought. It seems consistent to conceptually either return a single value or an array of values.

modules/collect/src/main/java/com/opengamma/strata/collect/array/ObjArray.java
```diff
@@ -0,0 +1,929 @@
...
+/**
+ * An immutable array of {@code Object} values.
+ * <p>
+ * This provides a cut down version of {@link List} that implements {@link Array}.
+ * <p>
+ * It is intended that the array does not contain {@code null}, however this is not validated.
+ * 
+ * @param <T>  the type of the element
+ */
+public final class ObjArray<T>
```
cjkent: I'm not sure why we need this class. It seems like an `ImmutableBean` implementing `Array` with an `ImmutableMap` property would provide the same functionality for a fraction of the effort.

cjkent: N.B. I edit the previous comment but I know Github doesn't send notifications of edits.

modules/collect/src/main/java/com/opengamma/strata/collect/array/BoxedDoubleArray.java
cjkent: I'm not sure about that. Having "scenario" in the name communicates the fact that it contains values for multiple scenarios. There is logic applied to those values to extract the single value when the user calls the single-scenario variants of `CalculationRunner.calculate*`. If we just use a general array type then users might use them for general arrays of values which might produce surprising behaviour.

modules/collect/src/main/java/com/opengamma/strata/collect/array/ObjArray.java
cjkent: Given my comment about keeping `ScenarioResult` as the return type of calculation functions, I don't think we need this class at all. `DefaultScenarioResult` does as much as we need. Although maybe it could do with a better name.

modules/collect/src/main/java/com/opengamma/strata/collect/array/BoxedDoubleArray.java
jodastephen: `BoxedDoubleArray` is a hand written bean, just like `DoubleArray`. Occasionally it is the right way to go, here it made sense as it is heavily based on `DoubleArray`.

cjkent: What's the advantage of hand-writing it?

cjkent: To expand on my point about returning `ScenarioResult` from functions - If a function returns an array, how can you tell whether it contains one value for each scenario, or whether it's an aggregate result calculated for all scenarios?

In the current design, a `ScenarioResult` indicates a collection of results calculated independently for each scenario and any other result type is an aggregate result.

jodastephen:
I've updated the PR with simpler package-scoped `ObjectArray`, `FilledObjectArray` and more methods on `Array`. The multi-measure function code makes good use of `Array` as a flexible internal structure with a simple API.

cjkent:
What's the advantage of using `Array` over `List` for reference types?

jodastephen:
Leaving this for now.

PR #725, opened 2016-01-18
Introduce multi-measure functions

jodastephen:
Setup interface and integrate.
Convert FRA and TermDeposit.

This PR gets the ball rolling on multi-measure functions. It sets out the new multi-measure function interface and two implementations. Integration into the calculation framework is simple and not that pretty, but effective enough until the rest of the functions are converted. The calculation of measures is separated from the functions for clarity, with measures currently package scoped. It may be necessary to to incorporate major updates to this PR in subsequent PRs to avoid too large a change.


modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/ScenarioResult.java
```diff
@@ -34,6 +34,8 @@
 
   /**
    * Returns the result at the specified index.
+   * <p>
+   * The index must be valid, between zero and the number of scenarios.
```
cjkent: This wording is ambiguous. It's only correct if "between" can be interpreted as including the lower bound but not the upper. Which seems a bit of a stretch.

modules/function/src/main/java/com/opengamma/strata/function/calculation/deposit/TermDepositCalculationFunction.java
```diff
@@ -0,0 +1,129 @@
...
+  public Map<Measure, Result<ScenarioResult<?>>> calculate(
+      TermDepositTrade trade,
+      Set<Measure> measures,
+      CalculationMarketData scenarioMarketData) {
+
+    // expand the trade once for all measures and all scenarios
+    ExpandedTermDeposit product = trade.getProduct().expand();
+
+    // loop around measures, calculating all scenarios for one measure
+    Map<Measure, Result<ScenarioResult<?>>> results = new HashMap<>();
```
cjkent: return measures.stream().collect(toMap(m -> m, m -> calculate(m, trade, product, scenarioMarketData));

modules/function/src/main/java/com/opengamma/strata/function/calculation/deposit/TermDepositCalculationFunction.java
```diff
@@ -0,0 +1,129 @@
...
+    SingleMeasureCalculation calculator = CALCULATORS.get(measure);
+    if (calculator == null) {
+      return Result.failure(FailureReason.INVALID_INPUT, "Unsupported measure: {}", measure);
+    }
+    return Result.of(() -> calculator.calculate(trade, product, scenarioMarketData));
+  }
+
+  //-------------------------------------------------------------------------
+  @FunctionalInterface
+  interface SingleMeasureCalculation {
```
cjkent: Can we abstract this out rather than having to define roughly the same interface in every function?

modules/function/src/main/java/com/opengamma/strata/function/calculation/fra/FraCalculationFunction.java
```diff
@@ -0,0 +1,161 @@
...
+  public Map<Measure, Result<ScenarioResult<?>>> calculate(
+      FraTrade trade,
+      Set<Measure> measures,
+      CalculationMarketData scenarioMarketData) {
+
+    // expand the trade once for all measures and all scenarios
+    ExpandedFra product = trade.getProduct().expand();
+
+    // loop around measures, calculating all scenarios for one measure
+    Map<Measure, Result<ScenarioResult<?>>> results = new HashMap<>();
```
cjkent: return measures.stream().collect(toMap(m -> m, m -> calculate(m, trade, product, scenarioMarketData));

modules/function/src/main/java/com/opengamma/strata/function/calculation/fra/FraCalculationFunction.java
```diff
@@ -0,0 +1,161 @@
...
+    SingleMeasureCalculation calculator = CALCULATORS.get(measure);
+    if (calculator == null) {
+      return Result.failure(FailureReason.INVALID_INPUT, "Unsupported measure: {}", measure);
+    }
+    return Result.of(() -> calculator.calculate(trade, product, scenarioMarketData));
+  }
+
+  //-------------------------------------------------------------------------
+  @FunctionalInterface
+  interface SingleMeasureCalculation {
```
cjkent: Abstract out?

cjkent:
That's everything

modules/function/src/main/java/com/opengamma/strata/function/calculation/deposit/TermDepositCalculationFunction.java
jodastephen: Deliberately chose to make the function unabstracted and with a minimum of streams. A key problem in the past has been getting people to debug in, step through and see where things are going wrong. By making the calculate method of these function classes as simple as possible, using just classic Java, it will help them debug. Abstractions in particular seem to be part of the problem in functions.

FWIW, not all of the interfaces take three arguments. Some trade types need 2 and CDS needs more.

I did consider making the `*MeasureCalculations` classes also avoid streams, but as it stands the collector is too useful there. I have fixed the code to avoid boxing in parRate/parSpread.

PR #730, opened 2016-01-20
Add LocalDateDoubleTimeSeries.mapDates

cjkent:
Time series already have a `mapValues` method but there was no corresponding `mapDates`.


cjkent:
retest this please

cjkent:
Don't review or merge this yet, I've thought of something I need to check.

cjkent:
The problem is the potential to reorder the dates. There are 3 ways we could deal with this

* Reorder the values - is this too subtle?
* Forbid reordering, throw an exception if the mapped dates aren't in ascending order
* Abandon the whole idea; this would be a shame as I have a concrete use case where it makes the calling code much nicer

jodastephen:
It would have to be option 2, exception if reordered.

modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
```diff
@@ -57,7 +62,7 @@
    * array for weekends and providing the different date
    * calculations for each case.
    */
-  static enum DenseTimeSeriesCalculation {
+  enum DenseTimeSeriesCalculation {
```
jodastephen: It was clearer with the static (the code reader should not need to know offhand that nested enums are static by default)

cjkent: They're not static by default, they're just static. Always.

jodastephen: Agreed, but it is better for the source to be explicit, then it doesn't rely on esoteric knowledge of the code reader.

PR #731, opened 2016-01-21
Remove calculation function hierarchy

jodastephen:
Rename `CalculationMultiFunction` to `CalculationFunction`
Remove alternatives so it is the only function interface
Fix code and tests to match


modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTask.java
```diff
@@ -230,30 +223,28 @@ public MarketDataRequirements requirements() {
...
-      result = Result.failure(ex);
+  /**
+   * Calculates the result for the specified market data.
+   * 
+   * @param calculationData  the market data
+   * @return the result
+   */
+  private Result<?> calculate(CalculationMarketData calculationData) {
+    ImmutableSet<Measure> measures = ImmutableSet.of(getMeasure());
+    Map<Measure, Result<ScenarioResult<?>>> map = function.calculate(target, measures, calculationData);
```
cjkent: Is the intention to change `CalculationTask` and `CalculationResult` to handle multiple values in a later PR?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CalculationFunction.java
```diff
@@ -39,4 +61,32 @@
     return Optional.empty();
   }
 
+  /**
+   * Determines the market data requirements that function has to perform its calculations.
```
cjkent: Sounds a bit awkward, maybe "Determines the market data required by the function to perform its calculations".

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CalculationFunction.java
```diff
@@ -39,4 +61,32 @@
...
+   * The set of measures must only contain measures that the function supports,
+   * as returned by {@link #supportedMeasures()}. The market data must provide at least the
+   * set of data requested by {@link #requirements(CalculationTarget, Set)}.
+   *
+   * @param target  the target of the calculation
+   * @param measures  the set of measures to calculate
+   * @param marketData  the market data to be used in the calculation
+   * @return the read-only map of calculated values, keyed by their measure
+   */
+  public abstract Map<Measure, Result<ScenarioResult<?>>> calculate(
```
cjkent: I don't think the return type is right. At the moment there is no requirement for a function to return a `ScenarioResult`. It's possible to write a function that returns a single result which is an aggregate value for all scenarios.

As a trivial example, it's possible to write a function that returns the maximum PV for a set of scenarios. That would be a single value applying to all scenarios. Constraining the result type to `ScenarioResult` implies the results will always be associated with a particular scenario.

The current return type of `CalculationSingleFunction.execute` is any type, not a subtype of `ScenarioResult`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/ScenarioResult.java
```diff
@@ -7,14 +7,13 @@
...
-import com.opengamma.strata.calc.runner.function.CalculationSingleFunction;
+import com.opengamma.strata.calc.runner.function.CalculationFunction;
 
 /**
  * A container for multiple results produced by performing a single calculation across multiple scenarios.
  * <p>
- * This class is intended to be used as the return value from the {@code execute} method of
- * implementations of {@link CalculationSingleFunction} and {@link CalculationMultiFunction}.
+ * A {@link CalculationFunction} produces a map of measure to result.
+ * Each result is an instance of this type, providing access to the scenarios that were calculated.
```
cjkent: This probably shouldn't be true.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTask.java
jodastephen: Yes, this PR was already large

PR #735, opened 2016-01-21
Enhance floating rate names

jodastephen:
Extend concept to have standard Strata names as well as FpML ones
Add constants

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/index/FloatingRateName.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/index/FloatingRateName.java
@@ -6,6 +6,7 @@
 package com.opengamma.strata.basics.index;
 
 import java.io.Serializable;
+import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/index/FloatingRateName.java
```diff
@@ -125,24 +126,25 @@ public static FloatingRateName of(String uniqueName) {
...
+  // parse a single section
+  private static void parseSection(
+      PropertySet section,
+      String indexNameSuffix,
+      FloatingRateType type,
+      HashMap<String, FloatingRateName> mutableMap) {
+
+    // find our names from the RHS of the key/value pairs
+    for (String key : section.keys()) {
+      mutableMap.put(key, new FloatingRateName(key, section.value(key) + indexNameSuffix, type));
```
jonathansenior: By replacing the map builder, we're losing the automatic check for duplicate keys - is this likely to be an issue? The last duplicate key seen will now win.

PR #745, opened 2016-01-26
Scenario result factories

jodastephen:
Add functional factories to `ScenarioResult` implementations, providing opportunity to optimize currency and multi-currency cases.
Using this approach reduces the number of framework stack frames.
Add scenario access methods to `CalculationMarketData`.
Add missing static factory methods to market data containers.
Add generics to `FxConvertibleList`.
Reduce importance of `RatesProvider` in functions, reducing impact of potential future changes.

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/CalculationMarketData.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/CalculationMarketData.java
@@ -6,7 +6,9 @@
 package com.opengamma.strata.calc.marketdata;
 
 import java.time.LocalDate;
+import java.util.stream.Stream;
 
+import com.opengamma.strata.basics.market.MarketData;
 import com.opengamma.strata.basics.market.MarketDataBox;
 import com.opengamma.strata.basics.market.MarketDataKey;
 import com.opengamma.strata.basics.market.ObservableKey;
@@ -21,6 +23,12 @@ import com.opengamma.strata.collect.timeseries.LocalDateDoubleTimeSeries;
  * The set of data provided by this interface is a subset of the set provided by {@link CalculationEnvironment}.
  * For example a function might request a USD discounting curve, but the scenario market data can contain
  * multiple curve groups, each with a USD discounting curve.
+ * <p>
+ * There are two ways to access the available market data.
+ * The first way is to use the method {@link #getValue(MarketDataKey)} which returns
+ * the data associated with a single key for all scenarios.
+ * The second way is to use the method {@link #scenarios()} or {@link #scenario(int)}
+ * which return all the data associated with a single scenario.
  * <p>
  * Typically a set of {@link MarketDataRules} are used to choose the item of market data from the global set.
  * <p>
@@ -41,6 +49,29 @@ public interface CalculationMarketData {
    * @return the number of scenarios
    */
   public abstract int getScenarioCount();
+
+  //-------------------------------------------------------------------------
+  /**
+   * Returns a stream of market data, one for each scenario.
+   * <p>
+   * The stream will return instances of {@link MarketData}, where each represents
+   * a single scenario view of the complete set of data.
+   *
+   * @return the stream of market data, one for the each scenario
+   * @throws IndexOutOfBoundsException if the index is invalid
+   */
+  public abstract Stream<MarketData> scenarios();
+
+  /**
+   * Returns market data for a single scenario.
+   * <p>
+   * This returns a view of the market data for the single specified scenario.
+   *
+   * @param scenarioIndex  the scenario index
+   * @return the market data for the specified scenario
+   * @throws IndexOutOfBoundsException if the index is invalid
+   */
+  public abstract MarketData scenario(int scenarioIndex);
 
   //-------------------------------------------------------------------------
   /**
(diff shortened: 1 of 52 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/CalculationMarketData.java
```diff
@@ -22,6 +24,12 @@
  * For example a function might request a USD discounting curve, but the scenario market data can contain
  * multiple curve groups, each with a USD discounting curve.
  * <p>
+ * There are two ways to access the available market data.
+ * The first way is to use the method {@link #getValue(MarketDataKey)} which returns
+ * the data associated with a single key for all scenarios.
+ * The second way is to use the method {@link #scenarios()} or {@link #scenario(int)}
+ * which return all the data associated with a single scenario.
+ * <p>
```
cjkent: Probably worth noting that it's more efficient to use `getValue` and most efficient to use `getScenarioValue`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
```diff
@@ -94,6 +96,32 @@ public static CurrencyValuesArray of(List<CurrencyAmount> amounts) {
...
+   * Obtains an instance using a function to create the entries.
+   * <p>
+   * The function is passed the scenario index and returns the value for that index.
+   * 
+   * @param size  the number of elements
+   * @param valueFunction  the function used to obtain each value
+   * @return an instance initialized using the function
+   * @throws IllegalArgumentException is size is zero or less
+   */
+  public static CurrencyValuesArray of(int size, IntFunction<CurrencyAmount> valueFunction) {
```
cjkent: There's quite a lot of boxing and churn going on here. It might be better to take `IntToDoubleFunction`.

If there is a strong use case for this signature with existing functions producing `CurrencyAmount` I still think an override taking `IntToDoubleFunction` is worthwhile.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
```diff
@@ -94,6 +96,32 @@ public static CurrencyValuesArray of(List<CurrencyAmount> amounts) {
...
+    ArgChecker.notNegativeOrZero(size, "size");
```
cjkent: I'd expect to be allowed to create an instance with size zero. You probably wouldn't want to, but you wouldn't want to have to check for it as a special case.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/FxConvertibleList.java
```diff
@@ -63,8 +67,8 @@
    * @return an instance with the specified values
    */
   @SafeVarargs
-  public static FxConvertibleList of(FxConvertible<?>... values) {
-    return new FxConvertibleList(ImmutableList.copyOf(values));
+  public static <R extends FxConvertible<?>> FxConvertibleList<R> of(R... values) {
```
cjkent: In other places we've tended to use the same type parameter name in static methods as in the class, i.e. `T` in this case.

modules/function/src/main/java/com/opengamma/strata/function/calculation/deposit/TermDepositMeasureCalculations.java
```diff
@@ -114,30 +122,19 @@ private static MultiCurrencyAmount calculatePv01(
       ExpandedTermDeposit product,
       CalculationMarketData marketData) {
 
-    return ratesProviderStream(marketData)
-        .map(provider -> calculateBucketedPv01(product, provider))
-        .collect(toScenarioResult());
+    return FxConvertibleList.of(
```
cjkent: I'm not sure about explicitly calling `FxConvertibleList.of`, it seems like a step backwards in usability.

Previously you didn't have to consider whether the currencies in your result could be automatically converted. When you called `toScenarioResult` the system checked each of the results and wrapped them in an `FxConvertibleList` if they all implemented `FxConvertible`, otherwise is used a `DefaultScenarioResult`.

Maybe there should be a method `ScenarioResult.of` which does the same thing.

I did consider taking it a step further and automatically using `CurrencyValuesArray` when all the values were `CurrencyAmount` and `MultiCurrencyValuesArray` for `MultiCurrencyAmount`. To be honest I can't remember why I decided against that, but it's possible things have changed since then and we could consider it now.

cjkent: I've remembered the problem with automatically creating `CurrencyValuesArray` and `MultiCurrencyValuesArray` - the return type of the function would have to be `ScenarioResult<CurrencyAmount>` instead of `CurrencyValuesArray`. Whether or not that's a problem is debatable.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
jodastephen: Since the currency is extracted from the function, the size must be at least size one.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/result/CurrencyValuesArray.java
jodastephen: Such an override is not currently needed, and there is a simple alternative, using `DoubleArray.of(int, IntToDoubleFunction)`. Added javadoc to explain.

PR #747, opened 2016-01-27
Add swap payment schedule adjustments

jodastephen:
Process the additional adjustment.
Clarify that dates are from the accrual schedule where unclear.

Reading closely, it seems that FpML will not use this, however it could produce different dates in certain edge cases, so it is a useful addition.

### DIFF
--- a/modules/basics/src/test/java/com/opengamma/strata/basics/schedule/SchedulePeriodTest.java
+++ b/modules/basics/src/test/java/com/opengamma/strata/basics/schedule/SchedulePeriodTest.java
@@ -5,6 +5,8 @@
  */
 package com.opengamma.strata.basics.schedule;
 
+import static com.opengamma.strata.basics.date.BusinessDayConventions.FOLLOWING;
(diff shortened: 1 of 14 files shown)


modules/basics/src/test/java/com/opengamma/strata/basics/schedule/SchedulePeriodTest.java
```diff
@@ -41,7 +43,8 @@
   private static final LocalDate JUL_05 = date(2014, JULY, 5);
   private static final LocalDate JUL_17 = date(2014, JULY, 17);
   private static final LocalDate JUL_18 = date(2014, JULY, 18);
-  private static final LocalDate AUG_17 = date(2014, AUGUST, 17);
+  private static final LocalDate AUG_17 = date(2014, AUGUST, 17);  // Sunday
+  private static final LocalDate AUG_18 = date(2014, AUGUST, 18);  // Sunday
```
marc-opengamma: 18 Aug 2014 is Monday

PR #752, opened 2016-01-28
Integrate currency conversion

jodastephen:
Remove `FxConvertibleList`.
Extend use of `CurrencyConvertible` with `isConvertible()` method.

The advantage of this approach is that a class is removed (and there is no need for `FxConvertibleSingleResult` which doesn't exist yet). The disadvantage is the extra `isConvertible()` method, suggesting maybe `CurrencyConvertible` should be renamed.


modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CurrencyConvertible.java
```diff
@@ -17,11 +17,30 @@
...
   /**
-   * Returns a copy of the object with any currency amounts converted into the reporting currency.
+   * Checks whether this object can be currency converted.
+   * <p>
+   * This will return true if some form of currency conversion will be attempted
+   * by {@link #convertedTo(Currency, CalculationMarketData)}.
+   * 
+   * @return true if currency conversion is possible
+   */
+  public abstract boolean isConvertible();
```
cjkent: Why is this method required? The design we discussed was for `convertedTo` to simply return `this` when called on something that isn't actually convertible.

It is also a blatant violation of tell-don't-ask.

cjkent:
That's everything.

jodastephen:
This has now been updated following recent PRs to avoid the need for a `boolean` method on `CurrencyConvertible`.

PR #754, opened 2016-01-29
Replace ReportingRules with ReportingCurrency

### LINKED ISSUE #692: Consider removing ReportingRules and using a reporting currency
`ReportingRules` was created on the assumption that one day it would contain more than just the reporting currency. That hasn't happened. For simplicity it might make sense to get rid of `ReportingRules` and directly specify the reporting currency instead.

We would lose one feature if we made this change. `ReportingRules` allows a different reporting currency to be specified for each row in the results. We have never used this ability. If we remove `ReportingRules` we would have the following options for reporting currency which are probably sufficient:

1. A single reporting currency is used for each column in the results grid. The reporting currency is specified in the `Column`.
2. There is no reporting currency specified in the `Column` and the default currency is used for the measure and trade. This default is specified by `CalculationFunction.defaultReportingCurrency`.

jodastephen:
Simplify concept, solving issue where empty rules had two meanings
Remove unused `ColumnDefinition` concept
Rename defaultReportingCurrency() to naturalCurrency() on function
Fixes #692


modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRules.java
```diff
@@ -36,45 +37,73 @@
...
+   * This defaults to an empty instance if using the builder.
+   */
   @PropertyDefinition(validate = "notNull")
   private final PricingRules pricingRules;
-
-  /** The rules defining what market data should be used in each calculation. */
+  /**
+   * The rules that define what market data should be used in each calculation.
+   * <p>
+   * This controls which view of the market data should be used for a calculation.
```
cjkent: I don't think "...view of the market data..." is obvious enough without more context and I don't think it adds anything more than the previous sentence.

It might be worth elaborating; explaining that there is a global set of market data and the rules choose the subset which is used for each calculation. Or you could just get rid of that sentence.

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRules.java
```diff
@@ -137,7 +169,10 @@ public PricingRules getPricingRules() {
 
   //-----------------------------------------------------------------------
   /**
-   * Gets the rules defining what market data should be used in each calculation.
+   * Gets the rules that define what market data should be used in each calculation.
+   * <p>
+   * This controls which view of the market data should be used for a calculation.
```
cjkent: Same as above, "view of market data"

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
```diff
@@ -34,66 +36,122 @@
...
+  /**
+   * The pricing rules that apply to this column, overriding the default rules.
+   * In most cases, there is no need to override the rules on a per-column basis.
+   */
   @PropertyDefinition(validate = "notNull")
   private final PricingRules pricingRules;
-
-  /** The market data rules that apply to this column in addition to the default rules. */
+  /**
+   * The market data rules that apply to this column, overriding the default rules.
```
cjkent: The previous description was more accurate. The rules are merged with the column rules taking precedence. But if the column rules don't have an answer then the default rules are used. "overriding" doesn't make that as clear as "in addition to".

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
```diff
@@ -34,66 +36,122 @@
...
+   * In most cases, there is no need to override the rules on a per-column basis.
+   */
   @PropertyDefinition(validate = "notNull")
   private final MarketDataRules marketDataRules;
+  /**
+   * The reporting currency that applies to this column, overriding the default reporting currency.
+   * The common case is to specify a default reporting currency
```
cjkent: The common case is to not specify anything on the individual columns and specify a reporting currency that applies to all columns.

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
```diff
@@ -34,66 +36,122 @@
...
-  private final ReportingRules reportingRules;
-
-  @ImmutableDefaults
-  private static void applyDefaults(Builder builder) {
-    builder.pricingRules(PricingRules.empty());
-    builder.marketDataRules(MarketDataRules.empty());
-    builder.reportingRules(ReportingRules.empty());
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance from the specified measure.
```
cjkent: "from the specified measure" painted an inaccurate picture in my mind of what was happening. As if the measure was somehow a factory for columns.

A column contains values of a measure.

modules/calc/src/main/java/com/opengamma/strata/calc/config/ReportingCurrency.java
```diff
@@ -23,63 +24,85 @@
...
-final class FixedReportingRules implements ReportingRules, ImmutableBean {
+public final class ReportingCurrency
+    implements ImmutableBean, Serializable {
 
   /**
-   * The reporting currency that is used for every target.
+   * An instance requesting the "natural" currency of the target.
+   * <p>
+   * When converting calculation results, conversion will occur to the "natural" currency of the target.
+   * The "natural" currency of a target is obtained from {@link CalculationFunction}.
```
cjkent: Link to the `naturalCurrency` method?

modules/calc/src/main/java/com/opengamma/strata/calc/config/ReportingCurrency.java
```diff
@@ -23,63 +24,85 @@
...
+import com.opengamma.strata.calc.runner.function.CalculationFunction;
 
 /**
- * Reporting rules that always returns the same currency.
+ * The reporting currency.
  * <p>
- * These rules always return the same currency from {@link #reportingCurrency}.
+ * This is used to specify the currency that the result should be reporting in.
+ * The currency specified may be explicit, using {@link #of(Currency)}, or implicit
+ * using {@link #NATURAL}.
```
cjkent: Probably worth mentioning where the natural currency comes from here as well as in the comments for `NATURAL`.

modules/calc/src/main/java/com/opengamma/strata/calc/config/ReportingCurrency.java
```diff
@@ -23,63 +24,85 @@
...
+   * When converting calculation results, conversion will occur to the "natural" currency of the target.
+   * The "natural" currency of a target is obtained from {@link CalculationFunction}.
+   */
+  public static final ReportingCurrency NATURAL = new ReportingCurrency(ReportingCurrencyType.NATURAL, null);
+
+  /**
+   * The type of reporting currency.
    */
   @PropertyDefinition(validate = "notNull")
+  private final ReportingCurrencyType type;
```
cjkent: This feels redundant. The caller already has two ways to find out whether a specific currency is avilable:

    reportingCurrency == ReportingCurrency.NATURAL;

and

    resportingCurrency.getCurrency().isPresent();

I don't think we need this too.

modules/calc/src/main/java/com/opengamma/strata/calc/config/ReportingCurrencyType.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.common.base.CaseFormat;
+import com.opengamma.strata.basics.currency.Currency;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * The available types of reporting currency.
+ * <p>
+ * There are two options - 'Natural' and 'Specific'.
+ */
+public enum ReportingCurrencyType {
```
cjkent: I don't think this adds enough to justify exsting.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTask.java
```diff
@@ -201,14 +198,11 @@ public MarketDataRequirements requirements() {
...
    * Returns an optional containing the first currency from the arguments or empty if both arguments are empty.
    */
-  private static Optional<Currency> reportingCurrency(Optional<Currency> ccy1, Optional<Currency> ccy2) {
-    if (ccy1.isPresent()) {
-      return ccy1;
-    }
-    if (ccy2.isPresent()) {
-      return ccy2;
+  private Optional<Currency> reportingCurrency() {
+    if (reportingCurrency.getCurrency().isPresent()) {
```
cjkent: Could this be

    return reportingCurrency.getCurrency().flatMap(function.naturalCurrency(target));

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -102,28 +102,27 @@ public static CalculationTasks of(
    * @param rowIndex  the row index of the value in the results grid
    * @param columnIndex  the column index of the value in the results grid
    * @param target  the target for which the measure will be calculated
-   * @param column  the column for which the value is calculated
+   * @param effectiveColumn  the effective column with all rules merged
```
cjkent: How about "the effective column containing the column rules merged with the default rules"?

cjkent:
That's everything

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTask.java
jodastephen: No, the flatMap() function isn't invoked if the optional is empty. There is no pretty way to do this, I tried.

modules/calc/src/main/java/com/opengamma/strata/calc/config/ReportingCurrency.java
jodastephen: Without it, information is implied from the state without being explicit. We follow a similar pattern in some other places, where the type is not strictly necessary but adds clarity when observing the state (this isn't an case where every last piece of memory matters either). It also provides for future extensibility, should a need for PAY_LEG currency or similar be needed. While it is an extra class, the complexity is low and I consider the benefits to be well justified.

cjkent: If we need a PAY_LEG currency then we can add it then. Having 3 ways to achieve the same thing strikes me as potentially confusing.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTask.java
cjkent: You're right, I was thinking about it backwards. It would need a non-existent variation of `orElse` returning an `Optional`.

PR #755, opened 2016-01-29
MapStream enhancements

cjkent:
Add an override to `MapStream.of` to create a stream from a collection of values and a function which generates a key from each value.

Add an override to `MapStream.toMap` and `Guavate.toImmutableMap` which handles multiple mappings to the same key by applying a merging function to the values.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -6,6 +6,7 @@
 package com.opengamma.strata.collect;
 
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -588,4 +621,64 @@ public R next() {
...
+      BiFunction<? super V, ? super V, ? extends V> mergeFn) {
+
+    V existingVal = map.get(key);
+
+    if (existingVal == null) {
+      map.put(key, val);
+    } else {
+      map.put(key, mergeFn.apply(existingVal, val));
+    }
+  }
```
jodastephen: Isn't this the same as `Map.merge()` ?

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -588,4 +621,64 @@ public R next() {
...
+   * the new value, and the return value is inserted.
+   *
+   * @param map1  the map into which values are copied
+   * @param map2  the map from which values are copied
+   * @param mergeFn  function applied to the existing and new values if the map contains the key
+   * @param <K>  the key type
+   * @param <V>  the value type
+   * @return {@code map1} with the values from {@code map2} inserted
+   */
+  private static <K, V> Map<K, V> mergeMaps(
```
jodastephen: It would be clearer that this mutates if it returned `void`.

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
cjkent: You're right. It seems my mental Javadocs are still pointing to Java 7.

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
cjkent: I'm at the mercy of the signature of `Collector.of` which required a `BinaryOperator`.

Issue #740, opened 2016-01-25
Add flag to Measure to control FX conversion


cjkent:
By default the system always tries to convert any value it can into the reporting currency. In some cases this is not what the user wants. For example, PV for instruments with multiple currencies is always converted into a single currency, but in some use cases the separate currency amounts are required.

A flag is needed on `Measure` to enable or disable automatic currency conversion. Different measures will need to be defined for the two cases, for example `PRESENT_VAUE` and `PRESENT_VALUE_MULTI_CURRENCY`.

Issue #562, opened 2015-11-17
Add type information to Measure


cjkent:
It would be useful if `Measure` contained information about the type of value it produced.

This needs some thought, it might not be as simple as it seems on the surface because of scenarios. Some measures, for example PV, produce a single value for a single scenario (`double`) and multiple values for multiple scenarios (a subtype of `ScenarioResult<Double>`). However it's easy to imagine measures that produce the same type whether there is one scenario or many. For example the maximum or average PV across multiple scenarios is a `double` regardless of the number of scenarios.

Does this mean `Measure` requires 2 type parameters? Or that there needs to be 2 types of measure, one for individual values and one for values calculated across multiple scenarios?

jonathansenior:
I think the effect that scenarios might have on the type probably belongs elsewhere, if anywhere. At the static Measure level I think what's important is to convey that a raw PV value is a `double`, while bucketed PV01 is another type, etc.

cjkent:
One of the major reasons for having the type on the measure is knowing what type to expect when getting values from the results. This is OK for PV where the value for multiple scenarios is just multiple instance of the value for a single scenario. But it breaks down for multiple scenarios if a function does anything more complicated than simply populating a list with the value from each scenario.

I don't think it is sufficient to just have a single type parameter on `Measure` unless we introduce a different type of measure for values aggregated across multiple scenarios.

We might also have to look at `Results`. Currently the same object is used to return results from one scenario or many and it is up to the user to know whether they're getting a single or multiple values. For the type information on `Measure` to be useful we might need different `Results` types.

cjkent:
This is potentially useful for controlling currency conversion. `FxConvertible` results will always be converted into a single currency amount at the moment and that might not always be what the users wants. If the measure specified the result had a type with multiple currencies the engine can skip the currency conversion.

jodastephen:
After much thought, it was concluded that the impact on usability would be greater than the benefits. Some convenience methods to avoid casting were added instead.

PR #756, opened 2016-02-01
Convert Measure into an ExtendedEnum

cjkent:
Convert measure to an `ExtendedEnum` so it can contain state but still be referred to by name.

This is in preparation for tackling #740 (FX conversion flag on `Measure`) and #562 (add types to `Measure`).

The important changes are in:

    Measure
    Measures
    StandardMeasures
    ImmutableMeasure
    Measures.ini

The remaining changes are caused by moving the standard measure constants from `Measure` to `Measures`.

The standard measures are defined in code in `StandardMeasures`. Defining a text format for measures is beyond the scope of these changes. If we decide we need one it would be more sensible to think about it once we've made all the changes to `Measure`.


modules/calc/src/main/java/com/opengamma/strata/calc/config/Measure.java
```diff
@@ -22,136 +22,30 @@
...
-  /**
-   * Obtains an instance from the specified name.
-   * <p>
-   * Measure names must only contains the characters A-Z, a-z, 0-9 and -.
+   * Obtains an instance from the specified unique name.
    *
-   * @param name  the name of the measure
-   * @return a measure with the specified name
+   * @param uniqueName  the unique name
+   * @return the convention
```
jodastephen: Not a convention

modules/calc/src/main/java/com/opengamma/strata/calc/config/Measure.java
```diff
@@ -22,136 +22,30 @@
...
+    return extendedEnum().lookup(uniqueName);
   }
 
   /**
-   * Creates an instance.
-   * 
-   * @param name  the name of the measure
+   * Gets the extended enum helper.
+   * <p>
+   * This helper allows instances of the convention to be looked up.
```
jodastephen: Not a convention

modules/calc/src/test/java/com/opengamma/strata/calc/config/pricing/PricingRuleTest.java
```diff
@@ -30,36 +31,32 @@
...
-  private static final Measure MEASURE1 = Measure.of("1");
-  private static final Measure MEASURE2 = Measure.of("2");
-  private static final Measure MEASURE3 = Measure.of("3");
-
   private static final FunctionGroup<TestTrade1> GROUP =
       DefaultFunctionGroup.builder(TestTrade1.class)
           .name("GroupName")
-          .addFunction(MEASURE1, FunctionConfig.of(TestFunction1.class))
-          .addFunction(MEASURE2, FunctionConfig.of(TestFunction2.class))
+          .addFunction(Measures.PRESENT_VALUE, FunctionConfig.of(TestFunction1.class))
```
jodastephen: Unfortunate that the static imports have been lost. Must also have left the file with an unused static import. Comment applies to numerous files.

cjkent: They haven't been lost, I've just replaced a locally defined constant with an externally defined constant.

Previously it was possible to define a local measure with an arbitrary name. Now they have to match the built in set of measures. In which case it makes sense to use the standard measures rather then defining local aliases.

PR #757, opened 2016-02-01
Add FX conversion flag to Measure

cjkent:
Add a flag to `Measure` specifying whether the calculated values should be automatically converted to the reporting currency.

Add a measure `PRESENT_VALUE_MULTI_CCY` representing present value with no currency conversion applied.

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/config/ImmutableMeasure.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/config/ImmutableMeasure.java
@@ -45,6 +45,12 @@ public final class ImmutableMeasure
   @PropertyDefinition(validate = "notNull", overrideGet = true)
   private final String name;
 
+  /**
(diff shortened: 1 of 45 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/config/ImmutableMeasure.java
```diff
@@ -99,9 +119,12 @@ public String toString() {
   }
 
   private ImmutableMeasure(
-      String name) {
+      String name,
+      boolean currencyConvertible) {
     JodaBeanUtils.notNull(name, "name");
+    JodaBeanUtils.notNull(currencyConvertible, "currencyConvertible");
```
jodastephen: Not null `boolean`?

cjkent: Well you can't be too careful :)

I've got a template for bean properties which adds `notNull` and I forgot to delete it.

Issue #542, opened 2015-11-10
Libor Cap/Floor - Pricer: Black and Normal models


marc-opengamma:
Implement Black (log-normal) and Bachelier (normal) models for vanilla cap/floor (fixing in advance, payment in-arrears).

PR #762, opened 2016-02-02
Time and zone for ibor index

yukiiwashita:
* Added LocalTime and ZoneId to IborIndex
* This work is required by #542


modules/basics/src/main/java/com/opengamma/strata/basics/index/IborIndex.java
```diff
@@ -81,6 +83,24 @@ public static IborIndex of(String uniqueName) {
   public abstract HolidayCalendar getFixingCalendar();
 
   /**
+   * Gets the fixing time of the index.
```
jodastephen: Remove these two methods from `IborIndex` and add `calculateFixingDateTime(LocalDate)` instead (returning `ZonedDateTime`). This avoids exposing the data over the `IborIndex` interface.

modules/basics/src/main/java/com/opengamma/strata/basics/index/ImmutableIborIndex.java
```diff
@@ -65,6 +67,20 @@
   @PropertyDefinition(validate = "notNull", overrideGet = true)
   private final HolidayCalendar fixingCalendar;
   /**
+   * The fixing time.
```
jodastephen: These properties are correct, but they will no longer have `overrideGet`

modules/product/src/main/java/com/opengamma/strata/product/rate/IborRateObservation.java
```diff
@@ -75,6 +76,15 @@ public void collectIndices(ImmutableSet.Builder<Index> builder) {
     builder.add(index);
   }
 
+  /**
```
jodastephen: This method is not needed with the new `calculateFixingDateTime(LocalDate)` method on the index

PR #766, opened 2016-02-02
Provide ability to encode scheme

jodastephen:
Indirectly allows URIs as schemes

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/id/StandardId.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/id/StandardId.java
@@ -28,6 +28,7 @@ import org.joda.convert.FromString;
 import org.joda.convert.ToString;
 
 import com.google.common.collect.ComparisonChain;
+import com.google.common.net.PercentEscaper;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/id/StandardId.java
```diff
@@ -117,6 +122,18 @@ public static StandardId parse(String str) {
...
+  /**
+   * Encode a string suitable for use as the scheme.
+   * <p>
+   * This uses percent encoding, just like URI.
+   * 
+   * @param scheme  the scheme to encode
+   * @return the encoded scheme
+   */
+  public static String encodeScheme(String scheme) {
+    return SCHEME_ESCAPER.escape(scheme);
```
jonathansenior: There's probably an implication that the result is a valid scheme, but the scheme regex mandates one letter to begin with, which it looks like this wouldn't enforce.

jodastephen: Altered regex to remove restriction.

PR #767, opened 2016-02-03
Enhance FpML parser

jodastephen:
Add party selector for more flexible selection
Add trade info parser to provide per-trade attribute/id capture
Remove unused code and increase test coverage


jonathansenior:
Looks great - seems to cover everything I can think of.

PR #771, opened 2016-02-04
Parse FpML inflation swaps

jodastephen:


modules/loader/src/main/java/com/opengamma/strata/loader/impl/fpml/SwapFpmlParserPlugin.java
```diff
@@ -326,131 +373,164 @@ private RateCalculation parseSwapCalculation(
...
+    document.validateNotPresent(inflationEl, "floorRateSchedule");
+    document.validateNotPresent(legEl, "stubCalculationPeriodAmount");  // TODO: parse fixed stub rate
+    InflationRateCalculation.Builder builder = InflationRateCalculation.builder();
+    // index
+    builder.index(document.parsePriceIndex(inflationEl));
+    // lag
+    builder.lag(document.parsePeriod(inflationEl.getChild("inflationLag")));
+    // interpolation
+    String interpStr = inflationEl.getChild("interpolationMethod").getContent();
+    builder.interpolated(interpStr.toLowerCase(Locale.ENGLISH).contains("linear") ? true : false);
```
jonathansenior: Just use the condition directly as a boolean expression

PR #773, opened 2016-02-05
Add MapStream.of(Stream) and CurveGroup.stream()

cjkent:


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -58,7 +58,7 @@
    * applying a function to each value.
    *
    * @param collection  the collection of values
-   * @param keyFunction  a function that returns the key for a value
+   * @param keyFunction  a function which returns the key for a value
```
jodastephen: My reading of [grammar](http://www.kentlaw.edu/academics/lrw/grinker/LwtaThat_Versus_Which.htm) [rules](http://www.quickanddirtytips.com/education/grammar/which-versus-that-0) suggests "that" is correct here.;-)

cjkent: This claims both versions are correct

http://www.oxforddictionaries.com/words/that-or-which

PR #776, opened 2016-02-08
Simplify reference data

jodastephen:
More closely match design of market data
No need for more advanced design at the moment
Add dedicated exception type


modules/basics/src/main/java/com/opengamma/strata/basics/market/ReferenceData.java
```diff
@@ -12,26 +12,28 @@
...
  */
 public interface ReferenceData {
 
   /**
-   * Obtains an instance from a map of typed reference data.
+   * Obtains an instance from a map of reference data.
    * <p>
-   * Each identifier in the map refers to an instance of {@link TypedReferenceData}.
-   * This allows multiple values to be connected to the same identifier, accessed by type.
+   * Each entry in the map is a single piece of reference data, keyed by the matching identifier.
```
jonathansenior: Perhaps "a _unit_ of reference data" might make the holiday calendar example clearer. I'd think of a _piece_ of data as something finer grained.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableReferenceData.java
```diff
@@ -44,21 +45,52 @@
...
+    // validation handles edge case where input by raw or polluted types
+    validateEntry(id, value);
+    return new ImmutableReferenceData(ImmutableMap.of(id, value));
+  }
+
+  // validates a single entry
+  private static <T> void validateEntry(ReferenceDataId<?> id, Object value) {
+    if (!id.getReferenceDataType().isInstance(value)) {
+      throw new ClassCastException(Messages.format(
+          "Value does not implement parameterized type for identifier '{}' of type '{}'", id, id.getClass().getSimpleName()));
```
jonathansenior: Include the value or its type in the message?

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableReferenceData.java
```diff
@@ -44,21 +45,52 @@
...
+   */
+  public static <T> ImmutableReferenceData of(ReferenceDataId<T> id, T value) {
+    // validation handles edge case where input by raw or polluted types
+    validateEntry(id, value);
+    return new ImmutableReferenceData(ImmutableMap.of(id, value));
+  }
+
+  // validates a single entry
+  private static <T> void validateEntry(ReferenceDataId<?> id, Object value) {
+    if (!id.getReferenceDataType().isInstance(value)) {
```
jonathansenior: Would be true if `value` is `null`, where throwing a `ClassCastException` would seem a bit misleading. Add specific handling or Javadoc covering null values?

jonathansenior:
That's everything.

PR #781, opened 2016-02-12
Enhance curve calibration

### LINKED ISSUE #780: Move CurveCalibrator and RootFinderConfig to MarketDataConfig
`CurveGroupMarketDataFunction` refers directly to the `CurveCalibrator` which refers to the `RootFinderConfig`. This means it would be necessary to create a whole new `MarketDataFactory` just to change the root finder configuration which is clearly wrong.

Root finder configuration doesn't exactly fit the existing model; in almost all cases users will use the default configuration and probably shouldn't have to specify it, but they should be able to easily override the defaults for a particular calculation.

jodastephen:
Change static constants to static methods
Consistent naming
Better factory methods
Allow root finder config to be overriden
Fixes #780


modules/function/src/main/java/com/opengamma/strata/function/marketdata/curve/CurveGroupMarketDataFunction.java
```diff
@@ -31,44 +32,42 @@
...
+ * <p>
+ * This function calibrates curves, turning a {@link CurveGroupDefinition} into a {@link CurveGroup}.
  */
 public class CurveGroupMarketDataFunction implements MarketDataFunction<CurveGroup, CurveGroupId> {
 
-  /** The analytics object that performs the curve calibration. */
+  /**
+   * The default analytics object that performs the curve calibration.
+   */
   private final CurveCalibrator curveCalibrator;
```
cjkent: I'm wondering if having the curve calibrator here is confusing. Would it be cleaner if the calibrator were always built using the root finder config from the `MarketDataConfig`. The logic to use the default root finder config would go in `MarketDataConfig` or `MarketDataConfigBuilder`.

cjkent: So we could get rid of the `curveCalibrator` field and the constructor taking a calibrator.

cjkent:
I'm happy to merge, I've run the tests locally and they all pass.

PR #786, opened 2016-02-15
Enhance CsvFile

jodastephen:
Add `CsvRow` to better manage each row
Allow for TSV files

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
@@ -6,32 +6,32 @@
 package com.opengamma.strata.collect.io;
 
 import static com.opengamma.strata.collect.Guavate.toImmutableList;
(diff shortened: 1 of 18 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
```diff
@@ -84,21 +101,21 @@ public static CsvFile of(CharSource source, boolean headerRow) {
...
+  private static Stream<ImmutableList<String>> parseLine(String line, char separator) {
     if (line.length() == 0 || line.startsWith("#") || line.startsWith(";")) {
       return Stream.empty();
     }
     ImmutableList.Builder<String> builder = ImmutableList.builder();
     int start = 0;
-    String terminated = line + ',';
-    int nextComma = terminated.indexOf(',', start);
+    String terminated = line + separator;
+    int nextComma = terminated.indexOf(separator, start);
```
jonathansenior: Probably best to avoid the word 'comma' given these changes.

Issue #783, opened 2016-02-14
List generation inefficiency


fmovlex:
I've noticed this issue in `PeriodicSchedule.java` - it probably exists in other spots as well.
Take for example `PeriodicSchedule#generateBackwards` (line 493 or 499 on my version) - and consider a non standard internal swap (e.g. float-float, pay 1m, over 50y, pay/rec legs -> ~1200 payment periods).

Using an `ArrayList` copies a giant amount of increasingly-sized arrays - and in a portfolio with hundreds of these swaps (or during HVaR) things get messy.
Using an `ArrayDeque` here and optimizing inserts for head/tail will really help here - for example, instead of adding `end` and `start` (which forces you to insert at position 1), use a `Deque` and head-insert everything, then head-insert `start` last.

I'm not really familiar with anything except swaps so I hesitate to change things myself, but a review of similar list-generating methods (and especially backward-generating) across the library would likely boost performance.
Thanks guys

jonathansenior:
Thanks for raising this. Looking at the code, I can see the inefficiency you describe due to the repeated insertion at position 1. Using an `ArrayDeque` and inserting only at the head would work nicely - thanks for the suggestion.

I couldn't find any other instances of similar logic, but this issue will affect all products with initial stubs. Out of interest, did you find this by profiling?

fmovlex:
We had a similar performance hit in my firm's swap pricer - it just caught my eye while looking through the `leg#expand` process.

There's a whole bunch of performance "gotchas" we've gathered from pricing swaps and especially calculating VaR which I'll happily share whenever I can - but since my view is limited to interest rates at the moment, it'll take some learning.

jodastephen:
My testing didn't find a major slowdown as a result of the insertion at index 1 pattern. I've raised a R to address it, using a private `List` implementation that acts like `ArrayDeque`.

PR #788, opened 2016-02-16
Enhance schedule calculation

jodastephen:
Use dedicated list implementation for backwards looping
Additional methods on Frequency

This is based on issue #783. My testing did not yield any significant performance issue to be solved. However, the principal of the issue seems sound. This PR adds a private list implementation that acts like `ArrayDeque`, but implements `List`. Other logic is moved around or optimized to avoid unnecessary calculation.

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/Frequency.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/Frequency.java
@@ -11,6 +11,7 @@ import static java.time.temporal.ChronoUnit.YEARS;
 
 import java.io.Serializable;
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.time.DateTimeException;
 import java.time.LocalDate;
 import java.time.Period;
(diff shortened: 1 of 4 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/Frequency.java
```diff
@@ -225,10 +252,25 @@ public static Frequency ofWeeks(int weeks) {
...
+        return P3M;
+      case 4:
+        return P4M;
+      case 6:
+        return P6M;
+      case 12:
+        return P12M;
+      default:
+        if (months > MAX_MONTHS) {
+          throw new IllegalArgumentException("Months must not exceed 12,000");
```
cjkent: Should use the value of `MAX_MONTHS` in the error message rather than hard-coding to 12,000.

modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
```diff
@@ -458,20 +459,12 @@ public Schedule createSchedule() {
...
-      unadj.add(0, startDate);
-    }
-    if (explicitFinalStub) {
-      unadj.add(endDate);
-    }
-    // handle override start date
-    unadj.set(0, effectiveStartDate);
-    return unadj;
+    StubConvention stubConv = generateImplicitStubConvention(explicitInitialStub, explicitFinalStub);
+    return (stubConv.isCalculateBackwards() ?
```
cjkent: Unnecessary parens around the whole return expression.

PR #796, opened 2016-02-19
Rate observations for inflation bond

yukiiwashita:
* Created rate observations (and corresponding rate observation functions) which store start index value. 
* This work is required by #784


jodastephen:
Renamed the classes to avoid the name "Bond" as they are not really bond specific.

PR #797, opened 2016-02-23
Replace containsValue with findValue on market data containers

cjkent:
Replace the boolean-returning `containsValue` methods on the market data containers with `findValue` returning an `Optional`. This is more consistent with the approach elsewhere in the codebase, and it gets rid of the need for two map lookups in the cases where a call to `containsValue` is immediately followed by a call to `getValue`.

The `containsTimeSeries` method has been removed. Originally the behaviour of `getTimeSeries` was the same as `getValue` in that it threw an exception if there was no data available. That was changed so `getTimeSeries` returns an empty series if no data is available. That made `containsTimeSeries` redundant, but it wasn't removed at the time.

### DIFF
--- a/examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
+++ b/examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
@@ -7,6 +7,7 @@ package com.opengamma.strata.examples.marketdata;
 
 import static com.opengamma.strata.collect.TestHelper.assertThrowsIllegalArg;
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
 
 import java.io.File;
@@ -242,14 +243,14 @@ public class ExampleMarketDataBuilderTest {
     assertEquals(MARKET_DATA_DATE, snapshot.getValuationDate().getSingleValue());
 
     for (ObservableId id : TIME_SERIES) {
-      assertTrue(snapshot.containsTimeSeries(id), "Time-series not found: " + id);
+      assertFalse(snapshot.getTimeSeries(id).isEmpty(), "Time-series not found: " + id);
     }
     assertEquals(snapshot.getTimeSeries().size(), TIME_SERIES.size(),
         Messages.format("Snapshot contained unexpected time-series: {}",
             Sets.difference(snapshot.getTimeSeries().keySet(), TIME_SERIES)));
 
     for (MarketDataId<?> id : VALUES) {
-      assertTrue(snapshot.containsValue(id), "Id not found: " + id);
+      assertTrue(snapshot.findValue(id).isPresent(), "Id not found: " + id);
     }
     MarketDataBox<CurveGroup> curveGroupBox = snapshot.getValue(CurveGroupId.of(DEFAULT_CURVE_GROUP));
     assertTrue(curveGroupBox.isSingleValue());
--- a/examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
+++ b/examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
@@ -250,7 +250,7 @@ public class ExampleMarketDataBuilderTest {
             Sets.difference(snapshot.getTimeSeries().keySet(), TIME_SERIES)));
 
     for (MarketDataId<?> id : VALUES) {
-      assertTrue(snapshot.findValue(id).isPresent(), "Id not found: " + id);
+      assertTrue(snapshot.containsValue(id), "Id not found: " + id);
     }
     MarketDataBox<CurveGroup> curveGroupBox = snapshot.getValue(CurveGroupId.of(DEFAULT_CURVE_GROUP));
     assertTrue(curveGroupBox.isSingleValue());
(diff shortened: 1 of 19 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/market/MarketData.java
```diff
@@ -60,12 +61,12 @@ public static MarketData of(
 
   //-------------------------------------------------------------------------
   /**
-   * Checks if this set of data contains a value for the specified key.
+   * Returns a box containing values for the specified ID if available.
```
jodastephen: "box" has a different meaning in other parts of the MD system. `ReferenceData` uses:

    * Finds the reference data value associated with the specified identifier.
    * <p>
    * If this reference data instance contains the identifier, the value will be returned.
    * Otherwise, an empty optional will be returned.

examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
```diff
@@ -242,14 +243,14 @@ private void assertBuilder(ExampleMarketDataBuilder builder) {
     assertEquals(MARKET_DATA_DATE, snapshot.getValuationDate().getSingleValue());
 
     for (ObservableId id : TIME_SERIES) {
-      assertTrue(snapshot.containsTimeSeries(id), "Time-series not found: " + id);
+      assertFalse(snapshot.getTimeSeries(id).isEmpty(), "Time-series not found: " + id);
```
jodastephen: This isn't the same logic. Not sure it matters, but cannot distinguish between time series that is present but empty and one that is absent.

examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
```diff
@@ -242,14 +243,14 @@ private void assertBuilder(ExampleMarketDataBuilder builder) {
...
     }
     assertEquals(snapshot.getTimeSeries().size(), TIME_SERIES.size(),
         Messages.format("Snapshot contained unexpected time-series: {}",
             Sets.difference(snapshot.getTimeSeries().keySet(), TIME_SERIES)));
 
     for (MarketDataId<?> id : VALUES) {
-      assertTrue(snapshot.containsValue(id), "Id not found: " + id);
+      assertTrue(snapshot.findValue(id).isPresent(), "Id not found: " + id);
```
jodastephen: When testing reference data I got a speed up of 15% or so by directly implementing `getValue` not going via `findValue`. Thus, there is a performance question to consider.

modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/DefaultMarketDataFactory.java
```diff
@@ -166,21 +165,21 @@ public MarketEnvironment buildMarketData(
 
       // Build any time series that are required but not available
       leafRequirements.getTimeSeries().stream()
-          .filter(not(marketData::containsTimeSeries))
-          .filter(not(suppliedData::containsTimeSeries))
+          .filter(id -> marketData.getTimeSeries(id).isEmpty())
+          .filter(id -> suppliedData.getTimeSeries(id).isEmpty())
```
jodastephen: The change in meaning seems significant to this logic. No way to prevent query for time-series just by putting empty one in input.

jodastephen:
To summarize, I think removing the two contains methods is unwise. Adding an additional find method is fine.

modules/basics/src/main/java/com/opengamma/strata/basics/market/MarketData.java
cjkent: That's a copy-paste error from the version of `findValue` returning a `MarketDataBox`

examples/src/test/java/com/opengamma/strata/examples/marketdata/ExampleMarketDataBuilderTest.java
cjkent: We gave up the ability to distinguish that when we changed `getTimeSeries` to return an empty series for an unknown ID. That was a deliberate decision. Having `containsTimeSeries` makes no sense after that.

modules/calc/src/main/java/com/opengamma/strata/calc/marketdata/DefaultMarketDataFactory.java
cjkent: Do we have a requirement to do that? Why would we want to?

cjkent:
I don't think there's any debate around removing `containsTimeSeries`. It makes no sense where `getTimeSeries` returns an empty series for an unknown ID.

Having `containsValue` is inconsistent with the approach elsewhere. It was a performance issue that made me consider removing it; calling `containsValue` followed by `getValue` is two map lookups. `findValue` trades a lookup for the cost of putting the value in an `Optional`.

If we find it's a problem it's trivial to add `containsValue`. But we can't easily remove it later and it adds API surface area.

jodastephen:
This PR does not demonstrate the case for the change. Outside the tests, there are 8 `containsValue` changes and 4 `containsTimeSeries` changes. All 12 are clearer using `contains` than `find` (and also faster). There is not a single case where `find` is the better option in the PR.

cjkent:
`containsTimeSeries` is meaningless now. A container always returns a time series for any key or ID. If the caller wants to know if there is data available there is the `isEmpty` method on the time series.

There were two reasons for the change to `containsValue`

1. We've been generally been removing `contains*` in favour of `find*` across the codebase

2. I have a use case in an upstream project where the market data mappings are decorated to allow equivalent market data to be substituted if the requested data is unavailable. This entails a call to `containsValue` immediately followed by a call to `getValue`. This caused a measurable performance hit. It is also an anti-pattern when dealing with maps but the existing API forces you into it.

PR #802, opened 2016-02-29
Capital indexed bond: trade model and pricer

### LINKED ISSUE #784: Inflation-indexed bond: product and trade

### LINKED ISSUE #785: Inflation-indexed bond: pricer

yukiiwashita:
* This work closes #784 
* This work closes #785


modules/product/src/main/java/com/opengamma/strata/product/bond/YieldConvention.java
```diff
@@ -21,7 +21,8 @@
  * specific to each yield convention. Accordingly, the computation of price, convexity
  * and duration from the yield should be based on this yield convention. 
  * <p>
- * Reference: "Bond Pricing", OpenGamma Documentation 5, Version 2.0, May 2013.
+ * References: "Bond Pricing", OpenGamma Documentation 5, Version 2.0, May 2013, 
+ * "Inflation Instruments: Swap Zero-coupon, Year-on-year and Bonds."
  */
 public enum YieldConvention {
```
marc-opengamma: Do we want a single YieldConvention with all the conventions: Fixed Coupon bonds, Inflation, Bills, etc. or does it make sense to have different enum?

modules/product/src/main/java/com/opengamma/strata/product/bond/CapitalIndexedBondTrade.java
```diff
@@ -0,0 +1,515 @@
...
+  @PropertyDefinition(validate = "notNull", overrideGet = true)
+  private final SecurityLink<CapitalIndexedBond> securityLink;
+  /**
+   * The quantity, indicating the number of bond contracts in the trade.
+   * <p>
+   * This will be positive if buying and negative if selling.
+   */
+  @PropertyDefinition
+  private final long quantity;
+
```
marc-opengamma: Do we need a "payment" of some type to represent the settlement (price, date, etc.)?

modules/product/src/main/java/com/opengamma/strata/product/bond/YieldConvention.java
jodastephen: When I looked at OG-Platform Biill yield conventions had nothing in common with bond ones, so I think separate types makes sense.

More broadly, I don't like the `if(convention = FOO)` style, as really the logic should be in the convention, but that is a bigger change.

yukiiwashita:
This work is closed by #835

PR #806, opened 2016-03-02
Holiday calendars from Reference data

jodastephen:
With this PR, holiday calendars are resolved from `ReferenceData` instead of at startup via an extended enum. This allows holiday calendar data to be altered and managed more dynamically. The standard holiday calendars can still be accessed by passing `ReferenceData.standard()` where necessary.

- the "expanded" trade model concept has been renamed to "resolved"
- most asset classes have 4 key classes - `FraTrade` holding `Fra`, and `ResolvedFraTrade` holiding `ResolvedFra`
- Unresolved trades, indices and conventions refer to holidays using `HolidayCalendarId`
- conversion from the unresolved to the resolved form requires reference data and performs as much date logic as possible (resolution converts `HolidayCalendarId` to `HolidayCalendar`)
- market data is looked up using index observation objects, with pre-calculated dates

The overall design now has three lifecycle phases for trades

1. Conventions and Templates, as seen in `CurveNode`
2. Unresolved trades, the standard application definition of a trade
3. Resolved trades, low level form for pricing

Conversion between phase 1 and 2 uses reference data to calculate the spot date from the trade date. Conversion between phase 2 and 3 uses reference data to lock all the dates for pricing.

Future work will remove the `SecurityLink` concept, with securities also being resolved from reference data.


jonathansenior:
Reviewed in person with @jodastephen and @cjkent. Performance impact verified as acceptable with nightly benchmarks.

PR #818, opened 2016-03-08
Periodic curve: parameter sensitivity

marc-opengamma:
- Corrected parameter sensitivity for periodic curve. Unit sensitivity now represent the sensitivity of the zero-rate to the curve parameter. Improved tests.
- Clarified quantile calculation javadoc.


modules/market/src/main/java/com/opengamma/strata/market/view/ZeroRatePeriodicDiscountFactors.java
```diff
@@ -202,7 +203,12 @@ public ZeroRateSensitivity zeroRatePointSensitivityWithSpread(
   @Override
   public CurveUnitParameterSensitivities unitParameterSensitivity(LocalDate date) {
     double relativeYearFraction = relativeYearFraction(date);
-    return CurveUnitParameterSensitivities.of(curve.yValueParameterSensitivity(relativeYearFraction));
+    double rp = curve.yValue(relativeYearFraction);
+    //    double rc = _compoundingPeriodsPerYear * Math.log(1 + rp / _compoundingPeriodsPerYear);
```
yukiiwashita: Remove this line or elaborate this to an implementation note on the relation between the two rates.

yukiiwashita:
For completeness we also need a proper implementation of zeroRatePointSensitivityWithSpread.

PR #821, opened 2016-03-09
Add SecurityId and convert Equity

jodastephen:
Add identifier compatible with reference data. New security trade model design. `ReferenceSecurity` will be renamed to `Security` once task complete.


modules/product/src/main/java/com/opengamma/strata/product/ReferenceSecurity.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * A single security that can be traded.
+ * <p>
+ * A security is one of the building blocks of finance, representing a fungible instrument that can be traded.
+ * This is intended to cover instruments such as listed equities and futures.
+ * It is intended that Over-The-Counter (OTC) instruments, such as an interest rate swap,
+ * are embedded directly within the trade, rather than handled as one-off securities.
+ * <p>
+ * A security may be referred to using a {@link SecurityId}.
+ * The identifier can be looked up in {@link ReferenceData}.
```
cjkent: Should this be "The security can be looked up in `ReferenceData` using the `SecurityId`?

modules/product/src/main/java/com/opengamma/strata/product/ReferenceSecurity.java
```diff
@@ -0,0 +1,147 @@
...
+import java.time.LocalDate;
+import java.util.Optional;
+
+import com.google.common.collect.ImmutableSet;
+import com.opengamma.strata.basics.currency.Currency;
+import com.opengamma.strata.basics.market.ReferenceData;
+import com.opengamma.strata.collect.Messages;
+
+/**
+ * A single security that can be traded.
```
cjkent: Why "single"? Is there any other kind of security?

modules/product/src/main/java/com/opengamma/strata/product/ReferenceSecurity.java
```diff
@@ -0,0 +1,147 @@
...
+   * Gets the currency that the security is traded in.
+   * 
+   * @return the trading currency
+   */
+  public abstract Currency getCurrency();
+
+  /**
+   * Gets the underlying security identifiers.
+   * <p>
+   * The set must contain all the security identifiers that this security refers to.
```
cjkent: Some more detail would be good about what "underlying" means here, including an example.

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,205 @@
...
+   * The identifier, expressed as a standard two-part identifier.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final StandardId standardId;
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance from a scheme and value.
+   * <p>
+   * The scheme must be non-empty and match the regular expression '{@code [A-Za-z0-9:/+.=_-]*}'.
```
cjkent: Could the restriction on the scheme be stated without requiring the user to parse a regular expression in their head?

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,205 @@
...
+   * @throws IllegalArgumentException if the scheme or value is invalid
+   */
+  public static SecurityId of(String scheme, String value) {
+    return of(StandardId.of(scheme, value));
+  }
+
+  /**
+   * Creates an instance from a standard two-part identifier.
+   * <p>
+   * This parses the identifier from the form produced by {@code toString()}
```
cjkent: Copy/paste error. This method doesn't parse anything.

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,205 @@
...
+   * Care must be taken when placing the resolved form in a cache or persistence layer.
+   * 
+   * @param <T>  the type of the security
+   * @param refData  the reference data, used to resolve the reference
+   * @param securityType  the type of the security
+   * @return the resolved security
+   * @throws ReferenceDataNotFoundException if the identifier is not found
+   * @throws ClassCastException if the identifier exists, but is not of the correct type
+   */
+  public <T extends ReferenceSecurity> T resolve(ReferenceData refData, Class<T> securityType) {
```
cjkent: Does this method add enough to justify its place on the API? Effectively it enables this:

    MySecurity security = securityId.resolve(refData, MySecurity.class);

instead of:

    MySecurity security = (MySecurity) securityId.resolve(refData);

Is it worth cluttering up the API for that?

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfoType.java
```diff
@@ -0,0 +1,59 @@
...
+  public static final SecurityInfoType<String> NAME = SecurityInfoType.of("Name");
+
+  /** Serialization version. */
+  private static final long serialVersionUID = 1L;
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance from the specified name.
+   * <p>
+   * Curve names may contain any character, but must not be empty.
```
cjkent: Copy/paste error - not a curve name

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
jodastephen: This is likely to be a key API for error messages (I've pushed a better message). It (or something similar) will be needed.

modules/product/src/main/java/com/opengamma/strata/product/SecurityProduct.java
```diff
@@ -0,0 +1,49 @@
...
+ * It will include any underlying securities.
+ * For example, a bond future option would include the underlying bond future
+ * and the basket of bonds.
+ * <p>
+ * The product of a security is distinct from the {@linkplain ReferenceSecurity security}.
+ * The security, referred to using a {@link SecurityId}, does not include underlying securities.
+ * <p>
+ * Implementations of this interface must be immutable beans.
+ */
+public interface SecurityProduct
```
cjkent: Why do we need `ReferenceSecurity` and `SecurityProduct`? There is one more level of indirection than I was expecting. In my mental model an equity trade contains a security ID and the ID refers to an equity (the thing that is currently called `EquityProduct`).

If I'm understanding this model correctly, the trade refers to a security ID, the ID refers to a `ReferenceSecurity` and the security refers to a `Product` implementation. Could `EquitySecurity` and `Equity` be collapsed into a single entity which is referenced by security ID and stored in the reference data?

jodastephen: In evaluating this, I attempted various things, however it turned out that the original way we considered of handling securities was too simple. Having a FooTrade contain a SecurityId that is resolved only at the "normal" resolution step does not allow ad hoc trades to be created as easily. It also requires a very large reference data setup, which is the opposite of what Strata strives for.

Normal (unresolved) trades directly contain the product and any underlying products, not the identifiers. Thus `IborFutureOptionTrade` holds `IborFutureOption` which holds `IborFuture`. ie. the only change from today is the removal of `SecurityLink`

Resolved trades are effectively unchanged.

Separately, new classes will be added for each security. `IborFutureOptionSecurity` and `IborFutureSecurity`. These contain identifiers for their underlying parts. The reference data map will contain these security classes, thus reference data is completely weakly coupled. Each security instance is capable of converting itself to either the product or the trade using reference data for identifier lookups.

A single `SecurityTrade` class can then be added, that has a `SecurityId`. When processed, this can be resolved in reference data, with `createTrade()` invoked to build the trade of the correct type. While we could add an `IborFutureSecurityTrade` (in addition to `IborFutureTrade`) there is no advantage over `SecurityTrade` in terms of object content.

Overall, it means that users that want to manage securities themselves can continue to do so. And users that want to put everything in reference data and weakly link will also be able to do so.

cjkent: Is the main reason for the extra level of indirection to retain the ability we have with `SecurityLink` to directly link the contract details from the trade without having to go via a reference data lookup?

I worry about creating a domain model that might make sense from an engineering point of view but not to someone with domain knowledge.

jodastephen: My discussions indicated that it was important to retain the direct link in the trades to the product. It also became apparent that only one trade class covers all securities, thus it also doesn't create too much code. (Unlike OTCs, there is a choice in representing security-based trades, and the design reflects that)

cjkent: I thought I was part of a conversation with exactly the opposite conclusion; that having an ID in the trades and a single indirection to get the security is the sort of design people would expect as it matches their mental model of the domain.

How would we describe a security to a domain expert? Would it match their understanding of the term "security"?

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,210 @@
...
+   * Resolved objects may be bound to data that changes over time.
+   * If the data changes, the resolved form will not be updated.
+   * Care must be taken when placing the resolved form in a cache or persistence layer.
+   * 
+   * @param refData  the reference data, used to resolve the reference
+   * @return the resolved security
+   * @throws ReferenceDataNotFoundException if the identifier is not found
+   */
+  @Override
+  public ReferenceSecurity resolve(ReferenceData refData) {
```
cjkent: It seems inconsistent and unnecessary to have the ID resolve itself on the data. All other IDs are resolved by calling a method on the data container, passing in the ID.

jodastephen: This matches holiday calendars.

cjkent: Maybe those should be brought into line too. Why have two ways to achieve the exact same thing? It adds confusion and inconsistency without any upside I can think of.

jodastephen:
This PR is being closed in favour of a new one.

PR #827, opened 2016-03-11
Change FxSwap factory to take FxRate

jodastephen:
Matches design of FxSingle and safer
Add method to FxRate to allow rate changes

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRate.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRate.java
@@ -10,7 +10,6 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
-import java.util.function.DoubleUnaryOperator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -233,23 +232,6 @@ public final class FxRate
    */
   public FxRate toConventional() {
     return pair.isConventional() ? this : FxRate.of(pair.toConventional(), 1 / rate);
-  }
-
-  //-------------------------------------------------------------------------
-  /**
-   * Returns an instance with the rate updated.
-   * <p>
-   * This returns an {@code FxRate} with the same currency pair and an updated rate.
-   * For example, the operator could be used to add the forward points of an FX swap.
-   * <pre>
-   *   forwardRate = nearRate.mapRate(rate -> rate + forwardPoints);
-   * </pre>
-   *
-   * @param operator  the operator to be applied to the existing rate
-   * @return a new instance with the rate updated
-   */
-  public FxRate mapRate(DoubleUnaryOperator operator) {
-    return new FxRate(pair, operator.applyAsDouble(rate));
   }
 
   //-------------------------------------------------------------------------
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRate.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRate.java
@@ -10,6 +10,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.function.DoubleUnaryOperator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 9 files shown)


modules/product/src/main/java/com/opengamma/strata/product/fx/FxSingle.java
```diff
@@ -138,9 +138,8 @@ public static FxSingle of(
   /**
    * Creates an {@code FxSingle} using a rate.
    * <p>
-   * This create an FX specifying a value date, notional in one currency, the second currency
-   * and the FX rate between the two.
-   * The currencies of the payments must differ.
+   * This create an FX specifying the amount, FX rate and value date.
```
cjkent: This sentence doesn't quite make sense. I suspect it was intended to be "This creates an `FxSingle` ...".

modules/product/src/main/java/com/opengamma/strata/product/fx/FxSingle.java
```diff
@@ -160,9 +159,8 @@ public static FxSingle of(CurrencyAmount amountCurrency1, FxRate fxRate, LocalDa
   /**
    * Creates an {@code FxSingle} using a rate, specifying a date adjustment.
    * <p>
-   * This create an FX specifying a value date, notional in one currency, the second currency
-   * and the FX rate between the two.
-   * The currencies of the payments must differ.
+   * This create an FX specifying the amount, FX rate and value date.
```
cjkent: Same as above

modules/product/src/main/java/com/opengamma/strata/product/fx/FxSwap.java
```diff
@@ -89,13 +90,13 @@ public static FxSwap of(FxSingle nearLeg, FxSingle farLeg) {
...
-   * The two currencies must not be equal.
+   * The two currencies are specified by the near FX rate.
+   * The amount must be specified using one of the currencies of the near FX rate.
    * The near date must be before the far date.
    * Conventions will be used to determine the base and counter currency.
    * 
    * @param amountCurrency1  the amount of the near leg in the first currency
-   * @param currency2  the second currency
-   * @param nearFxRate  the near FX rate, where {@code (1.0 * amountCurrency1 = fxRate * amountCurrency2)}
+   * @param nearRate  the near FX rate, where {@code (1.0 * amountCurrency1 = fxRate * amountCurrency2)}
```
cjkent: The second half of this comment no longer applies

modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxRate.java
```diff
@@ -236,6 +237,23 @@ public FxRate toConventional() {
...
+   * This returns an {@code FxRate} with the same currency pair and an updated rate.
+   * For example, the operator could be used to add the forward points of an FX swap.
+   * <pre>
+   *   forwardRate = nearRate.mapRate(rate -> rate + forwardPoints);
+   * </pre>
+   *
+   * @param operator  the operator to be applied to the existing rate
+   * @return a new instance with the rate updated
+   */
+  public FxRate mapRate(DoubleUnaryOperator operator) {
```
cjkent: I have a very bad feeling about this method; it goes completely against the intent of `FxRate`. `FxRate` was very specifically designed not to expose the rate field to users.

If you have two currencies `AAA` and `BBB` and an `FxRate` which converts between them, you don't know whether the numeric rate is `AAA/BBB` or `BBB/AAA` and you don't need to care. If you want to convert an amount from `AAA` to `BBB` you call

    rate.convert(amount, Currency.AAA, Currency.BBB);

If you need the numeric value of the rate for converting from `AAA` to `BBB` you call

    rate.fxRate(Currency.AAA, Currency.BBB);

The logic to invert the rate when necessary is in `FxRate`. This spares the user from ever having to think about the direction of FX rates which is fiddly and error-prone.

This method exposes the underlying numerical rate and the map operation has a completely different meaning depending on whether the rate is `AAA/BBB` or `BBB/AAA`.

In order to make this method consistent with the rest of the class it should probably have this signature:

    FxRate mapRate(Currency fromCurrency, Currency toCurrency, DoubleUnaryOperator operator);

We would have to decide what the currency pair should be in the returned `FxRate`. The obvious choices are:

* The same as the pair in the receiver
* `fromCurrency/toCurrency`
* The market conventional pair

jodastephen: I've removed the method. But really I feel we are making it too hard to access the rate.

cjkent: That is by design, as I said in the comment. If you need the rate as a double you just call FxRate.fxRate(base, counter). What’s hard about that? It’s also completely unambiguous.

PR #836, opened 2016-03-15
Minor trade model enhancements

jodastephen:
Remove `Attributable` as had no real value.
Restrict builder on rate observations.
Add additional reference data access methods.
Reorder properties in DSF
Enhance test coverage.


modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableReferenceData.java
```diff
@@ -96,7 +96,7 @@ public static ImmutableReferenceData of(Map<? extends ReferenceDataId<?>, ?> val
       }
       throw new ClassCastException(Messages.format(
           "Value '{}' does not implement parameterized type '{}' for identifier '{}'",
```
cjkent: Why "parameterized" in this sentence?

modules/basics/src/main/java/com/opengamma/strata/basics/market/ImmutableReferenceData.java
```diff
@@ -123,12 +123,17 @@ public boolean containsValue(ReferenceDataId<?> id) {
...
     if (value == null) {
-      throw new ReferenceDataNotFoundException(Messages.format(
-          "Reference data not found for identifier '{}' of type '{}'", id, id.getClass().getSimpleName()));
+      throw new ReferenceDataNotFoundException(msgValueNotFound(id));
     }
     return value;
   }
 
+  // extracted to aid inlining performance
+  private <T> String msgValueNotFound(ReferenceDataId<T> id) {
```
cjkent: Unused type parameter.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ReferenceData.java
```diff
@@ -89,6 +89,31 @@ public default boolean containsValue(ReferenceDataId<?> id) {
...
+   * @throws ReferenceDataNotFoundException if the identifier is not found
+   * @throws ClassCastException if the identifier exists, but is not of the correct type
+   */
+  @SuppressWarnings("unchecked")
+  public default <T, S extends T> S getValue(ReferenceDataId<T> id, Class<S> type) {
+    T value = getValue(id);
+    if (!type.isInstance(value)) {
+      throw new ClassCastException(Messages.format(
+          "Identifier '{}' resolved to a value of type '{}' where '{}' was expected",
+          this, value.getClass().getSimpleName(), type.getSimpleName()));
```
cjkent: The first argument should be `id`, not `this`.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ReferenceData.java
```diff
@@ -89,6 +89,31 @@ public default boolean containsValue(ReferenceDataId<?> id) {
...
   /**
+   * Gets the reference data value associated with the specified identifier, cast to a specific type.
+   * <p>
+   * If this reference data instance contains the identifier, and the value is of the correct type,
+   * then the value will be returned. Otherwise, an exception will be thrown.
+   * 
+   * @param <T>  the type of the reference data
+   * @param <S>  the type of the reference data subclass
+   * @param id  the identifier to find
+   * @param type  the type of the subclass
```
cjkent: "subclass" doesn't make things very clear. It's the type of the returned value.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ReferenceData.java
```diff
@@ -89,6 +89,31 @@ public default boolean containsValue(ReferenceDataId<?> id) {
...
+   * @return the reference data value
+   * @throws ReferenceDataNotFoundException if the identifier is not found
+   * @throws ClassCastException if the identifier exists, but is not of the correct type
+   */
+  @SuppressWarnings("unchecked")
+  public default <T, S extends T> S getValue(ReferenceDataId<T> id, Class<S> type) {
```
cjkent: Why is this method necessary? The ID already narrows the type of the returned value. Is it a common requirement to narrow it further? So common that it's worth having on the API rather than the caller downcasting the returned value?

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveInfoType.java
```diff
@@ -38,11 +38,11 @@
...
    *
-  * @param <T>  the type associated with the info
-   * @param name  the name of the curve
-   * @return a curve with the specified name
+   * @param <T>  the type associated with the info
+   * @param name  the name
+   * @return a type instance with the specified name
    */
   @FromString
   public static <T> CurveInfoType<T> of(String name) {
```
cjkent: I find it very surprising that this class has a type parameter that is never used anywhere. It contains no instances of that type and no `Class` object of the type. It's like a compile-time phantom which disappears and can never be used at runtime.

jodastephen: This is a Joda-Convert type, so it cannot contain `Class` without much more complexity that simply isn't justified.

The type is captured and used in the constants at the top of the file, which are used in other parts of the system to make queries more usable. The alternative is a String to String map which is less safe than this design (where only comments were changed).

cjkent: Fair enough

cjkent:
That's everything.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ReferenceData.java
jodastephen: Yes its common. Every security class needs to do this.

PR #842, opened 2016-03-16
Add basics of new security model

jodastephen:
Add identifier, info, trade and interfaces.

The trade model in Strata is split between OTCs and Security-based trades. The standard way to represent a security-based trade is using `SecurityIdTrade`. This uses an identifier to refer to the security, where each security uses an identifier to refer to its underlying securities. This requires reference data to be populated with the set of securities that are reference by the trades. This approach is particularly suitable for mark-market pricing.

A second way to represent a security-based trade is using a dedicated product type, such as `BondFutureTrade`. This approach stores the product, which contains the information necessary to price the trade against a model. With this approach, there is no need to populate reference data with securities, as all necessary data is contained in the trade.


modules/product/src/main/java/com/opengamma/strata/product/SecuritizedProduct.java
```diff
@@ -0,0 +1,44 @@
...
+ * The product details of a financial instrument that is traded as a security.
+ * <p>
+ * A securitized product contains the structure of a financial instrument that is traded as a {@link Security}.
+ * The product of a security is distinct from the security itself.
+ * A {@link Security} contains details about itself, with any underlying securities
+ * referred to by {@linkplain SecurityId identifier}.
+ * By contrast, the product contains the full model for pricing, including underlying products.
+ * <p>
+ * For example, the securitized product of a bond future option directly contains all
+ * the details of the future and the basket of bonds. Whereas, a bond future security
```
cjkent: Presumably this should say "a bond future option security"?

modules/product/src/main/java/com/opengamma/strata/product/SecuritizedProductTrade.java
```diff
@@ -0,0 +1,78 @@
...
+   * Gets the quantity that was traded.
+   * <p>
+   * This will be positive if buying and negative if selling.
+   * 
+   * @return the quantity
+   */
+  public abstract long getQuantity();
+
+  /**
+   * Gets the price that was traded.
```
cjkent: Is this the total price of the trade or the unit price (trade price / quantity)?

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * The identifier, expressed as a standard two-part identifier.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final StandardId standardId;
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains an instance from a scheme and value.
```
cjkent: It would be good to explain what a scheme is and provide examples, either here or in the class Javadoc. We're so familiar with the concept we take it for granted but I'm not sure it's completely obvious without an explanation.

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,164 @@
...
+   */
+  public static SecurityId of(StandardId standardId) {
+    return new SecurityId(standardId);
+  }
+
+  /**
+   * Parses an {@code StandardId} from a formatted scheme and value.
+   * <p>
+   * This parses the identifier from the form produced by {@code toString()}
+   * which is '{@code $scheme~$value}'.
```
cjkent: I'm not sure whether the `$` signs here make it more or less obvious what's happening. I can see what you're saying, but I don't know how open to misinterpretation it is. Maybe add an example.

modules/product/src/main/java/com/opengamma/strata/product/SecurityId.java
```diff
@@ -0,0 +1,164 @@
...
+   * @param str  the identifier to parse
+   * @return the security identifier
+   * @throws IllegalArgumentException if the identifier cannot be parsed
+   */
+  @FromString
+  public static SecurityId parse(String str) {
+    return new SecurityId(StandardId.parse(str));
+  }
+
+  // creates an identifier for a combined calendar
```
cjkent: Copy/paste error. Not a calendar.

modules/product/src/main/java/com/opengamma/strata/product/SecurityIdTrade.java
```diff
@@ -0,0 +1,551 @@
...
+ * <p>
+ * This trade represents a trade in a security, defined by a quantity and price.
+ * The security is referenced using {@link SecurityId}.
+ * The identifier may be looked up in {@link ReferenceData}.
+ * <p>
+ * The reference may be to any kind of security, including equities,
+ * bonds and exchange traded derivatives (ETD).
+ */
+@BeanDefinition(constructorScope = "package")
+public final class SecurityIdTrade
```
cjkent: I'm not sure about the name of this class, given how important it is. Is there a more industry-standard name we could use?

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfo.java
```diff
@@ -0,0 +1,414 @@
...
+   * This identifier uniquely identifies the security within the system.
+   * It is the key used to lookup the security in {@link ReferenceData}.
+   * <p>
+   * A real-world security will typically have multiple identifiers.
+   * The only restriction placed on the identifier is that it is sufficiently
+   * unique for the reference data lookup. As such, it is acceptable to use
+   * an identifier from a well-known global or vendor symbology.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final SecurityId id;
```
cjkent: Is the ID on the security too?

Also, I'm surprised this is optional. Why would you have a security with no ID?

modules/product/src/main/java/com/opengamma/strata/product/SecurityIdTrade.java
jonathansenior: `SecurityTrade`?

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfo.java
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  /**
+   * Obtains an instance from the identifier and additional information.
+   * 
+   * @param id  the security identifier
+   * @param infoMap  the map of additional information
+   * @return the security information
+   */
+  public static SecurityInfo of(SecurityId id, Map<SecurityInfoType<?>, Object> infoMap) {
```
cjkent: I'm not sure about this signature. `SecurityInfoType` doesn't have a field containing the `Class` so there is no way check the type at runtime. This map is completely type-unsafe and any problems won't be evident until the data is looked up and a `ClassCastException` is thrown.

If we can't have a `Class` field on `SecurityInfoType` then I think the least we should do is remove this method and have a builder with a method:

    Builder <T> addSecurityInfo(SecurityInfotype<T> type, T info);

That way we have some compile-time type checking of the info values.

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfoType.java
```diff
@@ -0,0 +1,59 @@
...
+package com.opengamma.strata.product;
+
+import org.joda.convert.FromString;
+
+import com.opengamma.strata.collect.type.TypedString;
+
+/**
+ * The type of additional security information.
+ * <p>
+ * This extension point allow arbitrary information to be associated with a security.
```
cjkent: s/allow/allows/

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfoType.java
```diff
@@ -0,0 +1,59 @@
...
+ * For example, it might be used to provide information about the trading platform.
+ * <p>
+ * Applications that wish to use security information should declare a static
```
cjkent: Maybe link to the `NAME` field as an example.

cjkent:
That's everything from me

modules/product/src/main/java/com/opengamma/strata/product/SecurityIdTrade.java
jodastephen: As I mentioned previously, more than one concept wants the name `SecurityTrade`. I stuck with what I felt was the most obvious and similar to today - `SecurityTrade` for the interface and `SecurityIdTrade` for the identifier-based class. It has the advantage that it is obviously identifier-based.

modules/product/src/main/java/com/opengamma/strata/product/SecurityInfo.java
jodastephen: It is not optional. There is a method to get the identifier from the security, but it delegates to the info object. This approach was closer to that used in `TradeInfo` and seemed to work well.

cjkent: You're right, it's not optional. But the class Javadoc says "All properties on this class are optional". I read that and didn't read the annotation on the field.

PR #854, opened 2016-03-17
Security trade model

jodastephen:
Refactor security trade model to better model real world.

The trade model in Strata is split between OTCs and Security-based trades. The standard way to represent a security-based trade is using SecurityIdTrade. This uses an identifier to refer to the security, where each security uses an identifier to refer to its underlying securities. This requires reference data to be populated with the set of securities that are reference by the trades. This approach is particularly suitable for mark-market pricing.

A second way to represent a security-based trade is using a dedicated product type, such as BondFutureTrade. This approach stores the product, which contains the information necessary to price the trade against a model. With this approach, there is no need to populate reference data with securities, as all necessary data is contained in the trade.


examples/src/main/resources/example-reports/security-report-template.ini
```diff
@@ -11,16 +11,16 @@ value = Trade.counterparty.value
 value = Trade.settlementDate
 
 [Maturity Date]
-value = Security.product.expiryDate
+value = Security.securityInfo.additionalInfo.ExpiryDate
```
cjkent: This doesn't look great. Is there any scope for shortening names or renaming fields? Something like

    value = Security.info.attributes.ExpiryDate

modules/product/src/main/java/com/opengamma/strata/product/GenericSecurityTrade.java
```diff
@@ -23,28 +23,17 @@
...
- * if jet fuel rises in price.
+ * This trade represents a trade in a security, defined by a quantity and price.
+ * The security is embedded directly, however the underlying product model is not available.
+ * The security may be of any kind, including equities, bonds and exchange traded derivative (ETD).
  */
-@BeanDefinition
-public final class GenericFutureTrade
-    implements SecurityTrade<GenericFuture>, ImmutableBean, Serializable {
+@BeanDefinition(constructorScope = "package")
+public final class GenericSecurityTrade
```
cjkent: What's this class for? It seems to live in the middle ground between securities, where other securities are linked by ID, and products, where other products are directly referenced.

modules/product/src/main/java/com/opengamma/strata/product/ModelledSecurity.java
```diff
@@ -0,0 +1,73 @@
...
+ * <p>
+ * Most securities have some form of underlying model that allows pricing from first principles.
+ * For example, an {@linkplain IborFutureSecurity STIR future} contains details of the
+ * underlying Ibor index, and a {@linkplain DeliverableSwapFutureSecurity DSF} contains
+ * information on the swap that will be delivered when the future expires.
+ * This underlying model is implemented as a {@code Product}.
+ * <p>
+ * Implementations of this interface must be immutable beans.
+ */
+public interface ModelledSecurity extends Security {
```
cjkent: This name is quite a mouthful. Are there any obvious alternatives? `ModelSecurity`?

modules/product/src/main/java/com/opengamma/strata/product/Security.java
```diff
@@ -1,62 +1,74 @@
...
    * <p>
-   * The standard identifier is used to identify the security.
-   * It will typically be an identifier in an external data system.
-   * <p>
-   * A security may have multiple active identifiers. Any identifier may be chosen here.
-   * Certain uses of the identifier, such as storage in a database, require that the
-   * identifier does not change over time, and this should be considered best practice.
+   * This identifier uniquely identifies the security within the system.
+   * 
+   * @return the security identifier
```
cjkent: Copy/paste error; this is the Javadoc for `getSecurityId`.

modules/product/src/main/java/com/opengamma/strata/product/Security.java
```diff
@@ -1,62 +1,74 @@
...
+   * Creates a trade based on this security.
    * <p>
-   * All securities refer to a single underlying product.
-   * The product captures the financial details of the security contract.
+   * This creates a trade of a suitable type for this security.
    * 
-   * @return the product
+   * @param tradeInfo  the trade information
+   * @param quantity  the number of contracts in the trade
+   * @param tradePrice  the price that was traded, in decimal form
```
cjkent: Isn't "in decimal form" redundant? What other form could it be given it's a `double`?

modules/product/src/main/java/com/opengamma/strata/product/SecurityTrade.java
```diff
@@ -5,97 +5,67 @@
...
+ * The security is embedded directly, however the underlying product model is not available.
+ * The security may be of any kind, including equities, bonds and exchange traded derivative (ETD).
  * <p>
  * Implementations of this interface must be immutable beans.
- * 
- * @param <P>  the type of the product
  */
-public interface SecurityTrade<P extends Product>
-    extends FinanceTrade, LinkResolvable<SecurityTrade<P>> {
+public interface SecurityTrade
```
cjkent: This feels like another oddity. My understanding of the model was that there was a clean separation between securities (linked by ID) and products (directly referenced). This has a security which is directly referenced which confuses the picture slightly. 

What's it for? Would it be easier to explain the trade / security / product model without this?

modules/product/src/main/java/com/opengamma/strata/product/SecurityTrade.java
```diff
@@ -5,97 +5,67 @@
...
-    return getSecurityLink().resolvedTarget();
-  }
+  public abstract long getQuantity();
 
   /**
-   * Gets the underlying product that was agreed when the trade occurred, throwing an exception if not resolved.
-   * <p>
-   * Returns the underlying product that captures the contracted financial details of the trade.
-   * This is obtained from {@link #getSecurityLink()}.
+   * Gets the price that was traded.
```
cjkent: This sounds odd to me. Do you trade a price? "The price at which the trade was executed" or something like that.

PR #860, opened 2016-03-21
Change DoubleArray.copyOf parameter type from List to Collection

cjkent:


modules/collect/src/main/java/com/opengamma/strata/collect/array/DoubleArray.java
```diff
@@ -263,19 +264,19 @@ public static DoubleArray ofUnsafe(double[] array) {
 
   //-----------------------------------------------------------------------
   /**
-   * Obtains an instance from a list of {@code Double}.
+   * Obtains an instance from a collection of {@code Double}.
```
jodastephen: Because order matters in an array, but not in a collection, we should document the order assunption

PR #866, opened 2016-03-21
Add firstIndexValue to inflation

jodastephen:
Allow InflationRateCalculation to store a locked index value
For swaps, this matches the FpML initialIndexLevel field
For bonds, this allows startIndexValue to be removed

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/swap/InflationRateCalculation.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/swap/InflationRateCalculation.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.OptionalDouble;
 import java.util.Set;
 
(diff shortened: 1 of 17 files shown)


modules/product/src/main/java/com/opengamma/strata/product/swap/InflationRateCalculation.java
```diff
@@ -176,8 +218,13 @@ public void collectIndices(ImmutableSet.Builder<Index> builder) {
   }
 
   // creates the rate observation
-  private RateObservation createRateObservation(SchedulePeriod period) {
+  private RateObservation createRateObservation(SchedulePeriod period, int scheduleIndex) {
+
+    // handle where index value at start date is known
     LocalDate endDate = period.getEndDate();
+    if (firstIndexValue != null && scheduleIndex == 0) {
```
yukiiwashita: size > 1 case has to be excluded, e.g., by ArgChecker, rather than creating several accrual periods with only the fist period keeping the start index value?

PR #868, opened 2016-03-22
CurveParameterMetadata Containing A Tenor But No Date Fixes #853

### LINKED ISSUE #853: CurveParameterMetadata containing a Tenor but no date
When calibrated curve data is provided it is often necessary to create metadata containing the tenor of each node but the actual node date is not important and is non-trivial to calculate.

davidcervi:


modules/market/src/main/java/com/opengamma/strata/market/curve/meta/TenorNoDateCurveNodeMetadata.java
```diff
@@ -0,0 +1,374 @@
...
+
+import com.opengamma.strata.basics.date.Tenor;
+import com.opengamma.strata.collect.ArgChecker;
+import com.opengamma.strata.market.curve.CurveParameterMetadata;
+
+/**
+ * Curve node metadata for a curve node with a specific tenor and no date.
+ */
+@BeanDefinition(builderScope = "private")
+public final class TenorNoDateCurveNodeMetadata
```
cjkent: Could we call this `TenorCurveNodeMetadata` and rename the existing one `TenorDateCurveNodeMetadata`?

davidcervi: yep cool makes sense

PR #875, opened 2016-03-23
Add position model

jodastephen:
Add interfaces and classes for positions in securities.

- `Position` - interface for positions
- `SecurityPosition` - implementation of position based on an identifier
- `GenericSecurityPosition` - implementation of position based on `GenericSecurity`
- `SecurityQuantity` - interface abstracting trades and positions


modules/product/src/main/java/com/opengamma/strata/product/Position.java
```diff
@@ -0,0 +1,53 @@
...
+   * Gets the net quantity of the security.
+   * <p>
+   * This returns the <i>net</i> quantity of the underlying security.
+   * The result is positive if the net position is <i>long</i> and negative
+   * if the net position is <i>short</i>.
+   * 
+   * @return the net quantity of the underlying security
+   */
+  @Override
+  public abstract long getQuantity();
```
jonathansenior: Not sure we can simplify this down to an integer quantity, as much as this is accurate for listed securities. Firms sometimes allocate a single real position across several portfolios (e.g. their clients) in some proportion, and will want to price these. This will usually result in a fractional position.

joburnham: Definitely for ETD it should be integer but it may be safer to go with a decimal - especially if Jon has experience of people creating fractional positions

jodastephen: Will change to `double` in a follow up PR.

modules/product/src/main/java/com/opengamma/strata/product/GenericSecurityPosition.java
```diff
@@ -0,0 +1,663 @@
...
+
+/**
+ * A position in a security, where the security is embedded ready for mark-to-market pricing.
+ * <p>
+ * This represents a position in a security, defined by long and short quantity.
+ * The security is embedded directly, however the underlying product model is not available.
+ * The security may be of any kind, including equities, bonds and exchange traded derivatives (ETD).
+ * <p>
+ * The net quantity of the position is stored using two fields - {@code longQuantity} and {@code shortQuantity}.
+ * Each field must be positive, however they may both be non-zero.
```
cjkent: To me "positive" implies non-zero. "These fields must not be negative" sounds better.

modules/product/src/main/java/com/opengamma/strata/product/GenericSecurityPosition.java
```diff
@@ -0,0 +1,663 @@
...
+   */
+  @Override
+  public SecurityId getSecurityId() {
+    return security.getSecurityId();
+  }
+
+  /**
+   * Gets the currency of the trade.
+   * <p>
+   * This is typically the same as the currency of the product.
```
cjkent: "typically"? Is it ever a different currency?

Also it should probably say "security" instead of "product".

cjkent:
Merging, there are only a couple of very minor things and they can be addressed in the PR changing the type of the quantity.

PR #877, opened 2016-03-29
Merge ImmutableRatesProvider

marc-opengamma:
Create new methods to merge several ImmutableRatesProvider.
This is useful when curves in several currencies are calibrated separately and then grouped together for valuation of multi-currency trades or portfolios.


modules/pricer/src/main/java/com/opengamma/strata/pricer/rate/ImmutableRatesProvider.java
```diff
@@ -111,6 +113,27 @@ private static void applyDefaults(Builder builder) {
...
+   * <p>
+   * If the two providers have curves or time series for the same currency or index, 
+   * an {@link IllegalAccessException} is thrown. 
+   * The FxRateProviders is not populated with the given provider; no attempt is done on merging the embedded FX providers.
+   * 
+   * @param fx  the FX provider for the resulting rate provider
+   * @param providers  the rates providers to be merged
+   * @return the combined rates provider
+   */
+  public static ImmutableRatesProvider combined(FxRateProvider fx, ImmutableRatesProvider... providers) {
```
cjkent: We've used the method name `combinedWith` in other places.

cjkent: Ignore me, I've just realised it's a static method.

PR #884, opened 2016-03-30
Remove ReferenceData.identifiers and make combinedWith a default method

### LINKED ISSUE #872: Remove ReferenceData.identifiers()
This method requires enumerating the entire store of reference data which is not desirable.

It is only used when combining multiple sets of reference data which can be implemented using delegation.

cjkent:
Removes `ReferenceData.identifiers` and changes `ReferenceData.combinedWith` to a default method which combines by delegating to two underlying `ReferenceData` instances.

Fixes #872


jodastephen:
A chaining approach may be more effective than this in performance terms as it could have a dedicated `getValue()` override. Merging as this will work fine and performance impact is unproved.

PR #885, opened 2016-03-30
Add support for positions and target to reporting

jodastephen:
Add positions and targets to reporting
Enhance securities in reporting


modules/report/src/main/java/com/opengamma/strata/report/framework/expression/SecurityTokenEvaluator.java
```diff
@@ -0,0 +1,83 @@
...
+import org.joda.beans.JodaBeanUtils;
+import org.joda.beans.MetaBean;
+
+import com.google.common.collect.Sets;
+import com.opengamma.strata.product.Security;
+import com.opengamma.strata.product.SecurityInfo;
+import com.opengamma.strata.product.SecurityPriceInfo;
+
+/**
+ * Evaluates a token against a trade to produce another object.
```
cjkent: s/trade/security/

PR #897, opened 2016-04-11
Created curve sensitivity re-bucketing util.

marc-opengamma:
Used to re-bucket sensitivities to different dates.
Often used to obtained a simplified or standardized view of curve risk.


marc-opengamma:
Utility will be expanded and a new PR open.

PR #900, opened 2016-04-12
Enhance calculation functions

jodastephen:
Add `targetType()` to `CalculationFunction`
Add `CalculationFunctions` to hold a map of functions
Add to `StandardComponents`

This is the first part of simplifying the pricing rules.


modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,221 @@
...
+   * <p>
+   * The list will be converted to a {@code Map} using {@link CalculationFunction#targetType()}.
+   * Each function must refer to a different target type.
+   * 
+   * @param functions  the functions
+   * @return the calculation functions
+   */
+  @SafeVarargs
+  public static CalculationFunctions of(CalculationFunction<?>... functions) {
+    return new CalculationFunctions(Stream.of(functions).collect(toImmutableMap(fn -> fn.targetType())));
```
cjkent: Just to note, you could use `MapStream` here:

    MapStream.of(functions, fn -> fn.targetType()).toMap();

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,221 @@
...
+  public static CalculationFunctions of(Map<Class<?>, ? extends CalculationFunction<?>> functions) {
+    return new CalculationFunctions(ImmutableMap.copyOf(functions));
+  }
+
+  @ImmutableValidator
+  private void validate() {
+    for (Entry<Class<?>, CalculationFunction<?>> entry : functions.entrySet()) {
+      ArgChecker.isTrue(
+          entry.getValue().targetType().isAssignableFrom(entry.getKey()),
+          "Invalid map, key and function mismatch: {}", entry.getKey());
```
cjkent: It would be good to include the function or it's type in the error message.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,221 @@
...
+
+  //-------------------------------------------------------------------------
+  /**
+   * Gets the function that handles the specified target.
+   * 
+   * @param target  the calculation target, such as a trade
+   * @param defaultFunction  the default function to use if not found
+   * @return the function
+   */
+  public CalculationFunction<?> getFunction(CalculationTarget target, CalculationFunction<?> defaultFunction) {
```
cjkent: This is a surprising API. Why would the caller have a default function to hand? Wouldn't it be better to have a default function as a field in `CalculationFunctions`?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CalculationFunction.java
```diff
@@ -44,6 +46,15 @@
 public interface CalculationFunction<T extends CalculationTarget> {
 
   /**
+   * Gets the target type that this function applies to.
+   * <p>
+   * The target type will be matched exactly, thus it should be a class, not an interface.
```
cjkent: Is this a wise restriction? Dealing with subtyping might be messy but Sesame had this same restriction and it was a source of problems.

It's probably not something to tackle now but we should think about it and probably leave the door open to it just in case.

modules/function/src/main/java/com/opengamma/strata/function/StandardComponents.java
```diff
@@ -142,4 +176,33 @@ public static PricingRules pricingRules() {
     return StandardPricingRules.standard();
   }
 
+  /**
+   * Returns the standard calculation functions.
+   * <p>
+   * These rules define how to calculate the standard measures for the standard asset classes.
```
cjkent: "rules" looks a bit out of place given that we're planning to get rid of pricing rules.

cjkent:
That's everything

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * The calculation functions.
+ * <p>
+ * This provides the complete set of functions that will be used in a calculation.
+ * Each {@link CalculationFunction} handles a specific type of {@link CalculationTarget},
+ * thus the functions are keyed in a {@code Map} by the target type {@code Class}.
+ */
+@BeanDefinition(style = "light")
+public final class CalculationFunctions implements ImmutableBean, Serializable {
```
cjkent: The `getFunction` and `findFunction` methods have a parameter for the target, not the target type. I think that's a very good thing as it leaves the door open to choosing the function based on more complicated criteria. For example it would be possible to have different functions for different types of swap where the target type is always `Swap` but the legs have different types. It would also allow target attributes to be used when choosing the function.

The only problem is that this class is a concrete final type and the methods for finding functions are not defined on an interface. Which means this is the only possible implementation.

Could we convert `CalculationFunctions` to an interface containing `getFunction` and `findFunction` and rename this class? Maybe `TargetTypeCalculationFunctions`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
cjkent: Could this signature be:

    public <T extends CalculationTarget> CalculationFunction<? super T> getFunction(T target)

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,221 @@
...
+    return function != null ? function : defaultFunction;
+  }
+
+  /**
+   * Finds the function that handles the specified target.
+   * 
+   * @param target  the calculation target, such as a trade
+   * @return the function
+   */
+  public Optional<CalculationFunction<?>> findFunction(CalculationTarget target) {
```
cjkent: Could this signature be:

    public <T extends CalculationTarget> Optional<CalculationFunction<? super T>> findFunction(T target)

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
jodastephen: As it happens, the generics don't match if I do that. So in this case, I'll leave as is

cjkent: Do we need to tweak the type signature on `MapStream` to make it work?

jodastephen:
Updated

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunctions.java
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Gets the function that handles the specified target.
+   * <p>
+   * If no function is found, a suitable default that can perform no calculations is provided.
+   * 
+   * @param <T>  the target type
+   * @param target  the calculation target, such as a trade
+   * @return the function
+   */
+  public default <T extends CalculationTarget> CalculationFunction<? super T> getFunction(T target) {
```
cjkent: This method doesn't have a parameter for the measure. Which means a set of `CalculationFunctions` must contain one function which can calculate all measures. Is this intentional? How are we planning to allow measures to be added? Container functions which compose multiple underlying functions?

jodastephen: Yes, that was the intention. A single function determined for all measures. (This is also why I had a class to match only by target type, but adding the interface allowed a small degree of extra flexibility, so I went with your suggestion).

PR #903, opened 2016-04-12
Add Strata AsciiTable

jodastephen:
Removes dependency

### DIFF
--- a/modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
+++ b/modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
@@ -1,0 +1,105 @@
+/**
+ * Copyright (C) 2016 - present by OpenGamma Inc. and the OpenGamma group of companies
+ * 
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.report.framework.format;
+
+import java.util.List;
+
+import com.google.common.base.Strings;
+
+/**
+ * An ASCII table generator.
+ * <p>
+ * Provides the ability to generate an ASCII table, typically used on the command line.
+ */
+final class AsciiTable {
+
+  //-------------------------------------------------------------------------
+  /**
+   * Restricted constructor.
+   */
+  private AsciiTable() {
+  }
+
+  /**
+   * Generates the ASCII table.
+   * 
+   * @param alignments  the table alignments
+   * @param headers  the table headers
+   * @param cells  the table cells
+   * @return the table
+   */
+  static String generate(
+      List<AsciiTableAlignment> alignments,
+      List<String> headers,
+      List<? extends List<String>> cells) {
+
+    int colCount = alignments.size();
+
+    // find max length of each column
+    int[] colLengths = new int[colCount];
+    for (int colIdx = 0; colIdx < colCount; colIdx++) {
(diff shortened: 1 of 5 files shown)


modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
```diff
@@ -0,0 +1,105 @@
...
+   * 
+   * @param alignments  the table alignments
+   * @param headers  the table headers
+   * @param cells  the table cells
+   * @return the table
+   */
+  static String generate(
+      List<AsciiTableAlignment> alignments,
+      List<String> headers,
+      List<? extends List<String>> cells) {
```
jonathansenior: Should initially ensure that the inputs are all consistent (i.e. same number of columns throughout) to avoid an `IndexOutOfBoundsException` with no helpful message.

modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
```diff
@@ -0,0 +1,105 @@
...
+      colLengths[colIdx] = headers.get(colIdx).length();
+    }
+    for (List<String> row : cells) {
+      for (int colIdx = 0; colIdx < colCount; colIdx++) {
+        colLengths[colIdx] = Math.max(colLengths[colIdx], row.get(colIdx).length());
+      }
+    }
+
+    // write table
+    StringBuilder buf = new StringBuilder(2048);
```
jonathansenior: Wondering if we could be a bit more intelligent about the capacity here given that we've just worked out the width of each column, and we know the number of rows.

modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
```diff
@@ -0,0 +1,105 @@
...
+
+  // write a separator line
+  private static void writeSeparatorLine(StringBuilder buf, int[] colLengths) {
+    for (int colIdx = 0; colIdx < colLengths.length; colIdx++) {
+      buf.append('+');
+      for (int i = 0; i < colLengths[colIdx] + 2; i++) {
+        buf.append('-');
+      }
+    }
+    buf.append("+\n");
```
jonathansenior: Should probably use `System.lineSeparator()` for the benefit of non-Unix-like consoles.

jonathansenior:
That's it. Very neat and simple - good to remove the dependency.

modules/report/src/main/java/com/opengamma/strata/report/framework/format/AsciiTable.java
```diff
@@ -0,0 +1,105 @@
...
+      List<AsciiTableAlignment> alignments,
+      List<String> headers) {
+
+    for (int colIdx = 0; colIdx < colLengths.length; colIdx++) {
+      buf.append('|')
+          .append(' ')
+          .append(formatData(buf, colLengths[colIdx], alignments.get(colIdx), headers.get(colIdx)))
+          .append(' ');
+    }
+    buf.append("|\n");
```
jonathansenior: `System.lineSeparator()`?

jodastephen:
Updated

PR #909, opened 2016-04-20
Simplify calculation rules

jodastephen:
Replace `PricingRules` by `CalculationFunctions`
Introduce general `CalculationParameters` concept
Change `ReportingCurrency` and `MarketDataRules` to be parameters

This is the main part of the change. Further commits will expose parameters to the requirements and calculate methods of functions, and adjust the results objects.


modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRules.java
```diff
@@ -22,88 +23,118 @@
...
-   * This controls which subset of the complete set of market data should be used for a calculation.
-   * This defaults to an empty instance if using the builder.
-   */
-  @PropertyDefinition(validate = "notNull")
-  private final MarketDataRules marketDataRules;
-  /**
-   * The reporting currency.
+   * Parameters are used to parameterize the {@link Measure} to be calculated.
+   * Each {@link Column} specifies the measure and column-specific parameters.
+   * These parameters are are the defaults that apply to all columns.
```
cjkent: "are are"

cjkent: When I read "These parameters are the defaults..." I thought it was referring to the column-specific parameters from the previous sentence and got confused. Maybe it would be clearer to say "The parameters in the calculation rules are the defaults that..."

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameter.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Gets the type that the parameter will be queried by.
+   * <p>
+   * Parameters can be queried using {@link CalculationParameters#findParameter(Class)}.
+   * This type is the key that callers must use in that method.
+   * 
+   * @return the type of the parameter implementation
+   */
+  public abstract Class<?> queryType();
```
cjkent: I find this slightly puzzling. What is the query type? Is it expected to be the type implementing `CalculationParameter`? Can we do better with the type signature so that's clear. When I look at this interface in isolation the query type looks like a type that's floating around unconnected to anything.

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRules.java
```diff
@@ -22,88 +23,118 @@
...
+   * If a specific type of parameter is defined in both the rules and column,
+   * then the column parameter takes precedence.
    * <p>
-   * This controls the currency that the calculated result will be converted to.
-   * This defaults to 'Natural' if using the builder.
+   * There are many possible parameter implementations.
+   * Two common ones are {@link ReportingCurrency} and {@link MarketDataRules}.
```
cjkent: I'm uneasy about `MarketDataRules` being bundled in with all the other parameters given how fundamental they are. They are fundamental to the operation of the calculation runner but making them just another parameter makes them look optional and doesn't make it clear that they're more important.

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
```diff
@@ -27,100 +26,88 @@
...
-   * The final set of rules for a column consist of the default rules merged with
-   * the column-specific rules, with the column-specific rules taking precedence.
-   * In most cases, there is no need to specify column-specific rules.
-   */
-  @PropertyDefinition(validate = "notNull")
-  private final MarketDataRules marketDataRules;
-  /**
-   * The reporting currency that applies to this column, overriding the default reporting currency.
+   * The parameters from {@link CalculationRules} and {@code Column} are combined.
+   * If a parameter with the same name is specified here and in {@link CalculationRules},
```
cjkent: Parameters don't have names.

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
```diff
@@ -27,100 +26,88 @@
...
+ * <p>
+ * {@link CalculationRunner} provides the ability to calculate a grid of results
+ * for a given set targets and columns. This class is used to define the columns.
+ * <p>
+ * A column is defined in terms of a unique name, measure to be calculated and
+ * a set of parameters that control the calculation. The functions to invoke
+ * and the default set of parameters are defined on {@link CalculationRules}.
  */
 @BeanDefinition
 public final class Column implements ImmutableBean {
```
cjkent: I was expecting `Column` to contain a set of `CalculationFunctions` to allow different functions to be used in different columns.

I worry we're throwing away a lot of flexibility in the quest for simplicity, possibly too much. If we want to reconfigure anything it feels like we're going to have to write a custom function with the behaviour baked in.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameter.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.opengamma.strata.basics.CalculationTarget;
+import com.opengamma.strata.calc.config.Measure;
+
+/**
+ * The base interface for calculation parameters.
+ * <p>
+ * Parameters are used to control the calculation.
+ */
+public interface CalculationParameter {
```
cjkent: Given how important these are it would be good to go into more detail and include examples of how they are expected to be used.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * The calculation parameters.
+ * <p>
+ * This provides a set of parameters that will be used in a calculation.
+ * Each parameter defines a {@linkplain CalculationParameter#queryType() query type},
+ * thus the functions are keyed in a {@code Map} by the query type {@code Class}.
+ */
+@BeanDefinition(style = "light")
+public final class CalculationParameters implements ImmutableBean, Serializable {
```
cjkent: The Javadocs don't give enough of a picture of what `CalculationParameters` are actually for.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -0,0 +1,244 @@
...
+   * Each parameter must refer to a different query type.
+   * 
+   * @param parameters  the parameters
+   * @return the calculation parameters
+   */
+  public static CalculationParameters of(CalculationParameter... parameters) {
+    if (parameters.length == 0) {
+      return EMPTY;
+    }
+    return new CalculationParameters(Stream.of(parameters).collect(toImmutableMap(p -> p.queryType())));
```
cjkent: Using `MapStream` is a bit neater:

    return new CalculationParameters(MapStream.of(parameters, p -> p.queryType()).toMap();

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -0,0 +1,244 @@
...
+   * Each parameter must refer to a different query type.
+   * 
+   * @param parameters  the parameters
+   * @return the calculation parameters
+   */
+  public static CalculationParameters of(List<? extends CalculationParameter> parameters) {
+    if (parameters.isEmpty()) {
+      return EMPTY;
+    }
+    return new CalculationParameters(parameters.stream().collect(toImmutableMap(p -> p.queryType())));
```
cjkent: `MapStream`

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * Combines this set of parameters with the specified set.
+   * <p>
+   * This set of parameters takes priority.
+   * 
+   * @param other  the other parameters
+   * @return the combined calculation parameters
+   */
+  public CalculationParameters combinedWith(CalculationParameters other) {
+    Map<Class<?>, CalculationParameter> map = new HashMap<>(other.getParameters());
```
cjkent: Return `other` if these parameters are empty? That will happen a lot because this method is used to combine a column's parameters with the default parameters.

modules/calc/src/main/java/com/opengamma/strata/calc/Column.java
jodastephen: The design is essentially to use parameters to tweak the behaviour of standard functions, rather than to allow the function to be changed. It may involve more function writing, or it may not, we don't really know yet. What we can say is that function overrides could be added later.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationResult.java
```diff
@@ -31,8 +23,8 @@
  * This stores the calculated result for a single cell in the output grid.
  * A set of related results for a single target can be stored in a {@link CalculationResults} instance.
  */
-@BeanDefinition(builderScope = "private")
-public final class CalculationResult implements ImmutableBean, Serializable {
```
cjkent: Why remove `Serializable`? It seems reasonable for result types to be serializable.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
jodastephen: The generics do not bind with `MapStream`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -8,47 +8,61 @@
...
+ * Each task can be executed to produce the result. Applications will typically
+ * use {@link CalculationRunner} or {@link CalculationTaskRunner} to execute the tasks.
  */
-public final class CalculationTasks {
+@BeanDefinition(style = "light")
+public final class CalculationTasks implements ImmutableBean {
 
   /**
-   * The targets.
+   * The targets that will be calculated.
```
cjkent: This doesn't make sense. "The targets for which values/results/measures will be calculated".

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -8,47 +8,61 @@
...
+   * <p>
+   * The result of the calculations will be a grid where each row is taken from this list.
```
cjkent: This doesn't make much sense either. The row isn't taken from this list. Each row is associated with a target from this list.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -8,47 +8,61 @@
...
+  @PropertyDefinition(validate = "notEmpty")
   private final List<Column> columns;
   /**
-   * The calculation tasks.
+   * The tasks that perform the individual calculations.
    * <p>
-   * The tasks in the list are arranged in row order.
-   * For example, if a grid has 5 columns, the calculations 0-4 in the list are the first row,
-   * calculations 5-9 are the second row and so on.
+   * The results can be visualized as a grid of columns with a row for each target.
```
cjkent: It's a grid of cells with a column for each measure and a row for each target.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -8,47 +8,61 @@
...
+   * There is one calculation for each cell in the grid. The calculations in the list are arranged in row order.
```
cjkent: This comment doesn't really apply to this list any more as a task can span multiple columns.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -57,128 +71,74 @@
...
-  private CalculationTasks(List<CalculationTask> calculationTasks, List<Column> columns) {
+  private CalculationTasks(List<CalculationTask> tasks, List<Column> columns) {
     this.columns = ImmutableList.copyOf(columns);
-    this.calculationTasks = ImmutableList.copyOf(calculationTasks);
+    this.tasks = ImmutableList.copyOf(tasks);
 
-    // Validate the number of tasks and number of columns tally
-    int taskCount = calculationTasks.size();
+    // validate the number of tasks and number of columns tally
+    long taskCount = tasks.stream()
```
cjkent: Probably worth renaming `taskCount` to `cellCount` now there isn't a 1-1 relationship between tasks and cells.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -192,72 +152,152 @@ private CalculationTasks(List<CalculationTask> calculationTasks, List<Column> co
...
-        .mapToObj(i -> calculationTasks.get(i * columnCount).getTarget())
-        .collect(toImmutableList());
+    int targetCount = (int) taskCount / columnCount;
+    CalculationTarget[] targets = new CalculationTarget[targetCount];
+    for (CalculationTask task : tasks) {
+      int rowIndex = task.getRowIndex();
+      if (targets[rowIndex] == null) {
+        targets[rowIndex] = task.getTarget();
+      } else if (targets[rowIndex] != task.getTarget()) {
+        throw new IllegalArgumentException("Tasks define two different targets for the same row");
```
cjkent: Probably worth including both targets in the exception message for easier debugging.

modules/calc/src/main/java/com/opengamma/strata/calc/CalculationRules.java
jodastephen: Suggest this is discussed as part of the wider market data rules review.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/MissingConfigCalculationFunction.java
```diff
@@ -23,13 +24,18 @@
...
-  // this must be public so that the constructor can be invoked
+  // TODO package scoped
 
   /**
    * Shared instance.
    */
   public static final CalculationFunction<CalculationTarget> INSTANCE = new MissingConfigCalculationFunction();
 
+  // restricted constructor
+  public MissingConfigCalculationFunction() {
```
cjkent: This constructor is redundant and the comment is wrong.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/Results.java
```diff
@@ -27,33 +30,42 @@
...
  * Results of performing calculations for a set of targets over a set of scenarios.
  */
-@BeanDefinition
+@BeanDefinition(builderScope = "private")
 public final class Results implements ImmutableBean {
 
-  /** The number of rows in the results. */
+  /**
+   * The number of rows in the results.
+   * This is used to interpret the list of
```
cjkent: The second half of the comment seems to have gone missing.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/Results.java
```diff
@@ -119,6 +131,23 @@ private String invalidColumnIndexMessage(int columnIndex) {
...
+  /**
+   * Processes a successful result by applying a function that alters each value.
+   * <p>
+   * This operation allows post-processing of result values.
+   * The specified function represents a conversion to be performed on each value.
+   *
+   * @param function  the function to transform the value with
+   * @return the new results
+   */
+  public Results map(Function<Result<?>, Result<?>> function) {
```
cjkent: What's the use-case for this?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+ * <p>
+ * The {@link CalculationRunner} produces a grid of results, with a row for each target
+ * and a column for each measure. The targets and columns that define the grid of results
+ * are passed in using an instance of {@code CalculationTasks}.
+ * This builder is used to produce the tasks for a single target.
+ * <p>
+ * The result of the builder consists of a number of tasks, each containing one or more task cells.
+ * Each task cell corresponds to a single column in the row.
+ */
+public final class TaskBuilder {
```
cjkent: This is an odd class. It doesn't follow the normal pattern for a builder. It only has one mutation method which is `executeSeparately`, otherwise all the data is passed into the constructor. And it has getters which is a departure for a builder.

Separate execution seems like it would be a natural property of the column as it's the measure which dictates how long it takes to calculate. If `executeSeparately` were remove then this whole class could be an immutable bean.

Are the getters necessary? Could it be reduced to a static method returning a list of tasks?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+      this.mappings = mappings;
+      this.parameters = parameters;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (obj instanceof GroupingKey) {
+        GroupingKey other = (GroupingKey) obj;
+        // mappings matched by ==, parameters by equals()
+        return mappings == other.mappings &&
```
cjkent: Why are mappings matched by reference equality? That seems like a potential cause of subtle bugs.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskColumn.java
```diff
@@ -0,0 +1,228 @@
...
+   * <p>
+   * This defines the calculation being performed, such as 'PresentValue' or 'ParRate'.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final Measure measure;
+  /**
+   * The calculation parameters, used to control the how the calculation is performed.
+   * <p>
+   * The parameters specified here must be specific to the target and measure.
+   * The reporting currency parameter will not be present.
```
cjkent: This comment makes me question whether reporting currency should be a parameter at all. It is obviously already receiving special treatment. It would seem clearer if reporting currency were fully separated and parameters only applied to targets and measures.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskColumn.java
```diff
@@ -0,0 +1,228 @@
...
+    // extract the known parameters
+    MarketDataRules mdRules = column.getMarketDataRules();
+    MarketDataMappings mappings = mdRules.mappings(target).orElse(NoMatchingRuleMappings.INSTANCE);
+    ReportingCurrency reportingCurrency = parameters.findParameter(ReportingCurrency.class)
+        .orElse(ReportingCurrency.NATURAL);
+
+    // filter the parameters to remove extracted items
+    CalculationParameters filteredParams = parameters
+        .filter(target, measure)
+        .without(MarketDataRules.class)
```
cjkent: Presumably this line is redundant now?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CalculationFunction.java
```diff
@@ -83,6 +86,20 @@
...
+   * Manages the calculation tasks.
+   * <p>
+   * This allows a function to control the tasks that are executed to calculate the results.
+   * <p>
+   * The tasks are created using a dedicated builder that provides access to the target and columns.
+   * Customization of the tasks can be achieved by calling methods on the builder.
+   * 
+   * @param builder  the calculation task builder
+   */
+  public default void manageTasks(TaskBuilder builder) {
```
cjkent: I'm not keen on this for a number of reasons.

* `manage` is a non-name that doesn't communicate anything about what's supposed to be happening
* The signature means there must be mutation going on which is something we've tried to avoid in Strata

Presumably this only exists to allow the function to call `builder.executeSeparately`? Isn't there a better way to achieve the same thing?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameter.java
```diff
@@ -23,7 +35,7 @@
    * 
    * @return the type of the parameter implementation
    */
-  public abstract Class<?> queryType();
+  public abstract Class<? extends CalculationParameter> queryType();
```
cjkent: Should the return type be the implementing type? If that's the case we should have a type parameter to make that happen.

With this signature it's perfectly possible for a parameter implementation to return a totally unrelated parameter type from `queryType`.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -47,7 +55,7 @@
    * The parameters, keyed by query type.
    */
   @PropertyDefinition(validate = "notNull")
-  private final ImmutableMap<Class<?>, CalculationParameter> parameters;
+  private final ImmutableMap<Class<? extends CalculationParameter>, CalculationParameter> parameters;
```
cjkent: Shouldn't both type parameters have wildcards?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/MissingConfigCalculationFunction.java
jodastephen: This is caused by splitting this commit from the next one. When the next commit is merged, this issue will be solved.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/Results.java
jodastephen: public Results calculateSingleScenario(
        CalculationTasks tasks,
        CalculationEnvironment marketData,
        ReferenceData refData) {

      // perform the calculations
      Results results = calculateMultipleScenarios(tasks, marketData, refData);

      // unwrap the results
      // since there is only one scenario it is not desirable to return scenario result containers
      return results.map(r -> unwrapScenarioResult(r));
    }

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskBuilder.java
jodastephen: This exists to allow the functions to intercede in building the tasks. Right now, it doesn't do much, but my intention is for it to grow to handle things like pre-calculating securities shared between bonds, futures and options etc.

Another use would be to allow a different function to be picked for some columns, based on the parameters.

Another use might be to setup certain cells as being unable to calculate for the target type.

(In general, this commit makes the function the hub of all pluggable logic. Part of sensibly allocating effort is controlling the creation of the tasks. Allowing the function to control them outside the control of the framework was fraught with difficulties, especially in ensuring that a result is generated for every cell. As such, I adopted this design - a mutable builder which functions can call.)

modules/calc/src/main/java/com/opengamma/strata/calc/runner/TaskBuilder.java
jodastephen: Mappings are not guaranteed to have an equals(), and if they do it might be complex. But I'll change it as it should cause no harm.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/function/CalculationFunction.java
jodastephen: Was called `createTasks()` and returned the tasks. But this leads to potentially unsafe states where not all cells have an associated result calculation.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
jodastephen: While accepting a wildcard on input is useful, there is no value in declaring the local variable that way.

PR #920, opened 2016-04-26
Enhance results model

jodastephen:
Rename `Results` field from "items" to "cells".
Add column header to help applications interpret results.


modules/calc/src/main/java/com/opengamma/strata/calc/ColumnHeader.java
```diff
@@ -0,0 +1,412 @@
...
+   * The measure that was calculated.
+   * <p>
+   * This defines the calculation that was performed, such as 'PresentValue' or 'ParRate'.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final Measure measure;
+  /**
+   * The currency of the result.
+   * <p>
+   * If the measure is {@linkplain Measure#isCurrencyConvertible() currency-convertible}
```
cjkent: I think "can be automatically converted to a different currency" would be more understandable than "currency-convertible".

modules/calc/src/main/java/com/opengamma/strata/calc/Results.java
```diff
@@ -58,45 +55,72 @@
...
               this.rowCount,
               this.columnCount));
     }
   }
 
+  //-------------------------------------------------------------------------
+  /**
+   * Gets the number of rows in the results.
+   * <p>
+   * The number of rows equals the number of target input to the calculation.
```
cjkent: s/target/targets/

PR #923, opened 2016-04-27
Usability enhancements for CalculationParameters and CalculationFunctions

cjkent:
Adds the following methods:

* `CalculationParameters.getParameter()` - Same as `findParameter` but throws an exception if no parameter if available. Follows the pattern used elsewhere in Strata.

* `CalculationFunctions.composedWith()` - Allows multiple sets of functions to be combined.


modules/calc/src/main/java/com/opengamma/strata/calc/runner/CompositeCalculationFunctions.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+
+import com.opengamma.strata.basics.CalculationTarget;
+import com.opengamma.strata.calc.runner.function.CalculationFunction;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * A set of calculation functions which combines the functions in two other sets of functions.
+ */
+class CompositeCalculationFunctions implements CalculationFunctions {
```
jodastephen: DefaultCalculationFunctions is a Joda-Bean "light" so it gets equals and hashCode and toString. Think this should be too.

modules/calc/src/test/java/com/opengamma/strata/calc/runner/CompositeCalculationFunctionsTest.java
```diff
@@ -0,0 +1,125 @@
...
+    CalculationFunctions fns2 = CalculationFunctions.of(fn1b, fn2);
+    CalculationFunctions composed = fns1.composedWith(fns2);
+
+    assertThat(composed.getFunction(new Target1())).isEqualTo(fn1a);
+    assertThat(composed.getFunction(new Target2())).isEqualTo(fn2);
+  }
+
+  //--------------------------------------------------------------------------------------------------
+
+    private static final class Fn1 implements CalculationFunction<Target1> {
```
jodastephen: To far indented

PR #947, opened 2016-05-09
Created new methods with adjusted forward rate for CMS periods.

marc-opengamma:
Created new methods for CMS coupon with adjusted forward rate and adjustment. The adjusted forward is the number to be 'plugged' in the standard formula pv=fwd*df*notional to obtain the correct present value.
Modified the CMS pricer to have a fixed tolerance in the numerical integration. This is to make the pricing linear in the notional, i.e. multiplying the notional by 10  should multiply the PV by 10 also. Due to the variable tolerance, it was not the case before the modification.
Added documentation to the 'RungeKuttaIntegrator1D' to clarify the meaning of the absolute and relative tolerance in the code.


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/cms/SabrExtrapolationReplicationCmsPeriodPricer.java
```diff
@@ -219,6 +220,57 @@ public CurrencyAmount presentValue(
...
+    ArgChecker.isTrue(cmsPeriod.getCmsPeriodType().equals(CmsPeriodType.COUPON),
+        "Adjusted forward rate available only for CMS coupons");
+    Currency ccy = cmsPeriod.getCurrency();
+    double dfPayment = provider.discountFactor(ccy, cmsPeriod.getPaymentDate());
+    double pv = presentValue(cmsPeriod, provider, swaptionVolatilities).getAmount();
+    return pv / (cmsPeriod.getNotional() * cmsPeriod.getYearFraction() * dfPayment);
+  }
+
+  /**
+   * Computes the adjusted forward rate for a CMS coupon.
```
yukiiwashita: The javadoc must be changed so that adjustment is explained.

PR #952, opened 2016-05-10
Enhance surface metadata

jodastephen:
Enhance surface metadata to match curve metadata design


modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadata.java
```diff
@@ -122,10 +133,43 @@ private static void applyDefaults(Builder builder) {
...
+  }
+
+  @Override
+  public DefaultSurfaceMetadata withParameterMetadata(List<? extends SurfaceParameterMetadata> parameterMetadata) {
     return toBuilder().parameterMetadata(parameterMetadata).build();
   }
 
+  //-------------------------------------------------------------------------
+  /**
+   * Returns a builder that allows this bean to be mutated.
```
cjkent: This comment isn't true - it creates a mutable builder containing the data from this object. This object can't be mutated.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * The surface name.
+   */
+  private SurfaceName surfaceName;
+  /**
+   * The x-value type, providing meaning to the x-values of the surface.
+   * <p>
+   * This type provides meaning to the x-values.
+   * <p>
+   * If using the builder, this defaults to {@link ValueType#UNKNOWN}.
```
cjkent: I'm not sure what "If using the builder" means given that this is the builder.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   * 
+   * @param xValueType  the x-value type
+   * @return this, for chaining
+   */
+  public DefaultSurfaceMetadataBuilder xValueType(ValueType xValueType) {
+    ArgChecker.notNull(xValueType, "xValueType");
+    this.xValueType = xValueType;
```
cjkent: These 2 lines could be combined as

    this.xValueType = ArgChecker.notNull(xValueType, "xValueType");

That's the pattern we've used elsewhere.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   * Adds additional information.
+   * <p>
+   * This is stored in the additional information map using {@code Map.putAll} semantics
+   * 
+   * @param info  the information to add
+   * @return this, for chaining
+   */
+  public DefaultSurfaceMetadataBuilder addInfo(Map<SurfaceInfoType<?>, Object> info) {
+    ArgChecker.notNull(info, "infoMap");
+    this.info.putAll(info);
```
cjkent: This isn't type safe. If one of the values doesn't match the expected type of the surface info there will be a `ClassCastException` when the value is looked up. An exception should be thrown here if possible.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+        this.parameterMetadata = new ArrayList<>();
+      }
+      this.parameterMetadata.clear();
+      this.parameterMetadata.addAll(parameterMetadata);
+    }
+    return this;
+  }
+
+  /**
+   * Sets the {@code parameterMetadata} property in the builder from an array of objects.
```
cjkent: "from an array of objects" is true but that's just an implementation detail of varags. It seems an odd way to describe what's happening here.

modules/market/src/main/java/com/opengamma/strata/market/surface/SurfaceMetadata.java
```diff
@@ -60,25 +61,58 @@
...
+   * If this metadata instance already contains additional info, the two maps will
+   * be merged, with the specified map taking priority, as per {@link Map#putAll(Map)}.
+   * <p>
+   * The map must contain no nulls. The value of each entry must match the parameterized
+   * type of the associated {@code SurfaceInfoType} key.
+   * 
+   * @param additionalInfo  the additional information to add
+   * @return the new curve metadata
+   */
+  public abstract SurfaceMetadata withInfo(Map<SurfaceInfoType<?>, Object> additionalInfo);
```
cjkent: This can never be implemented in a type safe way.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * The additional surface information.
+   * <p>
+   * This stores additional information for the surface.
+   */
+  private final Map<SurfaceInfoType<?>, Object> info = new HashMap<>();
+  /**
+   * The metadata about the parameters.
+   * <p>
+   * If present, the parameter metadata will match the number of parameters on the surface.
```
cjkent: Is this validated somewhere? When the metadata is added to the surface?

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadata.java
jodastephen: This is the standard wording from Joda-Beans used everywhere. I've changed it here.

cjkent: It should probably change in Joda Beans so that it's corrected everywhere.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
jodastephen: This will have been extracted from a Joda-Bean at some point.

modules/market/src/main/java/com/opengamma/strata/market/surface/DefaultSurfaceMetadataBuilder.java
jodastephen: Validated in `InterpolatedNodalSurface`.

PR #955, opened 2016-05-12
Swaption model lookup

jodastephen:
Add new calculation parameter `SwaptionMarketLookup`.
Will provide a replacement for `MarketDataMappings` when looking up data.
Allow the swaption volatilities, and thus the model, to be selected
Design allows for more intelligent lookup, such as where logic performs conversion. eg to convert LIBOR 3 month vols to LIBOR 6 month vols on the fly if desired.

Note that the market data key and ID are combined in this phase of the changes.


modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/DefaultSwaptionMarketLookup.java
```diff
@@ -0,0 +1,371 @@
...
+   * @param volatilityIds  the volatility identifiers, keyed by index
+   * @return the swaption lookup containing the specified volatilities
+   */
+  public static DefaultSwaptionMarketLookup of(Map<IborIndex, SwaptionVolatilitiesId> volatilityIds) {
+    return new DefaultSwaptionMarketLookup(volatilityIds);
+  }
+
+  //-------------------------------------------------------------------------
+  @Override
+  public ImmutableSet<IborIndex> getVolatilityIndices() {
```
cjkent: What's the intended use case of this method? It's currently unused.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/DefaultSwaptionMarketLookup.java
```diff
@@ -0,0 +1,371 @@
...
+    return volatilityIds.keySet();
+  }
+
+  @Override
+  public ImmutableSet<MarketDataKey<?>> getVolatilityIds(IborIndex index) {
```
cjkent: What's the intended use case of this method? It's currently unused.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  public abstract CalculationMarketData getMarketData();
+
+  /**
+   * Returns a copy of this instance with the specified market data.
+   * 
+   * @param marketData  the market data to use
+   * @return a market view based on the specified data
+   */
+  public abstract SwaptionCalculationMarketView withMarketData(CalculationMarketData marketData);
```
cjkent: What's the intended use case of this method? It's currently unused.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/DefaultSwaptionMarketLookup.java
jodastephen: The lookup interface is at a higher level than the market data interfaces. Where market data is opaque, and cannot tell you what it contains, the lookup objects can tell you what they contain. (A lookup implementation is typically going to be a hashmap from some kind of key to some kind of market data identifier). The `RatesMarketView` makes use of this when doing bucketed gamma. These swaption methods parallel those on rates.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
jodastephen: This allows the market data to be swapped out for another set, for example when bumping. This method matches the equivalent on rate market view, which is used in bucketed gamma.

cjkent: Where is it intended to be used? Inside a function? Where does the second set of `CalculationMarketData` come from?

jodastephen: Here is the code that uses it (in this case the `MarketData` version, not `CalculationMarketData`, but the two serve the same purpose)
```
    MarketData bumpedMarketData = marketView.getMarketData().withValue(curveId, bumpedCurve);
    RatesProvider bumpedRatesProvider = marketView.withMarketData(bumpedMarketData).ratesProvider();
    PointSensitivities pointSensitivities = PRICER.presentValueSensitivity(product, bumpedRatesProvider);
```

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionMarketLookup.java
```diff
@@ -0,0 +1,144 @@
...
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains swaption volatilities based on the specified market data.
+   * <p>
+   * The rates provider binds the lookup to the market data.
+   * 
+   * @param marketData  the complete set of market data for one scenario
+   * @return the volatilities
+   */
+  public abstract SwaptionVolatilities volatilities(IborIndex index, MarketData marketData);
```
cjkent: Why is this method here? It would seem more natural if it were on `SwaptionMarketView`

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Market data for swaptions, used for calculation across multiple scenarios.
+ * <p>
+ * This interface exposes the market data necessary for pricing a swaption.
+ * This includes volatilities and rates.
+ * <p>
+ * Implementations of this interface must be immutable.
+ */
+public interface SwaptionCalculationMarketView {
```
cjkent: I think "scenario" would be better than "calculation" in these names. It would make the purpose much clearer.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionMarketLookup.java
jodastephen: The method that standard functions use is on `SwaptionMarketView`. But I wanted the implementation of the method to be on the lookup. This means that you only need to implement an instance of the lookup interface to make changes (add your own logic). The two default implementations of `SwaptionCalculationMarketView` and `SwaptionMarketView` can then be shared. Without this method, it would be necessary for users to write three classes to change any logic in the lookup. As a side product, it also provides a useful shortcut to get the data if not using the market views.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
cjkent: I wonder if "data" would be clearer than "market"

    SwaptionDataLookup
    SwaptionScenarioDataView
    SwaptionDataView

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionMarketLookup.java
cjkent: I can see the reasoning, but I wonder if it muddies the picture too much. I've been trying to get a handle on the design and this method confuses things:

    MarketLookup - returns requirements and market views*
    CalculationMarketView - provides data for multiple scenarios
    MarketView - provides data for one scenario

* except the `volatilities` method which returns volatilities for a single scenario given market data for a single scenario.

See what I mean?

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
jodastephen: Happy with Scenario, I find just DataView a bit vague (its all data), and MarketDataView too long.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationFunction.java
```diff
@@ -90,17 +90,22 @@ public FunctionRequirements requirements(
...
-
     IborIndex index = product.getIndex();
-    DiscountCurveKey dfKey = DiscountCurveKey.of(product.getCurrency());
-    IborIndexCurveKey iborKey = IborIndexCurveKey.of(index);
-    SwaptionVolatilitiesKey volKey = SwaptionVolatilitiesKey.of(index);
-    IndexRateKey iborRateKey = IndexRateKey.of(index);
 
+    // use lookup to build requirements
+    SwaptionMarketLookup swaptionLookup = parameters.getParameter(SwaptionMarketLookup.class);
+    FunctionRequirements swaptionReqs = swaptionLookup.requirements(ImmutableSet.of(index));
```
cjkent: This makes me thinks that `SwaptionMarketLookup.requirements` should have a varargs parameter.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationFunction.java
```diff
@@ -90,17 +90,22 @@ public FunctionRequirements requirements(
...
+
+    ImmutableSet<MarketDataKey<?>> valueReqs = ImmutableSet.<MarketDataKey<?>>builder()
```
cjkent: Is the plan to get rid of this and replace it with a call to `RatesMarketLookup.requirements` or something like that?

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationMarketView.java
cjkent: I think we should canvas some more opinions. I find "market" vague. As a user I'd be thinking "Where do I get the data? Ah, a class called `Data`!"

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionCalculationFunction.java
jodastephen: Yes, idea is to call the two lookup objects and merge the results.

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionMarketLookup.java
```diff
@@ -0,0 +1,144 @@
...
+   */
+  public abstract ImmutableSet<IborIndex> getVolatilityIndices();
+
+  /**
+   * Gets the identifiers used to obtain the volatilities for the specified currency.
+   * <p>
+   * The result will typically refer to a surface or cube.
+   * If the index is not found, an exception is thrown.
+   *
+   * @param index  the index for which a discount curve is required
```
cjkent: Copy-paste error - it's not a discount curve

modules/function/src/main/java/com/opengamma/strata/function/calculation/swaption/SwaptionMarketLookup.java
```diff
@@ -0,0 +1,144 @@
...
+   */
+  public default SwaptionCalculationMarketView marketView(CalculationMarketData marketData) {
+    return DefaultSwaptionCalculationMarketView.of(this, marketData);
+  }
+
+  /**
+   * Obtains a filtered view of the complete set of market data.
+   * <p>
+   * This method returns an instance that binds the lookup to the market data.
+   * The input is {@link CalculationMarketData}, which contains market data for one scenario.
```
cjkent: Copy-paste error - it's `MarketData`, not `CalculationMarketData`

PR #957, opened 2016-05-13
Provide ability to override item in market data

jodastephen:
Add withValue(key, value) to `MarketData`


modules/basics/src/main/java/com/opengamma/strata/basics/market/ExtendedMarketData.java
```diff
@@ -0,0 +1,457 @@
...
+
+import com.opengamma.strata.collect.timeseries.LocalDateDoubleTimeSeries;
+
+/**
+ * A set of market data where an item has been added or overridden.
+ * <p>
+ * This decorates an underlying instance to add or replace a single identifier.
+ */
+@BeanDefinition(builderScope = "private", constructorScope = "package")
+final class ExtendedMarketData<T>
```
cjkent: Are these going to end up deeply nested? I can see that becoming a pain when debugging.

If so it might be worth having another implementation backed by a map of overrides to keep things simpler. This class could override `withValue` to create a map-backed version and the map-backed version could override `withValue` to create another map-backed impl.

Having said that, those changes can be made later.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ExtendedMarketData.java
```diff
@@ -0,0 +1,457 @@
...
+   * <p>
+   * The specified identifier can be queried in the result, returning the specified value.
+   * All other identifiers are queried in the underlying market data.
+   *
+   * @param key  the additional market data identifier
+   * @param value  the additional market data value
+   * @param underlying  the underlying market data
+   * @return a market data instance that decorates the original adding/overriding the specified identifier
+   */
+  public static <T> ExtendedMarketData<T> of(MarketDataKey<T> key, T value, MarketData underlying) {
```
cjkent: This doesn't need to be public given the class is package-private.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ExtendedMarketData.java
jodastephen: I thought we could add it later if we see its a problem. Most bumps I've seen are of a single thing on a base.

modules/basics/src/main/java/com/opengamma/strata/basics/market/ExtendedMarketData.java
jodastephen: True, but I don't bother with that in most cases as it causes no issues.

PR #959, opened 2016-05-13
Add rates lookup

jodastephen:
Lookup allows rates to be accessed by currency and index.
The code that uses rates lookup will follow in another PR.


modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
```diff
@@ -0,0 +1,280 @@
...
+import com.opengamma.strata.market.view.FxForwardRates;
+import com.opengamma.strata.market.view.FxIndexRates;
+import com.opengamma.strata.market.view.IborIndexRates;
+import com.opengamma.strata.market.view.OvernightIndexRates;
+import com.opengamma.strata.market.view.PriceIndexValues;
+import com.opengamma.strata.pricer.rate.AbstractRatesProvider;
+import com.opengamma.strata.pricer.rate.RatesProvider;
+
+/**
+ * A rates provider based on a rates lookup.
```
cjkent: A bit more detail would be good here. Just copying and pasting the second paragraph of Javadoc from the `of` method would be worthwhile.

cjkent: Actually, ignore that last comment.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
```diff
@@ -0,0 +1,280 @@
...
+  //-------------------------------------------------------------------------
+  @Override
+  public <T> T data(MarketDataKey<T> key) {
+    return marketData.getValue(key);
+  }
+
+  //-------------------------------------------------------------------------
+  @Override
+  public Optional<Curve> findCurve(CurveName name) {
+    return Stream.concat(lookup.getDiscountCurves().values().stream(), lookup.getForwardCurves().values().stream())
```
cjkent: How often are we lookup up curves by name? It seems like a pretty core operation and this doesn't seem like a very efficient way to do it.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
```diff
@@ -0,0 +1,280 @@
...
+@BeanDefinition(style = "light")
+final class DefaultLookupRatesProvider
+    extends AbstractRatesProvider
+    implements RatesProvider, ImmutableBean {
+
+  /**
+   * The lookup.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final DefaultRatesMarketDataLookup lookup;
```
cjkent: Why does this refer to the implementation type, not `RatesMarketDataLookup`. That doesn't smell good.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
```diff
@@ -0,0 +1,280 @@
...
+  //-------------------------------------------------------------------------
+  @Override
+  public double fxRate(Currency baseCurrency, Currency counterCurrency) {
+    return fxRateProvider.fxRate(baseCurrency, counterCurrency);
+  }
+
+  //-------------------------------------------------------------------------
+  @Override
+  public DiscountFactors discountFactors(Currency currency) {
+    SimpleCurveId curveId = lookup.getDiscountCurves().get(currency);
```
cjkent: This relies on referring to the implementation type. That's a bit of a smell. Do we need to add something to the interface?

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultRatesMarketData.java
```diff
@@ -0,0 +1,169 @@
...
+import org.joda.beans.Property;
+import org.joda.beans.PropertyDefinition;
+import org.joda.beans.impl.light.LightMetaBean;
+
+import com.opengamma.strata.basics.market.MarketData;
+import com.opengamma.strata.collect.ArgChecker;
+import com.opengamma.strata.pricer.rate.RatesProvider;
+
+/**
+ * The default market view for rates products.
```
cjkent: It's not called "view" any more.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultRatesMarketDataLookup.java
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  //-------------------------------------------------------------------------
+  @Override
+  public ImmutableSet<Currency> getDiscountCurveCurrencies() {
+    return discountCurves.keySet();
+  }
+
+  @Override
+  public ImmutableSet<MarketDataKey<?>> getDiscountCurveIds(Currency currency) {
```
cjkent: Can the type parameter be more specific? We know it's always a curve ID.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultRatesMarketDataLookup.java
```diff
@@ -0,0 +1,263 @@
...
+    return ImmutableSet.of(id);
+  }
+
+  @Override
+  public ImmutableSet<Index> getForwardCurveIndices() {
+    return forwardCurves.keySet();
+  }
+
+  @Override
+  public ImmutableSet<MarketDataKey<?>> getForwardCurveIds(Index index) {
```
cjkent: More specific type parameter?

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultRatesScenarioMarketData.java
```diff
@@ -0,0 +1,180 @@
...
+import org.joda.beans.MetaBean;
+import org.joda.beans.Property;
+import org.joda.beans.PropertyDefinition;
+import org.joda.beans.impl.light.LightMetaBean;
+
+import com.opengamma.strata.calc.marketdata.CalculationMarketData;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * The default market view for rates products, used for calculation across multiple scenarios.
```
cjkent: Not "market view".

modules/function/src/main/java/com/opengamma/strata/function/calculation/RatesMarketDataLookup.java
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Gets the identifiers used to obtain the discount curve for the specified currency.
+   * <p>
+   * If the currency is not found, an exception is thrown.
+   *
+   * @param currency  the currency for which identifiers are required
+   * @return the set of market data identifiers 
+   * @throws IllegalArgumentException if the currency is not found
+   */
+  public abstract ImmutableSet<MarketDataKey<?>> getDiscountCurveIds(Currency currency);
```
cjkent: More specific type parameter?

modules/function/src/main/java/com/opengamma/strata/function/calculation/RatesMarketDataLookup.java
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Gets the identifiers used to obtain the forward curve for the specified index.
+   * <p>
+   * If the index is not found, an exception is thrown.
+   *
+   * @param index  the index for which identifiers are required
+   * @return the set of market data identifiers 
+   * @throws IllegalArgumentException if the index is not found
+   */
+  public abstract ImmutableSet<MarketDataKey<?>> getForwardCurveIds(Index index);
```
cjkent: More specific type parameter?

cjkent:
That's everything

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
jodastephen: Currently there is not much lookup of curves by name. This can be more efficient, but only if we pass the curve identifier in sensitivities and other places. I haven't looked at that so far, as it hasn't been as necessary as I though it would be. Right now, this method is the best it can be in performance terms.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultLookupRatesProvider.java
jodastephen: The maps of discount currency to identifier and forward index to identifier are not public on the interface (as the interface could be implemented in other ways. This class is essentially an inner class of `DefaultRatesMarketDataLookup` and is package scope, so it is fine as is.

modules/function/src/main/java/com/opengamma/strata/function/calculation/DefaultRatesMarketDataLookup.java
jodastephen: In theory, the lookup could be based on `DiscountFactors` rather than a curve, or on data extracted from a surface. ie. the interface is deliberately designed to not commit the implementation to being a curve. (same as is done for swaptions, where volatilities might be from a surface, cube or something else). Probably the method name is wrong.

modules/function/src/main/java/com/opengamma/strata/function/calculation/RatesMarketDataLookup.java
```diff
@@ -144,40 +144,42 @@ public static RatesMarketDataLookup of(CurveGroupDefinition curveGroupDefinition
...
    * @param currency  the currency for which identifiers are required
    * @return the set of market data identifiers 
    * @throws IllegalArgumentException if the currency is not found
    */
-  public abstract ImmutableSet<MarketDataKey<?>> getDiscountCurveIds(Currency currency);
+  public abstract ImmutableSet<MarketDataKey<?>> getDiscountMarketDataIds(Currency currency);
 
   /**
-   * Gets the set of currencies that forward curves are provided for.
+   * Gets the set of currencies that forward rates are provided for.
```
cjkent: s/currencies/indices/

PR #983, opened 2016-05-23
Parameter metadata

jodastephen:
Change from curve/surface parameter metadata to general concept.

PR is mostly auto-refactoring of class names and package names.


cjkent:
Will review over-the-shoulder as part of a larger review.

PR #1008, opened 2016-06-01
Add helper for combining ParameterizedData

jodastephen:
Combining instances is tricky, this helper makes it reliable
Also, fix incorrect used of Preconditions

### DIFF
--- a/modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
+++ b/modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
@@ -1,0 +1,172 @@
+/**
+ * Copyright (C) 2016 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.market.param;
+
+import com.google.common.base.Preconditions;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * Helper that can be used to combine two or more underlying instances of {@code ParameterizedData}.
+ * <p>
+ * This is used by implementations of {@link ParameterizedData} that are based on more
+ * than one underlying {@code ParameterizedData} instance.
+ * <p>
+ * This helper should be created in the constructor of the combined instance.
+ * In each of the five {@code ParameterizedData} methods of the combined instance,
+ * this helper should be invoked. See {@code DiscountFxForwardRates} for sample usage.
+ */
+public final class ParameterizedDataCombiner {
+
+  /**
+   * The underlying instances.
+   */
+  private final ParameterizedData[] underlyings;
+  /**
+   * The lookup array.
+   */
+  private final int[] lookup;
+  /**
+   * The count of parameters.
+   */
+  private final int paramCount;
+
+  /**
+   * Obtains an instance that can combine the specified underlying instances.
+   * 
+   * @param instances  the underlying instances to combine
+   * @return the combiner
+   */
+  public static ParameterizedDataCombiner of(ParameterizedData... instances) {
+    return new ParameterizedDataCombiner(instances);
+  }
+
+  //------------------------------------------------------------------------- 
+  // creates an instance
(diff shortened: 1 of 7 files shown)


modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
```diff
@@ -0,0 +1,172 @@
...
+
+  //------------------------------------------------------------------------- 
+  // creates an instance
+  private ParameterizedDataCombiner(ParameterizedData[] underlyings) {
+    ArgChecker.notEmpty(underlyings, "underlyings");
+    int size = underlyings.length;
+    this.underlyings = underlyings;
+    int[] lookup = new int[size];
+    for (int i = 1; i < size; i++) {
+      lookup[i] = lookup[i - 1] + underlyings[i - 1].getParameterCount();
```
jacbop: should we validate that none of the underlyings are null upon construction?

modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
```diff
@@ -0,0 +1,172 @@
...
+   */
+  public static ParameterizedDataCombiner of(ParameterizedData... instances) {
+    return new ParameterizedDataCombiner(instances);
+  }
+
+  //------------------------------------------------------------------------- 
+  // creates an instance
+  private ParameterizedDataCombiner(ParameterizedData[] underlyings) {
+    ArgChecker.notEmpty(underlyings, "underlyings");
+    int size = underlyings.length;
```
jacbop: Validate that size > 0, or is it okay to have an empty combiner?

modules/market/src/test/java/com/opengamma/strata/market/param/ParameterizedDataCombinerTest.java
```diff
@@ -0,0 +1,98 @@
...
+    assertEquals(test.getParameter(1), 2d);
+    assertEquals(test.getParameter(2), 3d);
+    assertEquals(test.getParameter(3), 4d);
+    assertEquals(test.getParameter(4), 5d);
+    assertEquals(test.getParameterMetadata(0), ParameterMetadata.empty());
+    assertThrows(() -> test.getParameter(-1), IndexOutOfBoundsException.class);
+    assertThrows(() -> test.getParameter(5), IndexOutOfBoundsException.class);
+    assertThrows(() -> test.getParameterMetadata(-1), IndexOutOfBoundsException.class);
+    assertThrows(() -> test.getParameterMetadata(5), IndexOutOfBoundsException.class);
+    assertThrowsIllegalArg(() -> ParameterizedDataCombiner.of());
```
jacbop: I see the test with empty (size = 0) underlyings, but can we add tests with one or more null underlyings. I assume that should also throw an illegal arg exception.

modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
jodastephen: The approach in Strata is not to validate for null, except when creating state. Strata also allows an IAE or NPE when validating state. So, since this code will not allow nulls into the state (because an NPE will be thrown in the constructor) we treat it as fine.

The rationale is that we have almost eliminated null in Strata, thus validating against it is, for the most part, pointless.

modules/market/src/test/java/com/opengamma/strata/market/param/ParameterizedDataCombinerTest.java
jodastephen: As above, we don't tend to write tests for null.

modules/market/src/main/java/com/opengamma/strata/market/param/ParameterizedDataCombiner.java
jacbop: Can we add something along these lines to the code standards in the wiki? I did a search there before making these comments, but your explanation makes sense. I'll merge it now.

Issue #847, opened 2016-03-17
Add high-level swaption pricing API


traggatt:
There only needs to be two - one for the SwaptionTrade and another for the SwaptionProduct (i.e. with and without premium, respectively).

The current set of pricers is the cartesian product of:

- 2 settlement types: cash/physical
- 3 model types: black, SABR and normal, plus another general supertype pricer for shared measure implementations

Plus:

- 1 hull/white pricer which only works for `SwaptionProduct`
- another 7 which only work for `SwaptionTrade`

For most of these elements, it should be possible to infer the model, and settlement type from the input arguments. The model is defined by the type of `Volatilities` object passed and the settlement type by the `SwaptionProduct`.

There does not appear to be any reason why the public API can't expose a single set of functions as most pricing functions have the same argument and result types per product/trade. 

In the case where additional functions exist for a specific model type, these could still exist on the same pricing API but take a narrower set of arguments. e.g:

Function callable for any model and settlement type:
````
  public CurrencyAmount presentValue(
      SwaptionProduct swaption,
      RatesProvider ratesProvider,
      SwaptionVolatilities swaptionVolatilities)
````

Function callable only for SABR (since it only takes `SabrSwaptionVolatilities` as opposed to more general `SwaptionVolatilities`):
````
  public PointSensitivityBuilder presentValueSensitivity(
      SwaptionProduct swaption,
      RatesProvider ratesProvider,
      SabrSwaptionVolatilities swaptionVolatilities) {
````

jonathansenior:
Very valid point. I think there's simply a missing high-level API here. A lot of effort has already gone in to normalising the swaption pricing in Strata so that ultimately a single pricer is driven by different input data types (rather than having entirely separate pricers, as in OG-Platform). It makes sense that we should be able to benefit from this with a single swaption pricing API.

jodastephen:
Done.

PR #1031, opened 2016-06-06
Provide swaption pricing across cash and physical

jodastephen:
Single entry point for product and trade.

Javadoc in new code copied from existing pricers.
See #847

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
@@ -1,0 +1,166 @@
+/**
+ * Copyright (C) 2016 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.pricer.swaption;
+
+import java.time.LocalDate;
+
+import com.opengamma.strata.basics.currency.CurrencyAmount;
+import com.opengamma.strata.basics.currency.MultiCurrencyAmount;
+import com.opengamma.strata.basics.currency.Payment;
+import com.opengamma.strata.collect.ArgChecker;
+import com.opengamma.strata.market.product.swaption.SwaptionSensitivity;
+import com.opengamma.strata.market.product.swaption.SwaptionVolatilities;
+import com.opengamma.strata.market.sensitivity.PointSensitivityBuilder;
+import com.opengamma.strata.pricer.DiscountingPaymentPricer;
+import com.opengamma.strata.pricer.rate.RatesProvider;
+import com.opengamma.strata.product.swaption.ResolvedSwaption;
+import com.opengamma.strata.product.swaption.ResolvedSwaptionTrade;
+
+/**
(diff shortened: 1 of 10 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
```diff
@@ -0,0 +1,166 @@
...
+   * <p>
+   * The present value sensitivity of the product is the sensitivity of the present value to
+   * the underlying curves.
+   * 
+   * @param trade  the swaption trade
+   * @param ratesProvider  the rates provider
+   * @param swaptionVolatilities  the volatilities
+   * @return the present value curve sensitivity of the swap trade
+   */
+  public PointSensitivityBuilder presentValueSensitivityStickyStrike(
```
traggatt: Is it documented anywhere what the expected unit for a call like this is? As a user I would find it useful for the method to state what the magnitude of the sensitivity is, e.g. to 1bp, to 1%, 100% etc. If not stated, I would probably assume 1bp, though I know this isn't always the case at the pricer level...

traggatt: Might also be useful, in this case, to state what the implication of the sticky strike is.

modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
```diff
@@ -0,0 +1,166 @@
...
+  public MultiCurrencyAmount currencyExposure(
+      ResolvedSwaptionTrade trade,
+      RatesProvider ratesProvider,
+      SwaptionVolatilities swaptionVolatilities) {
+
+    return MultiCurrencyAmount.of(presentValue(trade, ratesProvider, swaptionVolatilities));
+  }
+
+  /**
+   * Calculates the current of the swaption trade.
```
traggatt: current cash

modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
```diff
@@ -0,0 +1,166 @@
...
+    ResolvedSwaption product = trade.getProduct();
+    CurrencyAmount pvProduct = productPricer.presentValue(product, ratesProvider, swaptionVolatilities);
+    Payment premium = trade.getPremium();
+    CurrencyAmount pvPremium = paymentPricer.presentValue(premium, ratesProvider);
+    return pvProduct.plus(pvPremium);
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Computes the currency exposure of the swaption trade
```
traggatt: Missing full stop! 😳

modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/VolatilitySwaptionTradePricer.java
jodastephen: Will deal with units in a later PR.

PR #1051, opened 2016-06-13
Store PV sensitivity to Market Quotes at calibration. 

marc-opengamma:
The curve calibrator currently store the Jacobian matrices for any calibration performed. 
This PR extend the type of info stored at calibration time to the "present value sensitivity to market quote". This type of info is useful to compute the notional equivalent to a given sensitivity. The PR also add flags on 'CurveGroupDefinition' to indicate which of those info should be computed and stored. 
The info stored can be used in the 'NotionalEquivalentCalculator' to obtain the notional of each instrument used in the calibration to obtain an equivalent sensitivity. This is tested in 'CalibrationNotionalEquivalentTest'.
Change the 'CurveNode.trade' to take a quantity a input. This is to be able to create node instruments with the equivalent notional.


modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/CurveCalibrator.java
```diff
@@ -71,6 +72,14 @@
    */
   private final CalibrationMeasures measures;
 
+  private static final CalibrationMeasures PV_MEASURES = CalibrationMeasures.of(
```
jodastephen: This should be an instance variable and not declared here.

modules/market/src/main/java/com/opengamma/strata/market/curve/CurveNode.java
```diff
@@ -50,15 +50,18 @@
    * This uses the observed market data to build the trade that the node represents.
    * The reference data is typically used to find the start date of the trade from the valuation date.
    * The resulting trade is not resolved.
+   * The notional of the trade is taken from the 'quantity' variable.
+   * The quantity is signed and will affect whether the trade is Buy or Sell.
```
jodastephen: Need to define buy/sell quantity on each node.

PR #1067, opened 2016-06-16
Swap measures

jodastephen:
New measures API for swaps


modules/measure/src/main/java/com/opengamma/strata/measure/swap/SwapMeasureCalculations.java
```diff
@@ -58,9 +59,14 @@
 final class SwapMeasureCalculations {
```
jonathansenior: Just to note that the XMeasureCalculations classes vary between final and non-final (although this only affects us due to their scope) - probably worth standardising these in a future patch.

PR #1080, opened 2016-06-20
Remove standard identifier from observables

jodastephen:
No need for `ObservableId` to refer to `StandardId` or `FieldName`


cjkent:
The original idea of `ObservableId` was that it represented a piece of data observable in an external data source and would therefore always be associated with an ID (e.g. a ticker) and field name.

Even if we don't actually use that fact in any of our code, is it worth leaving it there to make it clear what an observable ID is? What's the advantage of removing it?

jodastephen:
The `getStandardId()` method had been reduced to forcing a silly implementation in `IndexQuoteId` where an identifier was made up to satisfy the API. When we removed `ObservableIdMapping` we essentially removed the need for the standard id and field name here.

cjkent:
If we need to build up market data from raw data from (say) Bloomberg, how do we map index quotes to tickers?

jodastephen:
The implementor of `ObservableDataProvider` would be expected to map `QuoteId` or `IndexQuoteId` to their identifier. Forcing a conversion to `StandardId` limits their options for no gain.

PR #1094, opened 2016-06-21
Bond yield: allow 0 coupon

marc-opengamma:
Change yield for fixed coupon bonds to allow coupons with value 0. Previously a coupon of 0 was returning an error when computing yield.


jonathansenior:
Reported here: http://forums.opengamma.com/t/zero-coupon-bond-yield/530/2

PR #1112, opened 2016-06-24
Add getIds to MarketData and ScenarioMarketData

jonathansenior:
Allows the `CdsScenarioExample` to be simplified as shown.

### DIFF
--- a/modules/data/src/main/java/com/opengamma/strata/data/CombinedMarketData.java
+++ b/modules/data/src/main/java/com/opengamma/strata/data/CombinedMarketData.java
@@ -76,6 +76,14 @@ final class CombinedMarketData
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
+        .addAll(underlying1.getIds())
+        .addAll(underlying2.getIds())
+        .build();
+  }
+
+  @Override
   @SuppressWarnings("unchecked")
   public <T> Set<MarketDataId<T>> findIds(MarketDataName<T> name) {
     return ImmutableSet.<MarketDataId<T>>builder()
--- a/modules/data/src/main/java/com/opengamma/strata/data/ExtendedMarketData.java
+++ b/modules/data/src/main/java/com/opengamma/strata/data/ExtendedMarketData.java
@@ -95,6 +95,14 @@ final class ExtendedMarketData<T>
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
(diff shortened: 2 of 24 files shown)


modules/data/src/main/java/com/opengamma/strata/data/CombinedMarketData.java
```diff
@@ -76,6 +76,14 @@ public boolean containsValue(MarketDataId<?> id) {
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
```
jodastephen: This will throw an exception if the same id is in both underlying

modules/data/src/main/java/com/opengamma/strata/data/ExtendedMarketData.java
```diff
@@ -95,6 +95,14 @@ public boolean containsValue(MarketDataId<?> id) {
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
+        .addAll(underlying.getIds())
```
jodastephen: Exception if id is in underlying

modules/data/src/main/java/com/opengamma/strata/data/scenario/CombinedScenarioMarketData.java
```diff
@@ -105,6 +106,14 @@ public boolean containsValue(MarketDataId<?> id) {
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
+        .addAll(underlying1.getIds())
```
jodastephen: Exception if id is in underlying

modules/data/src/main/java/com/opengamma/strata/data/scenario/ExtendedScenarioMarketData.java
```diff
@@ -119,6 +119,14 @@ public boolean containsValue(MarketDataId<?> id) {
   }
 
   @Override
+  public Set<MarketDataId<?>> getIds() {
+    return ImmutableSet.<MarketDataId<?>>builder()
+        .addAll(underlying.getIds())
```
jodastephen: Exception if id is in underlying

modules/data/src/main/java/com/opengamma/strata/data/CombinedMarketData.java
jodastephen: Turns out that Guava doesn't allow duplicates in maps but does in sets.

PR #1120, opened 2016-06-27
Ibor future curve nodes

jodastephen:
Enable both relative and absolute curve nodes for Ibor future using two different templates.
Avoid hard coding the security identifier.


modules/market/src/main/java/com/opengamma/strata/market/observable/QuoteId.java
```diff
@@ -123,6 +124,19 @@ public QuoteId withObservableSource(ObservableSource obsSource) {
...
+  /**
+   * Returns a security identifier based on the standard identifier.
+   * <p>
+   * This convenience method allows a quote identifier to be converted to a security identifier.
+   * Since not all quotes are for securities, there is no guarantee that a security exists
+   * for the resulting identifier.
+   * 
+   * @return a security identifier formed by wrapping the standard identifier of this instance
+   */
+  public SecurityId toSecurityId() {
```
cjkent: Is this method worth having given how simple it is and how a quote isn't guaranteed to be for a security?

Would it be better to have `SecurityId.of(QuoteId)` so you're putting the onus on the user of knowing a quote represents a security?

modules/product/src/main/java/com/opengamma/strata/product/index/type/IborFutureTemplate.java
```diff
@@ -77,37 +36,54 @@
...
+   * <p>
+   * The future is selected from a sequence of futures based on a year-month.
+   * In most cases, the date of the future will be in the same month as the specified month,
+   * but this is not guaranteed.
+   * 
+   * @param yearMonth  the year-month to use to select the future
+   * @param convention  the future convention
+   * @return the template
+   */
+  public static AbsoluteIborFutureTemplate of(YearMonth yearMonth, IborFutureConvention convention) {
```
cjkent: I think the return type should be `IborFutureTemplate` as `AbsoluteIborFutureTemplate` is package-private.

PR #1126, opened 2016-06-28
Create sensible default parameter metadata

jodastephen:
`NodalCurve` and `NodalSurface` can have a better default parameter metadata.


traggatt:
Thanks. Will switch to using these classes after the next release.

PR #1133, opened 2016-06-29
Add explain PV for CMS products.

### LINKED ISSUE #1131: PV explain for CMS product
A PV explain method exists at the CMS leg level but not at the CMS product level. That method should be created. It will simply aggregate the PV explain of both legs.

marc-opengamma:
Closes #1131


modules/pricer/src/main/java/com/opengamma/strata/pricer/swap/DiscountingSwapLegPricer.java
```diff
@@ -593,8 +593,14 @@ CashFlows cashFlowEventsInternal(ResolvedSwapLeg leg, RatesProvider provider) {
   }
 
   //-------------------------------------------------------------------------
-  // explains the present value of the leg
-  void explainPresentValueInternal(ResolvedSwapLeg leg, RatesProvider provider, ExplainMapBuilder builder) {
+  /**
```
traggatt: Hmm, not sure if a method named `explainPresentValueInternal` should really be exposed as part of the public API. I'd suggest making it package-private again, then adding another method `explainPresentValue` which calls it passing in a clean builder. It could then return a fully built `ExplainMap`.

PR #1153, opened 2016-07-04
Better handling of results

jodastephen:
Add casting behaviour to `CalculationResult`
Better Javadoc explaining failures


modules/calc/src/main/java/com/opengamma/strata/calc/Results.java
```diff
@@ -207,17 +219,31 @@ private String invalidColumnNameMessage(ColumnName columnName) {
...
    */
   public <T> Result<T> get(int rowIndex, ColumnName columnName, Class<T> type) {
-    return get(rowIndex, columnName).map(v -> type.cast(v));
+    return cast(get(rowIndex, columnName), type);
+  }
+
+  @SuppressWarnings("unchecked")
+  private <T> Result<T> cast(Result<?> result, Class<T> type) {
+    // cannot use result.map() as we want the exception to be thrown
+    if (result.isFailure() || type.isInstance(result.getValue())) {
```
traggatt: Could we add a `cast()` to `Result` itself, maybe?

cjkent: `Result.getValue` returns a `T` so a cast method isn't needed. Just cast the value:

    Result<?> result = ...;
    String str = (String) result.getValue();

is just as good as

    Result<?> result = ...;
    String str = result.getValue(String.class);

and doesn't require an extra method.

It's valuable here because `CalculationResult` returns `Result<T>` so there's no way to get a `Result<Foo>` from a `CalculationResult<?>` without an unchecked cast.

traggatt: Only it would mean you wouldn't need `@SuppressWarnings("unchecked")` here. `Class` itself implements this pattern with `cast()`.

traggatt: Actually, just to clarify, the cast signature I was suggesting would return another `Result`, not the contained instance. e.g.

`public <S extends T> Result<S> cast(Class<S> targetClazz)`

I find this is quite a common use case.

PR #1161, opened 2016-07-05
Nodal curve node addition does not require index

### LINKED ISSUE #1159: Index parameter on NodalCurve.withNode is redundant
`NodalCurve.withNode` and `InterpolatedNodalCurve.withNode` require the node index and the x value to be specified. This is redundant as the index can be derived from the x value and having both parameters means it is possible to provide contradictory arguments.

jodastephen:
Determine whether to add or replace by x-value
Only have method that accepts parameter metadata
Additional tests
Fixes #1159

### DIFF
--- a/modules/market/src/test/java/com/opengamma/strata/market/curve/InterpolatedNodalCurveTest.java
+++ b/modules/market/src/test/java/com/opengamma/strata/market/curve/InterpolatedNodalCurveTest.java
@@ -18,6 +18,7 @@ import java.util.List;
 
 import org.testng.annotations.Test;
 
+import com.opengamma.strata.basics.currency.Currency;
 import com.opengamma.strata.collect.array.DoubleArray;
 import com.opengamma.strata.market.ValueType;
 import com.opengamma.strata.market.curve.interpolator.BoundCurveInterpolator;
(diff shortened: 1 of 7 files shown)


modules/market/src/test/java/com/opengamma/strata/market/curve/InterpolatedNodalCurveTest.java
```diff
@@ -146,58 +159,51 @@ public void test_withYValues_badSize() {
...
     assertThat(test.getParameterCount()).isEqualTo(SIZE + 1);
     assertThat(test.getMetadata()).isEqualTo(METADATA);
-    assertThat(test.getXValues()).isEqualTo(x);
-    assertThat(test.getYValues()).isEqualTo(y);
+    assertThat(test.getXValues()).isEqualTo(DoubleArray.of(0.5d, 1d, 2d, 3d));
+    assertThat(test.getYValues()).isEqualTo(DoubleArray.of(4d, 5d, 7d, 8d));
   }
 
-  public void test_withNode_atStart_metadata() {
+  public void test_withNode_replace_withMetadata() {
```
cjkent: There's no test for inserting a new node in the middle of the curve, only at the start. Is that worth having?

PR #1192, opened 2016-07-11
Add support for price measure

jodastephen:
Ibor future and DSF return price calculated from the model


jonathansenior:
Much clearer - thanks.

PR #1234, opened 2016-07-22
Add PartialResult

### LINKED ISSUE #1230: Result class which can include both a result and errors
The `Result` class is basically a variation on `Either`; it can contain a result or errors but not both.

There are often cases where an error is not fatal and a value can still be calculated, but the details of any errors are also required.

A `PartialResult` would be useful. It should always contain a value, even if it is empty. It can also optionally contain one or more `Failure` or `FailureItem` instances (not sure which would work better).

jodastephen:
New class for errors during loading
Enhance FailureItem with public factory methods and tests
Enhance stack trace generation
Fixes #1230


modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -69,6 +72,82 @@
...
+   * Obtains a failure from a reason and message.
+   * <p>
+   * An exception will be created internally to obtain a stack trace.
+   * The cause type will not be present in the resulting failure.
+   * 
+   * @param reason  the reason
+   * @param message  the failure message, not empty
+   * @return the failure
+   */
+  public static FailureItem of(FailureReason reason, String message) {
```
cjkent: I imagine a common use case would be something like:

    FailureItem item = FailureItem.of(FailureReason.ERROR, Messages.format("Couldn't build a Foo from {}", fooStr);

Would it be worth having that built in?

    public static FailureItem of(FailureReason reason, String message, Object... messageArgs)

modules/collect/src/test/java/com/opengamma/strata/collect/result/PartialResultTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Test {@link PartialResult}.
+ */
+@Test
+public class PartialResultTest {
```
jacbop: It seems that the callers need to decide ahead of time if they want the Result (Success/Failure) pattern or the PartialResult pattern.

If they choose the PartialResult path (as in the loading rows from a file example), then it appears that we expect that they will always get some form of valid result, and *may* get one or more errors.

They case I don't see and can imagine happening is where the caller is expected a PartialResult<T> and we don't have any valid result T. In this case, is the type T responsible for contain information as to whether it is valid or not?

So let's say we want to upload a portfolio of trades via:
```
PartialResult<MarginPortfolio> uploadTrades(List<TradeFile> trades);
```

What happens if all of the trades have errors and is now a FailureItem on the PartialResult? What if the MarginPortfolio does not support an empty set? Do we throw? Do we need to enforce that type T support some concept of failure or not being valid?

jodastephen: The expectation is that the success value is a list, set or map capable of holding zero entries. Doing otherwise seems to complicate the state further.

cjkent: This design matches exactly what I've been using in a client project. The value has to have a notion of empty. In most cases it's an empty list or an empty map, but in some cases it's a custom data structure which can be empty. It's up to the caller to decide whether it's successful or not.

cjkent: You could always use a `Result` as your value if you wanted to specifically signal a failure and couldn't provide a sensible empty value.

jacbop: Makes sense. Maybe worth adding some java doc and a test case to the unit tests to illustrate.

jodastephen: Docs added.

cjkent:
I'm not sure about the name `PartialResult`, even though it was my suggestion. In most cases the result will be a complete result, not a partial result, so it doesn't feel right for the class name to focus on the minority case.

Having said that, I don't have a better suggestion yet.

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureReason.java
```diff
@@ -14,50 +14,67 @@
...
   /**
-   * Some aspect of the calculation in the function has failed.
+   * The operation could not be performed.
+   * <p>
+   * This is used to indicate that a calculation failed.
    */
   CALCULATION_FAILED,
   /**
    * Failure occurred for some other reason.
+   * ,p>
```
cjkent: Something has gone wrong here.

cjkent:
Merging - let's come back to the name

PR #1240, opened 2016-07-22
Create column name directly from measure

### LINKED ISSUE #1239: Create column name directly from measure
We can create a `Column` using just a `Measure`, but not a `ColumnName`. This is annoying when using the extended `Results` API.

jodastephen:
Add factory method to create `ColumnName` from `Measure`
Fixes #1239

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/ColumnName.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/ColumnName.java
@@ -40,7 +40,6 @@ public final class ColumnName
    * @param measure  the measure to extract the name from
    * @return a column with the same name as the measure
    */
-  @FromString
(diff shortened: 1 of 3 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/ColumnName.java
```diff
@@ -33,6 +33,19 @@ public static ColumnName of(String name) {
...
 
   /**
+   * Obtains an instance from the specified measure.
+   * <p>
+   * The column name will be the same as the name of the measure.
+   *
+   * @param measure  the measure to extract the name from
+   * @return a column with the same name as the measure
+   */
+  @FromString
```
cjkent: Presumably this annotation shouldn't be here?

Issue #1269, opened 2016-08-08
ResourceLocator does not handle files in zip


jodastephen:
`ResourceLocator` has a method accepting `Path`, but it only works for files and not for files within zip files. Enhance `ResourceLocator` with general URL support and delegate to that for `Path` when necessary.

jodastephen:
Fixed by #1271

PR #1271, opened 2016-08-09
Support paths properly in ResourceLocator

jodastephen:
Previously, files in zips were not supported
This will matter more with JFS in Java 9
#1269


examples/src/main/java/com/opengamma/strata/examples/finance/CalibrationCheckExample.java
```diff
@@ -73,22 +74,22 @@
    * The location of the curve calibration groups file.
    */
   private static final ResourceLocator GROUPS_RESOURCE =
-      ResourceLocator.of(ResourceLocator.FILE_URL_PREFIX + PATH_CONFIG + "curves/groups.csv");
+      ResourceLocator.ofFile(new File(PATH_CONFIG + "curves/groups.csv"));
```
cjkent: Isn't it a bit retro to be using `File`? Why not

    ResourceLocator.ofPath(Paths.get(PATH_CONFIG).resolve("curves/groups.csv"));

PR #1279, opened 2016-08-16
Add CSV row-by-row parsing

### LINKED ISSUE #1278: CSV parsing row-by-row
Provide the ability to parse a CSV file row-by-row, without loading it fully into memory.

jodastephen:
Handles large files
Fixes #1278

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
@@ -6,19 +6,20 @@
 package com.opengamma.strata.collect.io;
 
 import static com.opengamma.strata.collect.Guavate.toImmutableList;
-import static java.util.stream.Collectors.toCollection;
 
+import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.stream.Stream;
+import java.util.function.Consumer;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.CharSource;
+import com.google.common.io.LineProcessor;
 import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.collect.Unchecked;
 
@@ -63,7 +64,7 @@ public final class CsvFile {
    * Parses the specified source as a CSV file.
    * 
    * @param source  the CSV file resource
-   * @param headerRow  whether the source has a header row
+   * @param headerRow  whether the source has a header row, an empty source must still contain the header
    * @return the CSV file
    * @throws UncheckedIOException if an IO exception occurs
    * @throws IllegalArgumentException if the file cannot be parsed
@@ -79,7 +80,7 @@ public final class CsvFile {
    * For example, a tab-separated file is very similar to a CSV file, the only difference is the separator.
    * 
    * @param source  the file resource
-   * @param headerRow  whether the source has a header row
+   * @param headerRow  whether the source has a header row, an empty source must still contain the header
    * @param separator  the separator used to separate each field, typically a comma, but a tab is sometimes used
    * @return the TSV file
    * @throws UncheckedIOException if an IO exception occurs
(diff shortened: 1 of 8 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
```diff
@@ -88,29 +89,135 @@ public static CsvFile of(CharSource source, boolean headerRow) {
...
+   * <p>
+   * The specified consumer is invoked for each row.
+   * 
+   * @param source  the CSV file resource
+   * @param headerRow  whether the source has a header row, an empty source may omit the header
+   * @param consumer  the row consumer
+   * @throws UncheckedIOException if an IO exception occurs
+   * @throws IllegalArgumentException if the file cannot be parsed
+   */
+  public static void parse(CharSource source, boolean headerRow, Consumer<CsvRow> consumer) {
```
cjkent: This is an API which requires a stateful implementation. The consumer must be a mutable accumulator of data. Also, if the file has any structure beyond a simple flat file the consumer must contain a mutable state indicator or state machine.

A method returning an `Iterator<CsvRow>` would be much easier to work with. It might also be worth looking at Guava's `PeekingIterator`.

jodastephen: Because this is dealing with file open/close, an iterator is not sufficient. It would have to return an `AutoCloseable` with the caller expected to use it with care, and requires an additional public class.

cjkent: Designing the API to handle closing of the resource limits your options; you are bound to end up with an API which requires a stateful accumulator. I would prefer an API which leaves resource management to the user (which is simple) but allows an easier implementation for more complicated file formats.

There are parallels with XML parsing in Java. The original XML API was SAX which was conceptually similar to this API. The user provided a listener to the XML parser and the listener was necessarily stateful. SAX was painful to use and was superseded by the much easier to use StAX which has a pull-based API, much like an iterator.

We use a third-party CSV parser in one of our client projects which has complex CSV-based file formats. This API isn't sufficient for that use case and wouldn't enable us to remove the third-party parser. Migrating those file loaders to be stateful implementations of `Consumer` would make the already complicated code even more complex.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvIterator.java
```diff
@@ -0,0 +1,281 @@
...
+ * Provides the ability to iterate over a CSV file together with the ability to parse it from a {@link CharSource}.
+ * The separator may be specified, allowing TSV files (tab-separated) and other similar formats to be parsed.
+ * See {@link CsvFile} for more details of the CSV format.
+ * <p>
+ * This class processes the CSV file row-by-row.
+ * To load the entire CSV file into memory, use {@link CsvFile}.
+ * <p>
+ * This class must be used in a try-with-resources block to ensure that the underlying CSV file is closed:
+ * <pre>
+ *  try (CsvIterable iterable = CsvIterable.of(source, true)) {
```
cjkent: There are quite a few references to `CsvIterable` in the Javadoc and in some message strings. Did the class have a late name change but the docs and strings didn't get updated?

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvIterator.java
```diff
@@ -0,0 +1,281 @@
...
+   */
+  public static CsvIterator of(CharSource source, boolean headerRow, char separator) {
+    ArgChecker.notNull(source, "source");
+
+    return Unchecked.wrap(() -> {
+      BufferedReader reader = source.openBufferedStream();
+      if (!headerRow) {
+        return new CsvIterator(reader, separator, ImmutableList.of(), ImmutableMap.of());
+      }
+      String line = reader.readLine();
```
cjkent: If this line or the following if-block throws an exception the reader will not be closed.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvIterator.java
```diff
@@ -0,0 +1,281 @@
...
+  public Stream<CsvRow> asStream() {
+    Spliterator<CsvRow> spliterator = Spliterators.spliteratorUnknownSize(this, Spliterator.ORDERED | Spliterator.NONNULL);
+    return StreamSupport.stream(spliterator, false);
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Checks whether there is another row in the CSV file.
+   * 
+   * @return the peeked row
```
cjkent: This isn't right.

cjkent:
That's everything

PR #1285, opened 2016-08-17
Added transfer of FX rate in synthetic curve calibration. Closes #1284

### LINKED ISSUE #1284: SyntheticCurveCalibrator does not transfer FX rates
In the SyntheticCurveCalibrator, new curves are calibrated to provided settings, but the FX rates that may have been available in the initial RatesProvider are not transfer to the new provider. An exchange rate between any currency in the new settings should be be available in the new RatesProvider created.

marc-opengamma:
Closes #1284


modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/SyntheticCurveCalibrator.java
```diff
@@ -172,6 +181,13 @@ MarketData marketData(
         mapIdSy.put(k, mq);
       }
     }
+    // Generate quotes for FX pairs. The first currency is arbitrarily selected as starting point. 
+    // The crosses are automatically generated by the MarketDataFxRateProvider used in calibration.
+    for (int loopccy = 1; loopccy < ccyRequired.size(); loopccy++) {
+      CurrencyPair ccyPair = CurrencyPair.of(ccyRequired.get(0), ccyRequired.get(loopccy));
```
jodastephen: Need to handle that `ccyRequired` might be empty

PR #1299, opened 2016-08-23
Add ReferenceData.minimal()

### LINKED ISSUE #1291: Provide set of minimal reference data without currency holiday calendars
The standard reference data contains default values for the holiday calendars for some major currencies. It also contains a holiday calendar called `NO_HOLIDAYS` which is used in the implementation of some Strata classes.

It is common for users to supply their own holiday data instead of using the built-in calendars. The requires the user to supply a set of reference data. This reference data must include `NO_HOLIDAYS` or Strata breaks.

The only option at the moment is to manually extract the `NO_HOLIDAYS` calendar from the standard reference data and insert it into the user reference data. This is effectively exposing some internal implementation details and seems ugly.

It would be better if the factory method on `ReferenceData` included the mandatory holiday calendar(s) without the users having to do anything.

jodastephen:
Fixes #1291

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/ReferenceData.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/ReferenceData.java
@@ -5,6 +5,7 @@
  */
 package com.opengamma.strata.basics;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/ReferenceData.java
```diff
@@ -47,7 +48,22 @@ public static ReferenceData of(Map<? extends ReferenceDataId<?>, ?> values) {
...
+   * <p>
+   * The {@linkplain #standard() standard} reference data contains common holiday calendars
+   * and indices, but may not be suitable for production use. The minimal reference data contains
+   * just those identifiers that are needed by Strata, and that are non-controversial.
+   * These are {@link HolidayCalendars#NO_HOLIDAYS}, {@link HolidayCalendars#SAT_SUN},
+   * {@link HolidayCalendars#FRI_SAT} and {@link HolidayCalendars#THU_FRI}.
+   *
+   * @return minimal reference data
+   */
+  public static ReferenceData minimal() {
```
cjkent: This PR only solves half the problem when creating reference data which doesn't use the default holidays. In order to create reference data the code will look something like this:

    ReferenceData userRefData = ReferenceData.of(...);
    ReferenceData refData = userRefData.combinedWith(ReferenceData.minimal());

That's not a particularly obvious API. Why not simply include `NO_HOLIDAYS` in the reference data created with `ReferenceData.of()`? Strata doesn't work without `NO_HOLIDAYS` so I can't see a reason for wanting to create a set of reference data that doesn't include it.

jodastephen: I didn't really want to do this as it is a bit weird behaviour, but since the system doesn't work without them, accepting this nastiness will probably avoid some user bugs.

cjkent: I agree it's not ideal, but it seems like the least bad of the two choices.

PR #1313, opened 2016-08-25
Fix MultiCurrencyAmountArray.size() when built from empty amounts

### LINKED ISSUE #1312: MultiCurrencyAmountArray.of(size, fn) doesn't work with empty MultiCurrencyAmounts
The following code produces a `MultiCurrencyAmountArray` of size zero even though it is created from two `MultiCurrencyAmounts`

    MultiCurrencyAmountArray array = MultiCurrencyAmountArray.of(
        MultiCurrencyAmount.empty(),
        MultiCurrencyAmount.empty());

The following code always produces a `MultiCurrencyAmountArray` with a size of zero regardless of the requested size:

    MultiCurrencyAmountArray.of(size, () -> MultiCurrencyAmount.empty());

The size of the array is derived from the size of the internal arrays holding the currency values. If there are no values a size of zero is incorrectly returned.

cjkent:
`MultiCurrencyAmountArray.size()` uses the size of the amounts for one of its currencies. This means it incorrectly reports a size of zero if it is built from multiple empty `MultiCurrencyAmount` instances or if the function passed to the factory method returns empty amounts.

This PR fixes the logic to retain the size even if the amounts are all empty.

Fixes #1312

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
@@ -409,7 +409,7 @@ public final class MultiCurrencyAmountArray
 
   //-----------------------------------------------------------------------
   /**
-   * Gets this size of the array.
+   * Gets the size of this array.
    * @return the value of the property
    */
   public int getSize() {
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
@@ -56,7 +56,7 @@ public final class MultiCurrencyAmountArray
     implements FxConvertible<CurrencyAmountArray>, ImmutableBean, Serializable {
 
   /**
-   * This size of the array.
+   * The size of this array.
    */
   @PropertyDefinition(validate = "notNegative")
   private final int size;
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
@@ -404,7 +404,7 @@ public final class MultiCurrencyAmountArray
 
   //-----------------------------------------------------------------------
   /**
-   * Gets the number of values for each currency.
+   * Gets this size of the array.
    * @return the value of the property
    */
   public int getSize() {
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
@@ -56,14 +56,16 @@ public final class MultiCurrencyAmountArray
     implements FxConvertible<CurrencyAmountArray>, ImmutableBean, Serializable {
 
   /**
+   * This size of the array.
+   */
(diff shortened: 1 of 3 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
```diff
@@ -56,14 +56,16 @@
     implements FxConvertible<CurrencyAmountArray>, ImmutableBean, Serializable {
 
   /**
+   * This size of the array.
```
jonathansenior: "This" and "the" seem to be transposed.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
```diff
@@ -56,14 +56,16 @@
     implements FxConvertible<CurrencyAmountArray>, ImmutableBean, Serializable {
 
   /**
+   * The size of this array.
```
jonathansenior: What are the semantics of `size` supposed to be in this class? It looks like the previous semantics were that `size` was the number of currencies, i.e. consistent with `size` on `MultiCurrencyAmount`. Now it seems to be either this, or the number of `MultiCurrencyAmount` items (i.e. scenarios) that it was constructed from, depending on how it is constructed.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
```diff
@@ -129,7 +131,7 @@ public static MultiCurrencyAmountArray of(int size, IntFunction<MultiCurrencyAmo
    */
   public static MultiCurrencyAmountArray of(Map<Currency, DoubleArray> values) {
     values.values().stream().reduce((a1, a2) -> checkSize(a1, a2));
-    return new MultiCurrencyAmountArray(values);
+    return new MultiCurrencyAmountArray(values.size(), values);
```
jonathansenior: Here, `size` is the number of currencies.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
```diff
@@ -93,7 +95,7 @@ public static MultiCurrencyAmountArray of(List<MultiCurrencyAmount> amounts) {
       }
     }
     Map<Currency, DoubleArray> doubleArrayMap = MapStream.of(valueMap).mapValues(v -> DoubleArray.ofUnsafe(v)).toMap();
-    return new MultiCurrencyAmountArray(doubleArrayMap);
+    return new MultiCurrencyAmountArray(size, doubleArrayMap);
```
jonathansenior: Here, `size` is the number of `MultiCurrencyAmount` items in the input list.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
cjkent: This is wrong, it's supposed to be the number of currency amounts. This should be the size of one of the values in `values`.

modules/basics/src/main/java/com/opengamma/strata/basics/currency/MultiCurrencyAmountArray.java
jonathansenior: Actually I'm wrong here about the semantics - it used to be the size of the array which is in practice the number of scenarios. After your updates this is now the case everywhere, so I will merge.

PR #1318, opened 2016-08-31
Cross Curve Gamma

yukiiwashita:
* Cross curve gamma is computed by finite difference method on curve delta. 
* Gamma for a curve is not computed if the base curve delta does not contain a sensitivity to that curve.


modules/pricer/src/main/java/com/opengamma/strata/pricer/sensitivity/MultiCurveCrossGammaCalculator.java
```diff
@@ -175,4 +233,57 @@ public DoubleArray apply(DoubleArray t) {
...
+    return CrossGammaParameterSensitivity.of(
+        baseDeltaSingle.getMarketDataName(),
+        baseDeltaSingle.getParameterMetadata(),
+        nodalCurve.getName(),
+        metadata,
+        baseDeltaSingle.getCurrency(),
+        sensi);
+  }
+
+  private CrossGammaParameterSensitivity combineSnesitivities(
```
marc-henrard: Spelling of method name.

PR #1347, opened 2016-09-27
Remove TolerantResult

### LINKED ISSUE #1346: Remove TolerantResult
No-one likes the name and there is no good alternative.
Add `FailureItems` to wrap a `List<FailureItem>`.
End users work with `Pair<List<MyClass>, FailureItems>`.

NOTE: This class was never in a supported release. It was added after v1.0 and removed before v1.1.

jodastephen:
Add FailureItems
Fixes #1346


modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItemsBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+   * @param failure  the failure to add
+   * @return this, for chaining
+   */
+  public FailureItemsBuilder addFailure(FailureItem failure) {
+    listBuilder.add(failure);
+    return this;
+  }
+
+  /**
+   * Adds a failure to the list.
```
cjkent: Should be something like "Adds multiple failures to the list."

PR #1352, opened 2016-09-28
Extend FpML root element logic

### LINKED ISSUE #1351: Support FPML documents having clearingStatus as root node
Currently, the Strata FpmlDocumentParser expects the root node to include both `trade` and `party` elements.

I'm working with an FPML document having a structure similar to the one below, and Strata fails when trying to parse it:
```
<clearingStatus ...>
  <clearingStatusItem>
    <trade>
      ...
    </trade>
  </clearingStatusItem>
  <party>
    ...
  </party>
</clearingStatus>
```

jodastephen:
Allow party and trade elements to be separated
Fixes #1351

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
@@ -7,6 +7,7 @@ package com.opengamma.strata.loader.fpml;
 
 import static com.opengamma.strata.collect.Guavate.toImmutableSet;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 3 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
```diff
@@ -195,29 +196,60 @@ private FpmlDocumentParser(
 
   // intelligently finds the FpML root element
   private static XmlElement findFpmlRoot(XmlElement root) {
-    if (isFpmlRoot(root)) {
-      return root;
+    XmlElement fpmlRoot = isFpmlRoot(root);
```
andreiruse: Since `isFpmlRoot()` doesn't return a boolean anymore, is it worth renaming it to something more appropriate, such as `getFpmlRoot()`?

PR #1361, opened 2016-09-30
Functions must receive filtered set of measures

### LINKED ISSUE #1336: CalculationFunction should only be asked for measures it declares it supports
The method `CalculationFunction.supportedMeasures` returns the set of measures the function is able to calculate. If the user requests unsupported measures they are passed to the `requirements` and `execute` method and the function must handle them and return a failure `Result` containing `UNSUPPORTED`.

It would simplify every function if the set of measures passed into the function only included measures that the function claims to support. The `Result` instances for unsupported measures could be created by `CalculationTask`.

jodastephen:
A function should only receive a request for measures it supports
This does not apply to requirements()
Fixes #1336

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunction.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunction.java
@@ -103,6 +103,8 @@ public interface CalculationFunction<T extends CalculationTarget> {
    * Determines the market data required by this function to perform its calculations.
    * <p>
    * Any market data needed by the {@code calculate} method should be specified.
+   * <p>
+   * The set of measures may include measures that are not supported by this function.
    *
    * @param target  the target of the calculation
    * @param measures  the set of measures to be calculated
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/runner/DerivedCalculationFunctionWrapper.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/runner/DerivedCalculationFunctionWrapper.java
@@ -15,6 +15,7 @@ import java.util.Set;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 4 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/runner/DerivedCalculationFunctionWrapper.java
```diff
@@ -122,11 +123,13 @@ public FunctionRequirements requirements(
...
 
     // The caller didn't ask for the derived measure so just return the measures calculated by the delegate
-    if (!measures.contains(derivedFunction.measure())) {
+    Measure derivedMeasure = derivedFunction.measure();
+    if (!measures.contains(derivedMeasure)) {
       return delegate.calculate(target, measures, parameters, marketData, refData);
     }
     // Add the measures required to calculate the derived measure to the measures requested by the caller
-    Set<Measure> requiredMeasures = Sets.union(measures, derivedFunction.requiredMeasures());
+    SetView<Measure> allRequiredMeasures = Sets.union(measures, derivedFunction.requiredMeasures());
```
cjkent: Did you mean to use `SetView` as the type here? That happens to me when I let the IDE choose the type.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationFunction.java
```diff
@@ -103,6 +103,8 @@
    * Determines the market data required by this function to perform its calculations.
    * <p>
    * Any market data needed by the {@code calculate} method should be specified.
+   * <p>
+   * The set of measures may include measures that are not supported by this function.
```
cjkent: Do you not think it's worth filtering the measures passed to the `requirements` method too?

jodastephen: The docs of requirements() didn't specify the measures were limited, whereas the docs to calculate() did. Since doing so has an overhead, and code in requirements() usually doesn't care, it seemed like the best balance.

Issue #1331, opened 2016-09-09
Support bonds in the calculation and measures API


cjkent:
There is pricer support but no function.

PR #1368, opened 2016-10-03
Bond API

jodastephen:
This change renames `LegalEntityDiscountingProvider` to `ImmutableLegalEntityDiscountingProvider` and adds a new interface called `LegalEntityDiscountingProvider`. It also renames `BondGroup` to `RepoGroup` with all associated access methods and properties renamed. Within `ImmutableLegalEntityDiscountingProvider`, `bondMap` is renamed to `repoCurveGroups`, and `legalEntityMap` is renamed to `issuerCurveGroups`. All this is **backwards incompatible**.

The incompatibility was driven by two separate flaws. Firstly, the provider should not have referred to bonds, as it will be used for bills at some point in the future. Secondly, there should always have been a separation between the class and interface to allow for the calculation/measure API to be added.

An attempt was made to add deprecated workarounds, but with two interlinked issues, the deprecations were not practical. Migration should not be complex, simply rename each item as listed above.

See #1331


modules/pricer/src/main/java/com/opengamma/strata/pricer/bond/ImmutableLegalEntityDiscountingProvider.java
```diff
@@ -0,0 +1,758 @@
...
+        throw new IllegalArgumentException("Invalid valuation date for the issuer curve: " + entry.getValue());
+      }
+      if (!issuerCurveGroups.containsValue(entry.getKey().getFirst())) {
+        throw new IllegalArgumentException("No map to the legal entity group from ID: " + entry.getKey().getFirst());
+      }
+    }
+  }
+
+  //-------------------------------------------------------------------------
+  /**
```
yukiiwashita: Remove this as the same comment is in the interface.

PR #1389, opened 2016-10-13
Allow CSV to use Reader

### LINKED ISSUE #1384: Allow CsvIterator and CsvFile to be created from a Reader
They can only be created from a `CharSource` which means they are not a simple drop-in replacement for other CSV libraries which use `Reader` (for example Commons CSV and Opencsv).

jodastephen:
Fixes #1384

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
@@ -7,6 +7,7 @@ package com.opengamma.strata.collect.io;
 
 import static com.opengamma.strata.collect.Guavate.toImmutableList;
 
+import java.io.Reader;
 import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -17,6 +18,7 @@ import java.util.Map;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.CharSource;
+import com.google.common.io.CharStreams;
 import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.collect.Unchecked;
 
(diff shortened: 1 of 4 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
```diff
@@ -88,7 +90,34 @@ public static CsvFile of(CharSource source, boolean headerRow) {
...
-    ImmutableList<String> lines = Unchecked.wrap(() -> source.readLines());
+    List<String> lines = Unchecked.wrap(() -> source.readLines());
+    return create(lines, headerRow, separator);
+  }
+
+  /**
+   * Parses the specified reader as a CSV file where the separator is specified and might not be a comma.
+   * <p>
+   * This factory method takes a {@link Reader}.
+   * Callers are encouraged to use {@link CharSource} instead of {@code Reader} where possible.
```
cjkent: Is there a reason to recommend `CharSource` instead of `Reader`? It's a puzzling recommendation with no explanation. `Reader` is a JDK class familiar to pretty much any Java dev.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvIterator.java
```diff
@@ -114,28 +115,58 @@ public static CsvIterator of(CharSource source, boolean headerRow, char separato
...
+   * The caller is responsible for closing the reader, such as by calling {@link #close()}.
+   * 
+   * @param reader  the file reader
+   * @param headerRow  whether the source has a header row, an empty source must still contain the header
+   * @param separator  the separator used to separate each field, typically a comma, but a tab is sometimes used
+   * @return the TSV file
+   * @throws UncheckedIOException if an IO exception occurs
+   * @throws IllegalArgumentException if the file cannot be parsed
+   */
+  public static CsvIterator of(Reader reader, boolean headerRow, char separator) {
```
cjkent: It would be better to have a `Reader` version of both factory methods. Assuming most files use commas as a separator it is a bit annoying that you need to specify the separator every time if you're using a reader but not if you are using a char source. It also seems surprisingly asymmetric.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvFile.java
```diff
@@ -88,7 +90,34 @@ public static CsvFile of(CharSource source, boolean headerRow) {
...
+   * For example, a tab-separated file is very similar to a CSV file, the only difference is the separator.
+   * 
+   * @param reader  the file resource
+   * @param headerRow  whether the source has a header row, an empty source must still contain the header
+   * @param separator  the separator used to separate each field, typically a comma, but a tab is sometimes used
+   * @return the TSV file
+   * @throws UncheckedIOException if an IO exception occurs
+   * @throws IllegalArgumentException if the file cannot be parsed
+   */
+  public static CsvFile of(Reader reader, boolean headerRow, char separator) {
```
cjkent: Same as the comment on `CsvIterator` - would be good to have a `Reader` version of both factory methods for better usability and symmetry.

PR #1390, opened 2016-10-17
Added GBP/EUR cross-currency convention.

marc-henrard:


modules/product/src/main/java/com/opengamma/strata/product/swap/type/StandardXCcyIborIborSwapConventions.java
```diff
@@ -66,6 +67,26 @@
           .spotDateOffset(DaysAdjustment.ofBusinessDays(2, GBLO_USNY))
           .build();
 
+  /**
+   * GBP LIBOR 3M v EUR EURIBOR 3M.
+   * The spread is on the GBP leg.
+   */
+  public static final XCcyIborIborSwapConvention GBP_LIBOR_3M_EUR_EURIBOR_3M =
```
jodastephen: Since EUR/GBP is the correct way around for the currency pair, I suspect this should be named `EUR_EURIBOR_3M _GBP_LIBOR_3M`. Unless we are naming them to indicate the order of the spread leg? (I'm happy with either convention, but we should document the naming convention we are using in a comment in the class)

marc-henrard: I had in mind to keep the leg with the market quote as the first one. This is what we have done for all others (FIXED v LIBOR, XCCY, IBOR+spread v IBOR). I will add some comments in that direction in the java doc.

PR #1395, opened 2016-10-19
Allow Strata to be on the classpath more than once

### LINKED ISSUE #1394: Allow Strata to be on the classpath more than once
The classpath should contain only one copy of Strata. If it contains two or more, the system fails due to finding more than one copy of the extended enum files.

Change the logic to allow more than one copy of an enum file so long as the content is the same.

jodastephen:
Only allow this if the data files resolve to the same content
Fixes #1394

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/ResourceConfig.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/ResourceConfig.java
@@ -255,8 +255,18 @@ public final class ResourceConfig {
           result.add(ResourceLocator.ofClasspathUrl(urls.get(0)));
           break;
         default:
-          log.severe("More than one file found on the classpath: " + name + ": " + urls);
-          throw new IllegalStateException("More than one file found on the classpath: " + name + ": " + urls);
+          // handle case where Strata is on the classpath more than once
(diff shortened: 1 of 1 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/ResourceConfig.java
```diff
@@ -255,8 +255,18 @@ public static IniFile combinedIniFile(List<ResourceLocator> resources) {
           result.add(ResourceLocator.ofClasspathUrl(urls.get(0)));
           break;
         default:
-          log.severe("More than one file found on the classpath: " + name + ": " + urls);
-          throw new IllegalStateException("More than one file found on the classpath: " + name + ": " + urls);
+          // handle case where Strata is on the classpath more than once
+          // only accept this if the data being read is the same in all URLs
+          ResourceLocator baseResource = ResourceLocator.ofClasspathUrl(urls.get(0));
+          for (int i = 1; i < urls.size(); i++) {
+            ResourceLocator otherResource = ResourceLocator.ofClasspathUrl(urls.get(1));
```
cjkent: I think this should be `urls.get(i)`

PR #1397, opened 2016-10-20
Add ValueWithFailures

### LINKED ISSUE #1396: Add ValueWithFailures
Can be used instead of `Pair<T, FailureItems>`

jodastephen:
Fixes #1396


modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -0,0 +1,398 @@
...
+/**
+ * A value with associated failures.
+ * <p>
+ * This captures a common use case where an operation can tolerate some failure.
+ * This is often referred to as partial success or partial failure.
+ * The class stores the value, of any object type, and a list of failures that may be empty.
+ * <p>
+ * The success result must be able to handle the case where everything fails.
+ * In most cases, the success result will be a collection type, such as {@link List}
+ * or {@link Map}, which can be empty if operation failed completely.
```
jonathansenior: if _the_ operation?

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -0,0 +1,398 @@
...
+    return new ValueWithFailures<>(successValue, ImmutableList.copyOf(failures));
+  }
+
+  /**
+   * Creates an instance wrapping the success value and failures.
+   *
+   * @param <T>  the type of the success value
+   * @param successValue  the success value
+   * @param failures  the failures
+   * @return an instance result wrapping the value and failures
```
jonathansenior: Additional word "result" in this description vs the method above

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -0,0 +1,398 @@
...
+ */
+@BeanDefinition(builderScope = "private")
+public final class ValueWithFailures<T>
+    implements ImmutableBean, Serializable {
+
+  /**
+   * The success value.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final T value;
```
jonathansenior: This is referred to interchangeably as "result" and "value" - should probably pick "value" given the name of the class and this field.

PR #1399, opened 2016-10-25
Created DiscountFactorsDecoratedForward. 

marc-henrard:
This is the first part of a multi-leg development. The final goal is to create tools to compute generic "theta" values, i.e. impact of time, for interest rate books.
This require to be able to reprice books with curves at a later date (usually T+1). Their are several ways to create the curves at a later date; this implementation uses a forward approach.
Within that approach we need to move the discount factors (this PR), ibor values, overnight values, and FX rates (forthcoming PRs) at the future date.
While working on discount factors, I moved some methods implemented by copy and past at the class level to the interface level as default implementation.


marc-henrard:
PR moved to Granite.

PR #1403, opened 2016-10-31
Enhance schedule generation

jodastephen:
Allow non-EOM roll conventions to behave like EOM ones based on pre-adjusted dates.
See [FpML clarification](http://www.fpml.org/forums/topic/can-a-roll-convention-imply-a-stub/#post-7659).


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/StandardRollConventions.java
```diff
@@ -33,6 +33,11 @@ public LocalDate adjust(LocalDate date) {
       ArgChecker.notNull(date, "date");
       return date.withDayOfMonth(date.lengthOfMonth());
     }
+
+    @Override
+    public int getDayOfMonth() {
+      return 31;
```
cjkent: This seems slightly odd. I'm sure it works, but is it worth a comment explaining what's going on?

PR #1424, opened 2016-12-01
Allow base sensitivity to be split

### LINKED ISSUE #1423: Combine and Split sensitivity data
`UnitParameterSensitivity` and `CurrencyParameterSensitivity` allow sensitivity to be stored for a single market data name (such as a curve). In certain cases, it is desirable to be able to create a curve from more than one underlying curve, hiding the underlying curves through some or all of the calculation. Proposed solution is to allow the singular sensitivity classes to be created as a combination of other sensitivity classes, with the ability to split it back out to the fundamental curves at a later point.

jodastephen:
Add combine/split concept to base sensitivity classes
Fixes #1423


modules/market/src/main/java/com/opengamma/strata/market/param/ParameterSize.java
```diff
@@ -0,0 +1,345 @@
...
+import com.opengamma.strata.data.MarketDataName;
+
+/**
+ * The market data name and the associated number of parameters.
+ * <p>
+ * This holds the name and the number of parameters that define it.
+ * For example, the curve name and the number of curve nodes.
+ */
+@BeanDefinition(builderScope = "private")
+public final class ParameterSize
```
marc-henrard: This is very similar to the object 'CurveParameterSize'. Do we want to keep both?

modules/market/src/main/java/com/opengamma/strata/market/param/CurrencyParameterSensitivity.java
```diff
@@ -77,6 +88,16 @@
    */
   @PropertyDefinition(validate = "notNull")
   private final DoubleArray sensitivity;
+  /**
+   * The split of parameters between the underlying parameterized data.
+   * <p>
+   * A single {@code UnitParameterSensitivity} represents the sensitivity to a single {@link ParameterizedData} instance.
```
marc-henrard: CurrencyParameterSensitivity instead of  UnitParameterSensitivity

modules/market/src/main/java/com/opengamma/strata/market/param/CurrencyParameterSensitivity.java
```diff
@@ -346,6 +505,20 @@ public DoubleArray getSensitivity() {
   }
 
   //-----------------------------------------------------------------------
+  /**
+   * Gets the split of parameters between the underlying parameterized data.
+   * <p>
+   * A single {@code UnitParameterSensitivity} represents the sensitivity to a single {@link ParameterizedData} instance.
```
marc-henrard: As above

jodastephen:
We'll keep `CurveParameterSize` separate from `ParameterSize` as the curve version is used on existing APIs and there is no great need to change.

PR #1434, opened 2016-12-14
Added first order derivative for discount factors with respect to time.

marc-henrard:
DiscountFactors currently returns the discount factor for a given year fraction/time. The method is a function from double to double; it is natural in some circumstances to compute its derivative.
This PR add the time derivative in the interface methods and implement it for the 3 current DiscountFactors implementations.
The same method need to be implemented in other modules which implement this interface.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/DiscountFactors.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/DiscountFactors.java
@@ -120,6 +120,17 @@ public interface DiscountFactors
    * @throws RuntimeException if the value cannot be obtained
    */
   public abstract double discountFactor(double yearFraction);
+  
+  /**
+   * Returns the discount factor derivative with respect to the year fraction or time.
+   * <p>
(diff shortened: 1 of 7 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/DiscountFactors.java
```diff
@@ -120,6 +120,17 @@ public default double discountFactor(LocalDate date) {
...
+  /**
+   * Returns the discount factor derivative with respect to the year fraction or time.
+   * <p>
+   * The year fraction must be based on {@code #relativeYearFraction(LocalDate)}.
+   * 
+   * @param yearFraction  the year fraction 
+   * @return the discount factor derivative
+   * @throws RuntimeException if the value cannot be obtained
+   */
+  public abstract double discountFactorTimeDerivative(double yearFraction);
```
jodastephen: This is a backwards incompatible change. It is probably the right thing to do, but will need noting in release notes.

PR #1436, opened 2016-12-14
Extended enum case insensitive lookup

jodastephen:
Allow extended enums to be looked up by upper case name
as well as by correct name

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/StandardReferenceData.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/StandardReferenceData.java
@@ -5,6 +5,9 @@
  */
 package com.opengamma.strata.basics;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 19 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/StandardReferenceData.java
```diff
@@ -19,11 +22,11 @@
    */
   static final ImmutableReferenceData STANDARD;
   static {
-    ImmutableMap.Builder<ReferenceDataId<?>, Object> builder = ImmutableMap.builder();
+    Map<ReferenceDataId<?>, Object> map = new HashMap<>();
```
andreiruse: Not related to the changes, why not `Map<ReferenceDataId<?>, HolidayCalendar>` ?

Also, same question for the next static{} block below

jodastephen: Reference data can hold many different things, not just holidays,

PR #1444, opened 2016-12-19
Index curve calibration

yukiiwashita:
* Index curve calibrator calling the single name CDS calibrator. 
* Index node.
* Index trade object for calibration.
* Legal entity information.


modules/market/src/main/java/com/opengamma/strata/market/curve/node/CdsIndexIsdaCreditCurveNode.java
```diff
@@ -0,0 +1,904 @@
...
+  @PropertyDefinition(validate = "notNull")
+  private final ImmutableList<StandardId> legalEntityIds;
+  /**
+   * The market quote convention.
+   * <p>
+   * The CDS index is quoted in par spread, points upfront or quoted spread.
+   * See {@link CdsQuoteConvention} for detail.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final CdsQuoteConvention quoteConvention;
```
jodastephen: Should there be a `CdsIndexTemplate` to capture the legal entity IDs, quote convention etc. Lots of code in the `trade()` method below.

PR #1454, opened 2016-12-21
Create a public CalculationListener returning a future

cjkent:
This PR renames `AggregatingListener` and moves it to the top level so it can be used by external code.

Having a `CalculationListener` returning a `CompletableFuture` is essential for writing fully asynchronous code with Strata. It seems a shame that users will have to implement it themselves when we already have one in Strata.


jodastephen:
Approved, as there is nothing wrong with this. However, the new public class does seem a little isolated. Another problem is that the user has is they need to pass in the columns to the `ResultsListener` (and typically other Strata classes have factory methods). Not sure there is a way to setup the columns automatically, but it would be nice. Maybe a `calculationStarted(columns)` method on `CalculationListener` ?

cjkent:
I agree that it's not particularly elegant to pass the columns to the listener and the calculation runner. I like your suggestion. I'll merge this PR and raise a follow up to add `calculationStarted()`

PR #1455, opened 2016-12-21
Add CalculationListener.calculationsStarted

cjkent:
Adds a method to `CalculationListener` that is invoked before any results are calculated. This allows the listener to know about the columns without the user having to pass the columns to both the listener and the calculation runner.

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationListener.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationListener.java
@@ -5,21 +5,42 @@
  */
 package com.opengamma.strata.calc.runner;
 
+import java.util.List;
+
 import com.opengamma.strata.basics.CalculationTarget;
 import com.opengamma.strata.calc.CalculationRunner;
+import com.opengamma.strata.calc.Column;
 
 /**
  * Listener that is notified when calculations are performed by a {@link CalculationRunner}.
  * <p>
  * It is guaranteed that the methods of a listener will only be invoked by a single thread at any
- * time. Therefore listener implementations are not necessarily required to be thread safe.
+ * time. It is not guaranteed to be the same thread invoking a listener each time. The calling
+ * code is synchronized to ensure that any changes in the listener state will be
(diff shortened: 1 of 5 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/runner/ResultsListener.java
```diff
@@ -32,16 +32,16 @@
   private final List<CalculationResult> results = new ArrayList<>();
 
   /** The columns that define what values are calculated. */
-  private final List<Column> columns;
+  private List<Column> columns;
```
jodastephen: volatile? Since the listener may be called from different threads.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationListener.java
```diff
@@ -5,21 +5,42 @@
...
 
   /**
-   * Invoked when a calculation completes.
+   * Invoked when the calculations start; guaranteed to be invoked
+   * before {@link #resultReceived(CalculationTarget, CalculationResult)} and
+   * {@link #calculationsComplete()}.
+   *
+   * @param columns the columns for which values are being calculated
+   */
+  public default void calculationsStarted(List<Column> columns) {
```
jodastephen: Take list of targets as well as columns?

modules/calc/src/main/java/com/opengamma/strata/calc/runner/ResultsListener.java
cjkent: The calling code goes to great lengths to ensure everything is synchronised so all changes are visible to all threads. The Javadoc on `CalculationListener` specifies that implementations don't have to be thread safe.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationListener.java
cjkent: Seems reasonable. Our implementations won't use it but there's no harm adding it and I can see why it might be useful. It's also available at the call site so it's easy to add.

PR #1459, opened 2016-12-29
Added support for flexible line item separator in CsvOutput, Fixes #1439

### LINKED ISSUE #1439: CsvOutput: extend the tool to allow flexible separation character
CsvOutput creates CSV files with a fixed item separation character ','. 
A flexible line separator is possible with the current tool, but not a flexible item separator. This is required by some formats that uses a "tab separated" style.

davidcervi:


modules/collect/src/test/java/com/opengamma/strata/collect/io/CsvOutputTest.java
```diff
@@ -19,39 +19,40 @@
 public class CsvOutputTest {
 
   private static final String LINE_SEP = System.lineSeparator();
+  private static final String LINE_ITEM_SEP = ",";
```
jonathansenior: We should test something other than ',' as well.

modules/collect/src/test/java/com/opengamma/strata/collect/io/CsvOutputTest.java
```diff
@@ -19,40 +19,49 @@
 public class CsvOutputTest {
 
   private static final String LINE_SEP = System.lineSeparator();
+  private static final String LINE_ITEM_SEP_COMMA = ",";
+  private static final String LINE_ITEM_SEP_TAB = " ";
```
jonathansenior: Looks like a space. You'll need `\t` for a tab.

PR #1469, opened 2017-01-12
Add methods to CsvRow to handle empty fields, Fixes #1228

### LINKED ISSUE #1228: Add methods to CsvRow to handle empty fields
The `getField` and `findField` methods on `CsvRow` return an empty string if there is no value for a field. This is such a common case that it would be helpful if there were methods which treated an empty value as missing:

```
String getValue(String fieldName);
Optional<String> findValue(String fieldName);
```

davidcervi:


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvRow.java
```diff
@@ -179,6 +179,80 @@ public String getField(Pattern headerPattern) {
...
+   * If the header is not found or the value found is an empty string, then an IllegalArgumentException is thrown.
+
+   * @param headerPattern the header pattern to match
+   * @return the trimmed field value
+   * @throws IllegalArgumentException if the header is not found or if the value in the field is empty.
+   */
+  public String getValue(Pattern headerPattern) {
+    String value = getField(headerPattern);
+    if (value.isEmpty()) {
+      throw new IllegalArgumentException("No value was provided for header pattern" + headerPattern);
```
cjkent: I think "No value was found..." would sound better. When I see "provided" it makes me think of a value that I should have provided as the caller.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvRow.java
```diff
@@ -179,6 +179,80 @@ public String getField(Pattern headerPattern) {
...
+   * If the header is not found or the value found is an empty string, then an IllegalArgumentException is thrown.
+   *
+   * @param header the column header
+   * @return the trimmed field value, empty
+   * @throws IllegalArgumentException if the header is not found or if the value in the field is empty.
+   */
+  public String getValue(String header) {
+    String value = getField(header);
+    if (value.isEmpty()) {
+      throw new IllegalArgumentException("No value was provided for field " + header);
```
cjkent: Same as above

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvRow.java
```diff
@@ -179,6 +179,80 @@ public String getField(Pattern headerPattern) {
     return Optional.empty();
   }
 
+  /**
+   * Gets a single field value from the row by header pattern
+   * <p>
+   * This returns the value of the first column where the header matches the specified header pattern.
+   * If the header is not found or the value found is an empty string, then an IllegalArgumentException is thrown.
+
```
cjkent: Missing asterisk in the Javadoc.

PR #1487, opened 2017-03-01
Better document the tick concepts

jodastephen:


joburnham:
For example, the price of an ICE Brent Crude future is based on the price of a barrel of crude oil in USD
The tick size of this contract is 0.01 (equivalent to 1 cent)
The contract size is 1,000 barrels
Therefore the tick value is 0.01 * 1,000 = 10 USD

PR #1492, opened 2017-03-06
Add standard ETD identifiers

jodastephen:
Include support for Flex futures/options


modules/product/src/main/java/com/opengamma/strata/product/etd/EtdIdUtils.java
```diff
@@ -0,0 +1,160 @@
...
+   * @param strikePrice  the strike price
+   * @return the identifier
+   */
+  public static SecurityId optionId(
+      ExchangeId exchangeId,
+      String contractCode,
+      YearMonth expiryMonth,
+      EtdStyle style,
+      int version,
+      PutCall putCall,
```
cjkent: There is no null check for this. If it's null the method will silently return an ID for a call option which probably isn't what we want.

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdOptionSecurity.java
```diff
@@ -207,22 +265,34 @@ public EtdContractSpecId getContractSpecId() {
    * This is used to describe the instance of the option.
    * @return the value of the property, not null
    */
-  public YearMonth getExpiry() {
-    return expiry;
+  public YearMonth getExpiryMonth() {
```
cjkent: I'm not sure this change is an improvement. The value is a year-month so calling it "month" is slightly misleading. When I see `getExpiryMonth` I expect it to return something like `Month.MARCH`, not a `YearMonth`.

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdStyleType.java
```diff
@@ -0,0 +1,63 @@
...
+import org.joda.convert.FromString;
+import org.joda.convert.ToString;
+
+import com.google.common.base.CaseFormat;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * The type of an Exchange Traded Derivative (ETD) product, either a future or an option.
+ */
+public enum EtdStyleType {
```
cjkent: This is an awkward name. `EtdContractType`?

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdIdUtils.java
```diff
@@ -0,0 +1,160 @@
...
+public final class EtdIdUtils {
+
+  /**
+   * Scheme used for ETDs.
+   */
+  public static final String ETD_SCHEME = "OG-ETD";
+  /**
+   * Prefix for futures.
+   */
+  private static final String FUT_PREFIX = "F/";
```
jodastephen: Since strike can be negative, a dash is not a good option. Any other option? I think the slash reads nicely, but take your point about possible use in a URL

jodastephen: See what you think of colons

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdStyleType.java
jodastephen: It is the type of the `EtdStyle` class

cjkent: But it's a name that doesn't have any real-world meaning. `EtdStyleType` means nothing to someone with domain knowledge. `EtdContractType` is a pretty meaningful description of what it is. 

I thought of `EtdContractType` after looking at the Javadoc for each of the fields and seeing they call contain "contract".

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdIdUtils.java
cjkent: Colons are definitely an improvement but they still have to be URL encoded. That was why `~` was chosen originally. "OG-ETD~O:ECAG:OGBS" is encoded to "OG-ETD~O%3AECAG%3AOGBS". But "OG-ETD~O-ECAG-OGBS" doesn't need any encoding.

What do you think?

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdStyleType.java
jodastephen: "contract type" would be taken to refer to future vs option. Jo is happy enough with `EtdStyle` as a name.

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdStyleType.java
```diff
@@ -0,0 +1,63 @@
...
+package com.opengamma.strata.product.etd;
+
+import org.joda.convert.FromString;
+import org.joda.convert.ToString;
+
+import com.google.common.base.CaseFormat;
+import com.opengamma.strata.collect.ArgChecker;
+
+/**
+ * The type of an Exchange Traded Derivative (ETD) product, either a future or an option.
```
cjkent: I've just spotted this - copy/paste error

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdStyleType.java
cjkent: I can live with `EtdStyle`. But `StyleType` doesn't make any sense.

cjkent: @joburnham, @phuvan and I have been discussing this and have a couple of alternatives that avoid `StyleType`

* Rename `EtdStyle` to `EtdInfo` and `EtdStyleType` to `EtdStyle`
* Rename `EtdStyleType` to `EtdVariant`

jodastephen:
Reopening relative to develop.

PR #1516, opened 2017-05-09
Legal entity rates loader

yukiiwashita:
* Implemented **LegalEntityRatesCurvesCsvLoader**. 
* Reuse the existing csv formats except the curve group. An example is `legal-entity-groups.csv`.
* Resulting object is **LegalEntityCurveGroup** from which the rates provider is created via **LegalEntityDiscountingMarketDataLookup**.
* Move **RepoGroup** and **LegalEntityGroup**.


modules/market/src/test/java/com/opengamma/strata/market/curve/RepoGroupTest.java
```diff
@@ -3,12 +3,14 @@
...
  * Please see distribution for license.
  */
-package com.opengamma.strata.pricer.bond;
+package com.opengamma.strata.market.curve;
 
 import static org.testng.Assert.assertEquals;
 
 import org.testng.annotations.Test;
 
+import com.opengamma.strata.market.curve.RepoGroup;
```
jodastephen: Test needs to move package too

modules/market/src/test/java/com/opengamma/strata/market/curve/LegalEntityGroupTest.java
```diff
@@ -3,12 +3,14 @@
  *
  * Please see distribution for license.
  */
-package com.opengamma.strata.pricer.bond;
+package com.opengamma.strata.market.curve;
```
jodastephen: Test needs to move package too

modules/measure/src/main/java/com/opengamma/strata/measure/StandardComponents.java
```diff
@@ -75,10 +77,12 @@
    * The standard calculation functions.
    */
   private static final CalculationFunctions STANDARD = CalculationFunctions.of(
+      new BondFutureTradeCalculationFunction(),
```
jodastephen: Javadoc also needs to be updated within the class.

modules/market/src/main/java/com/opengamma/strata/market/curve/RepoGroup.java
```diff
@@ -3,7 +3,7 @@
  *
  * Please see distribution for license.
  */
-package com.opengamma.strata.pricer.bond;
+package com.opengamma.strata.market.curve;
```
jodastephen: It is rather unfortunate that this class is moved a second time!

modules/loader/src/main/java/com/opengamma/strata/loader/csv/LegalEntityRatesCurvesCsvLoader.java
```diff
@@ -0,0 +1,375 @@
...
+    ImmutableListMultimap.Builder<LocalDate, LegalEntityCurveGroup> builder = ImmutableListMultimap.builder();
+
+    for (Map.Entry<LocalDate, Map<CurveName, Curve>> curveEntry : allCurves.entrySet()) {
+      LocalDate date = curveEntry.getKey();
+      Map<CurveName, Curve> curves = curveEntry.getValue();
+      for (Map.Entry<CurveGroupName, Map<Pair<RepoGroup, Currency>, CurveName>> repoEntry : repoGroups.entrySet()) {
+        CurveGroupName groupName = repoEntry.getKey();
+        Map<Pair<RepoGroup, Currency>, Curve> repoCurves = MapStream.of(repoEntry.getValue())
+            .mapValues(name -> curves.get(name))
+            .peek(e -> {
```
jodastephen: peek() not generally a good thing to use, is there an alternative?

modules/loader/src/main/java/com/opengamma/strata/loader/csv/LegalEntityRatesCurvesCsvLoader.java
```diff
@@ -207,24 +207,28 @@
...
-            .peek(e -> {
-              if (e.getValue() == null) {
-                throw new IllegalArgumentException(
-                    "Repo curve values for " + e.getKey().toString() + " in group " + groupName.getName() +
-                        " are missing on " + date.toString());
-              }
-            })
+            .mapValues(
+                name -> {
+                  Curve curve = curves.get(name);
```
jodastephen: We try to avoid blocks by creating extra methods. So, this would be `name -> queryCurve(name, curves, "Repo")

(this extra method can be used for both of these cases)

PR #1520, opened 2017-06-06
Update HolidayCalendarIds.java

nimo23:
Added: identifier for the holiday calendar of Frankfurt, Germany, with code 'DEFR'.
Added: identifier for the holiday calendar of the Frankfurt Stock Exchange (Xetra), with code 'FSX'.


nimo23:
Added to https://github.com/OpenGamma/Strata/pull/1521

PR #1521, opened 2017-06-06
Update GlobalHolidayCalendars.java

nimo23:
Added: holiday calendar for "Frankfurt, Germany", with code 'DEFR'
Added: holiday calendar for the "Frankfurt Stock Exchange (Xetra)", with code 'FSX'


jodastephen:
Thanks for the PR, sorry it took ages to look at. Ultimately, we decided to stick with keeping Strata tied to the FpML codes for holidays, and so we couldn't take in the exchange calendar. Maybe this will change at some point in the future, but our goal is to provide example calendars rather than all calendars.

PR #1533, opened 2017-07-24
Abstract swap conventions

jodastephen:
All single currency swap conventions are similar, so this creates an abstraction.

Additional `Guavate` method to simulate `Optional.or`.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -71,21 +71,46 @@ public final class Guavate {
   /**
    * Uses a number of suppliers to create a single optional result.
    * <p>
+   * This invokes each supplier in turn until a non empty optional is returned.
+   * As such, not all suppliers are necessarily invoked.
+   * <p>
    * The Java 8 {@link Optional} class does not have an {@code or} method,
    * so this provides an alternative.
    * 
-   * @param <T>  the type of element in the iterable
+   * @param <T>  the type of element in the optional
    * @param suppliers  the suppliers to combine
-   * @return the list that combines the inputs
+   * @return the first non empty optional
    */
   @SuppressWarnings("unchecked")
   @SafeVarargs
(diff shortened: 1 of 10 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -69,6 +69,30 @@ private Guavate() {
 
   //-------------------------------------------------------------------------
   /**
+   * Uses a number of suppliers to create a single optional result.
+   * <p>
+   * The Java 8 {@link Optional} class does not have an {@code or} method,
+   * so this provides an alternative.
+   * 
+   * @param <T>  the type of element in the iterable
```
cjkent: Copy-paste error

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -69,6 +69,30 @@ private Guavate() {
...
+   * @param suppliers  the suppliers to combine
+   * @return the list that combines the inputs
```
cjkent: Copy-paste error

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -69,6 +69,30 @@ private Guavate() {
...
+   */
+  @SuppressWarnings("unchecked")
+  @SafeVarargs
+  public static <T> Optional<T> firstNotEmpty(Supplier<Optional<? extends T>>... suppliers) {
```
cjkent: The equivalent Guava method for null is called `firstNonNull` (note "non" instead of "not"). Would it be better to be consistent with that?

cjkent: This method is useful for the use cases in this PR, but would it be worth adding an overload with the signature:

    Optional<T> firstNotEmpty(Optional<? extends T>... optionals)

That might be more generally useful.

PR #1542, opened 2017-08-03
Allow notional exchange flags to be set independently of FX Reset.

brianweller89:
The initial, intermediate and final notional exchanges are no longer automatically set to true when an FX reset is present. Clients must now explicitly specify if the exchanges should be present.

[FpML](http://www.fpml.org/spec/fpml-5-7-5-rec-1/html/recordkeeping/schemaDocumentation/schemas/fpml-shared-5-7_xsd/complexTypes/PrincipalExchanges.html) supports independent setting of all three flags.

Specifically, XCCY swaps with no initial exchange are a traded instrument and could not be constructed in Strata previously.


modules/product/src/main/java/com/opengamma/strata/product/swap/NotionalSchedule.java
```diff
@@ -72,7 +71,9 @@
    * exchange of money between the counterparties.
    * <p>
    * When building the notional schedule, if an {@code FxResetCalculation} is present,
-   * then the notional exchange flags will be set to true.
+   * then at least one of the notional exchange flags should be set to true.
+   * <p>
+   * If all notional exchange flags are false then setting the fx reset definition has no impact.
```
jodastephen: How about throwing an exception if all are false?

brianweller89: Added

PR #1544, opened 2017-08-10
Swap Calculation Function to support custom logic when resolving a swap trade

brianweller89:
Allows client to inject more complicated logic for converting SwapTrade to ResolvedSwapTrade.

Default behaviour is unchanged.


brianweller89:
No longer required

PR #1549, opened 2017-08-30
Enhance floating rate names

jodastephen:
Add missing `FloatingRateIndex` abstraction
Add default fixed leg day count
Add more flexible parsing methods
Link `FloatingRateName` and `FloatingRateIndex` better


modules/basics/src/main/java/com/opengamma/strata/basics/index/IborIndexCsvLookup.java
```diff
@@ -115,17 +116,18 @@ private IborIndexCsvLookup() {
...
+    boolean active = Boolean.parseBoolean(row.getValue(ACTIVE_FIELD));
+    DayCount dayCount = DayCount.of(row.getValue(DAY_COUNT_FIELD));
+    HolidayCalendarId fixingCal = HolidayCalendarId.of(row.getValue(FIXING_CALENDAR_FIELD));
+    int offsetDays = Integer.parseInt(row.getValue(OFFSET_DAYS_FIELD));
+    HolidayCalendarId offsetCal = HolidayCalendarId.of(row.getValue(OFFSET_CALENDAR_FIELD));
+    HolidayCalendarId effectiveCal = HolidayCalendarId.of(row.getValue(EFFECTIVE_DATE_CALENDAR_FIELD));
+    Tenor tenor = Tenor.parse(row.getValue(TENOR_FIELD));
+    LocalTime time = LocalTime.parse(row.getValue(FIXING_TIME_FIELD), TIME_FORMAT);
+    ZoneId zoneId = ZoneId.of(row.getValue(FIXING_ZONE_FIELD));
+    DayCount fixedLegDayCount = row.findValue(FIXED_LEG_DAY_COUNT).map(s -> DayCount.of(s)).orElse(dayCount);
```
brianweller89: I wonder how safe this defaulting logic is for IBOR? More often than not the DCFS for fixed and float are different.

I don't think it's unreasonable that an explicit fixed DCF must be provided when defining an IBOR index

PR #1550, opened 2017-08-31
Extract common parsing logic to utility class

jodastephen:
May be useful to other parsers.

This is primarily moving existing code and adding Javadoc. The time parser is new.

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
@@ -5,17 +5,70 @@
  */
 package com.opengamma.strata.loader;
 
+import static java.time.temporal.ChronoField.HOUR_OF_DAY;
+import static java.time.temporal.ChronoField.MINUTE_OF_HOUR;
+import static java.time.temporal.ChronoField.NANO_OF_SECOND;
+import static java.time.temporal.ChronoField.SECOND_OF_MINUTE;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
+import java.util.Locale;
+
 import com.opengamma.strata.basics.index.FxIndex;
 import com.opengamma.strata.basics.index.IborIndex;
 import com.opengamma.strata.basics.index.Index;
 import com.opengamma.strata.basics.index.OvernightIndex;
 import com.opengamma.strata.basics.index.PriceIndex;
 import com.opengamma.strata.collect.Messages;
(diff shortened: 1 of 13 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -46,6 +99,178 @@ public static Index findIndex(String reference) {
...
+      case "YES":
+      case "Y":
+        return true;
+      case "FALSE":
+      case "F":
+      case "NO":
+      case "N":
+        return false;
+      default:
+        throw new IllegalArgumentException("Unknown boolean value, must 'True' or 'False' but was '" + str + "'");
```
cjkent: The error message doesn't match the actual behaviour.

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -46,6 +99,178 @@ public static Index findIndex(String reference) {
...
+    }
+  }
+
+  /**
+   * Parses a double from the input string.
+   * 
+   * @param str  the string to parse
+   * @return the parsed value
+   * @throws IllegalArgumentException if the string cannot be parsed
```
cjkent: It will actually throw `NumberFormatException`. It's a subclass of `IllegalArgumentException` so the comment isn't exactly wrong. But it could be more right.

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -46,6 +99,178 @@ public static Index findIndex(String reference) {
...
+  public static BuySell parseBuySell(String str) {
+    switch (str.toUpperCase(Locale.ENGLISH)) {
+      case "BUY":
+      case "B":
+        return BuySell.BUY;
+      case "SELL":
+      case "S":
+        return BuySell.SELL;
+      default:
+        throw new IllegalArgumentException("Unknown BuySell value, must 'Buy' or 'Sell' but was '" + str + "'");
```
cjkent: Missing "be" after "must". Also the error message is a bit misleading.

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -46,6 +99,178 @@ public static Index findIndex(String reference) {
...
+    switch (str.toUpperCase(Locale.ENGLISH)) {
+      case "PAY":
+      case "P":
+        return PayReceive.PAY;
+      case "RECEIVE":
+      case "REC":
+      case "R":
+        return PayReceive.RECEIVE;
+      default:
+        throw new IllegalArgumentException("Unknown PayReceive value, must 'Pay' or 'Receive' but was '" + str + "'");
```
cjkent: Missing "be", slightly misleading.

Issue #1493, opened 2017-03-07
Allow ValueWithFailures to be easily combined


cjkent:
I've come across a few cases where I've had to retrofit `ValueWithFailures` to code that uses streams. In order to collect the values and failures from multiple calculations it was necessary to convert the streams code into a for loop and accumulate into mutable containers. This is unfortunate and could be avoided if there were an easy way to combine multiple `ValueWithFailures` instances. The obvious options are:

1. Add `ValueWithFailures.combinedWith(ValueWithFailures<T>, BinaryOperator<T>)`. This could be used with `Stream.reduce()`
2. Add `Guavate.toValueWithFailures(T, BinaryOperator<T>)` returning a `Collector`
3. Add `ValueWithFailures.of(T, BinaryOperator<T>, ValueWithFailures<T>...)` and `ValueWithFailures(T, BinaryOperator<T>, List<ValueWithFailures<T>>)`

I think option 1) might be awkward because it would require nested lambdas inside the call to `reduce`. Option 2) would fit the streams use case pretty naturally and seems like a good option. Option 3) is probably worth doing too because it's simple and uncontroversial.

cjkent:
Option 1) was addressed by 7e152f08cb27329cf3d7101a774be4380a774ac6.

Leaving the issue open as 2) and 3) would still be useful.

jodastephen:
Item 2 has been added.

PR #1552, opened 2017-09-01
Added new collector to combine ValueWithFailures

davidcervi:
Added new collector to allow ValueWithFailures to be easily combined.

Fixes Option 2) of #1493

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
@@ -13,6 +13,8 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.function.BinaryOperator;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
 
 import org.joda.beans.Bean;
 import org.joda.beans.BeanBuilder;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -87,6 +89,27 @@
   }
 
   //-------------------------------------------------------------------------
+
+  /**
+   * Returns a {@link Collector} which performs a reduction of its
+   * {@link ValueWithFailures} input elements under a specified {@link BinaryOperator} using the
+   * provided identity.
```
jodastephen: In Strata we avoid using `@link` in the first sentence as it appears in higher-level Javadoc.

```
 * Returns a collector that can be used to create an instance from a stream of amounts.
 * <p>
 * <your more detailed description>
```

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -87,6 +89,27 @@
   }
 
   //-------------------------------------------------------------------------
+
```
jodastephen: No blank line after separators

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -87,6 +89,27 @@
...
+  /**
+   * Returns a {@link Collector} which performs a reduction of its
+   * {@link ValueWithFailures} input elements under a specified {@link BinaryOperator} using the
+   * provided identity.
+   *
+   * @param identityValue the identity value
+   * @param operator the operator used for the reduction.
+   * @param <T> the type of the success value in the {@link ValueWithFailures}
```
jodastephen: Strata puts `<T>` parameter first.
Strata has two spaces not one after parameter name

```
 * @param <T> the type of the success value in the {@link ValueWithFailures}
 * @param identityValue  the identity value
 * @param operator  the operator used for the reduction.
```

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -87,6 +89,27 @@
...
+  public static <T> Collector<ValueWithFailures<T>, ?, ValueWithFailures<T>> toValueWithFailures(
+      T identityValue,
+      BinaryOperator<T> operator) {
+
+    BinaryOperator<ValueWithFailures<T>> reduceFunction =
+        (result1, result2) -> result1.combinedWith(result2, operator);
+
+    return Collectors.reducing(ValueWithFailures.of(identityValue), reduceFunction);
+  }
+
```
jodastephen: Should be a separator after static methods and before instance ones (because there is already in this class)

PR #1553, opened 2017-09-01
Default day count from index

jodastephen:
Use additional `FloatingRateIndex` logic to default day count.

Determines the index for each leg first, then calculates the default day count, then parses.

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
@@ -18,8 +18,10 @@ import static com.opengamma.strata.loader.csv.TradeCsvLoader.START_DATE_FIELD;
 import java.time.LocalDate;
 import java.time.Period;
 import java.util.ArrayList;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.primitives.Ints;
@@ -34,7 +36,9 @@ import com.opengamma.strata.basics.date.DaysAdjustment;
 import com.opengamma.strata.basics.date.HolidayCalendarId;
 import com.opengamma.strata.basics.date.HolidayCalendarIds;
 import com.opengamma.strata.basics.date.Tenor;
+import com.opengamma.strata.basics.index.FloatingRateIndex;
 import com.opengamma.strata.basics.index.FloatingRateName;
+import com.opengamma.strata.basics.index.FloatingRateType;
 import com.opengamma.strata.basics.index.FxIndex;
 import com.opengamma.strata.basics.index.IborIndex;
 import com.opengamma.strata.basics.index.OvernightIndex;
@@ -70,6 +74,7 @@ import com.opengamma.strata.product.swap.RateCalculation;
 import com.opengamma.strata.product.swap.RateCalculationSwapLeg;
 import com.opengamma.strata.product.swap.ResetSchedule;
 import com.opengamma.strata.product.swap.Swap;
+import com.opengamma.strata.product.swap.SwapLeg;
 import com.opengamma.strata.product.swap.SwapTrade;
 
 /**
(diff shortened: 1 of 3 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -147,27 +152,106 @@
...
+    Optional<String> payReceiveOpt = Optional.of(getValue(row, legPrefix, DIRECTION_FIELD));
     int i = 1;
-    while (payReceive.isPresent()) {
-      legs.add(parseLeg(row, "Leg " + i + " "));
+    while (payReceiveOpt.isPresent()) {
+      // parse this leg
+      FloatingRateIndex index = parseIndex(row, legPrefix);
+      if (index != null) {
+        // floating leg
+        indices.add(index);
```
brianweller89: indices.add(index) can be outside the if loop rather than here; will still add null for the fixed rate

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -147,27 +152,106 @@
...
+        dayCounts.add(index.getDefaultFixedLegDayCount());
+      } else {
+        // fixed leg
+        indices.add(null);
+        if (!findValue(row, legPrefix, DAY_COUNT_FIELD).isPresent()) {
+          missingDayCount = true;
```
brianweller89: Do we even need the missingDayCount check? Feels like this is an unnecessary layer of complexity in what is already quite complex code.

What's the harm of passing a non-null default into the parseLegs method? Will just be ignored if the fixed leg has defined value.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -147,27 +152,106 @@
...
+      payReceiveOpt = findValue(row, legPrefix, DIRECTION_FIELD);
     }
+    // determine the default day count for the fixed leg
+    DayCount defaultFixedLegDayCount = null;
+    if (missingDayCount) {
+      if (dayCounts.size() > 1) {
+        throw new IllegalArgumentException(
+            "Swap leg must define day count on fixed legs when more than 2 floating legs");
+      }
+      defaultFixedLegDayCount = dayCounts.iterator().next();
```
brianweller89: I like the readibility of Iterables.getOnlyElement() in these situations

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -293,68 +377,49 @@ private static NotionalSchedule parseNotionalSchedule(CsvRow row, String leg) {
   private static RateCalculation parseRateCalculation(
       CsvRow row,
       String leg,
+      FloatingRateIndex index,
+      DayCount defaultFixedLegDayCount,
       BusinessDayAdjustment bda,
       Frequency accrualFrequency,
```
brianweller89: accrual frequency not used anymore

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -147,27 +152,106 @@
...
+      throw new IllegalArgumentException(
+          "Swap leg must define either '" + leg + FIXED_RATE_FIELD + "' or  '" + leg + INDEX_FIELD + "'");
+    }
+    // use FloatingRateName to identify Ibor vs other
+    String indexStr = indexOpt.get();
+    FloatingRateName frn = FloatingRateName.parse(indexStr);
+    if (frn.getType() == FloatingRateType.IBOR) {
+      // re-parse Ibor using tenor, which ensures tenor picked up from indexStr if present
+      Frequency freq = Frequency.parse(getValue(row, leg, FREQUENCY_FIELD));
+      Tenor iborTenor = freq.isTerm() ? frn.getDefaultTenor() : Tenor.of(freq.getPeriod());
```
brianweller89: Not immediately related to this PR but our getDefaultTenor() logic is not fully correct here.

Default tenor for most major currencies is 6M, with only USD being 3M

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
jodastephen: Case is where there is a fixed leg and two floating legs with different day counts. Must only trigger the "check day unique" logic when day count is actually missing.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
jodastephen:
```
* This is useful for providing a basic default where errors need to be avoided.
* The value returned is not intended to be based on market conventions."
```

PR #1559, opened 2017-09-13
Provide ability to identify Strata CSV format

jodastephen:
Change column name to be sufficiently unique


modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvLoader.java
```diff
@@ -280,7 +298,7 @@ private TradeCsvLoader(ReferenceData refData) {
     try (CsvIterator csv = CsvIterator.of(charSource, true)) {
       if (!csv.headers().contains(TYPE_FIELD)) {
         return ValueWithFailures.of(ImmutableList.of(),
-            FailureItem.of(FailureReason.PARSING, "CSV file does not contain 'Type' header: {}", charSource));
+            FailureItem.of(FailureReason.PARSING, "CSV file does not contain '" + TYPE_FIELD + "' header: {}", charSource));
```
cjkent: Shouldn't this be:

    FailureItem.of(FailureReason.PARSING, "CSV file does not contain '{}' header: {}", TYPE_FIELD, charSource));

PR #1568, opened 2017-09-26
Synthetic calibrator: typed market data object

marc-henrard:
Synthetic curve calibrator: returns a more explicit object with the market data. 
It is useful for other processes to know the exact type of the MarketData object returned.

### DIFF
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -243,7 +243,8 @@
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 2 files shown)


modules/pom.xml
```diff
@@ -243,6 +243,18 @@
           </execution>
         </executions>
       </plugin>
+      <!--  Running Maven tests without slow HTML reports -->
```
jodastephen: This is already set in the corporate parent

PR #1572, opened 2017-10-03
Parse variable notional/rates

jodastephen:
There are lots of variable elements in Strata swaps, but the most crucial are notionals and fixed rates. While in theory these could be different on each leg, in most cases they are not. Thus, this PR concentrates on the standard variable cases - notional and fixed rate affecting all legs equally.


modules/loader/src/main/java/com/opengamma/strata/loader/csv/SwapTradeCsvLoader.java
```diff
@@ -61,11 +67,69 @@
...
+  private static SwapTrade parseVariableNotional(SwapTrade trade, List<CsvRow> variableRows) {
+    // parse notionals
+    ImmutableList.Builder<ValueStep> stepBuilder = ImmutableList.builder();
+    for (CsvRow row : variableRows) {
+      LocalDate date = LoaderUtils.parseDate(row.getValue(START_DATE_FIELD));
+      row.findValue(NOTIONAL_FIELD)
+          .map(str -> LoaderUtils.parseDouble(str))
+          .ifPresent(notional -> stepBuilder.add(ValueStep.of(date, ValueAdjustment.ofReplace(notional))));
+    }
+    ImmutableList<ValueStep> vnotionals = stepBuilder.build();
```
brianweller89: vNotionals rather than vnotionals

modules/loader/src/main/java/com/opengamma/strata/loader/csv/SwapTradeCsvLoader.java
```diff
@@ -61,11 +67,69 @@
...
+  // variable fixed rate
+  private static SwapTrade parseVariableRates(SwapTrade trade, List<CsvRow> variableRows) {
+    ImmutableList.Builder<ValueStep> stepBuilder = ImmutableList.builder();
+    for (CsvRow row : variableRows) {
+      LocalDate date = LoaderUtils.parseDate(row.getValue(START_DATE_FIELD));
+      row.findValue(FIXED_RATE_FIELD)
+          .map(str -> LoaderUtils.parseDoublePercent(str))
+          .ifPresent(fixedRate -> stepBuilder.add(ValueStep.of(date, ValueAdjustment.ofReplace(fixedRate))));
+    }
+    ImmutableList<ValueStep> vnotionals = stepBuilder.build();
```
brianweller89: rates rather than notionals

PR #1577, opened 2017-10-05
Add new plus/minus methods for combining CurrencyScenarioArrays

brianweller89:


modules/data/src/main/java/com/opengamma/strata/data/scenario/CurrencyScenarioArray.java
```diff
@@ -142,7 +142,64 @@ public CurrencyScenarioArray convertedTo(Currency reportingCurrency, ScenarioFxR
...
+  }
+  
+  /**
+   * Returns a new array containing the values from this array with the values from the amount added.
+   * <p>
+   * The amount is added to each element in this array.
+   * The currency must be the same as the currency of this array.
+   *
+   * @param amount  the amount to add
+   * @return a new array containing the values from this array added to the values in the other array
```
jodastephen: Not quite right

modules/data/src/main/java/com/opengamma/strata/data/scenario/CurrencyScenarioArray.java
```diff
@@ -142,7 +142,64 @@ public CurrencyScenarioArray convertedTo(Currency reportingCurrency, ScenarioFxR
...
+  
+  /**
+   * Returns a new array containing the values from this array with the values from the other array subtracted.
+   * <p>
+   * The amounts are subtracted from the matching element in this array.
+   * The currency must be the same as the currency of this array.
+   * The arrays must have the same size.
+   *
+   * @param other  another array of multiple currency values.
+   * @return a new array containing the values from this array added with the values from the other array subtracted
```
jodastephen: Not "added"

modules/data/src/main/java/com/opengamma/strata/data/scenario/CurrencyScenarioArray.java
```diff
@@ -142,7 +142,64 @@ public CurrencyScenarioArray convertedTo(Currency reportingCurrency, ScenarioFxR
...
+  }
+  
+  /**
+   * Returns a new array containing the values from this array with the values from the amount subtracted.
+   * <p>
+   * The amount is subtracted from each element in this array.
+   * The currency must be the same as the currency of this array.
+   *
+   * @param amount  the amount to subtract
+   * @return a new array containing the values from this array with the values from the amount subtracted
```
jodastephen: Not quite right

modules/data/src/main/java/com/opengamma/strata/data/scenario/CurrencyScenarioArray.java
brianweller89: Confession time; I copy/pasted all of these new Javadoc from CurrencyAmountArray!

Will clean that up as well.

modules/data/src/main/java/com/opengamma/strata/data/scenario/CurrencyScenarioArray.java
brianweller89: What's wrong with this one? Only issue I can see is 'values' rather than value

jodastephen: `@return a new array containing the values from this array with the specified amount subtracted`

PR #1580, opened 2017-10-20
Support adjusted start date for StandardRollConventions (e.g. IMM).

brianweller89:
Previous code only supported conversion of adjusted to unadjusted dates for numeric or EOD convention.

This code attempts to calculate an unadjusted date from an adjusted date as follows
-  Calculate the valid unadjusted date for the month of the passed date
-  Adjust the calculated date according to the trade adjustment convention 
-  Compare the calculated adjusted date to the passed date. 
- If the passed date matches the calculated adjusted date it is assumed the passed date has been pre-adjusted and the unadjusted date is returned instead.

This relies on the assumption that the adjusted roll date will never be in a different month to the original unadjusted date. This is relatively safe for the existing StandardRollConventions as the rules for choosing a day in the month will never result in a date close to EOM.  The latest possible calculated roll day for the existing roll convention is the 21st of the month (for IMM, if the first of the month is a Thursday). This is sufficiently far from EOM that no reasonable adjustment will result in the adjusted date being in the next month. Similarly the earliest possible date is 7th of the month (for IMMNZD).

The existing implementation for numeric roll convention suffers from a similar problem e.g. if unadjusted roll day of 28th is adjusted using FOLLOWING then the resulting adjusted date could be in the following month. This would result in the date not being identified as adjusted.

This logic will not work for weekly roll conventions in most cases. adjust() method for weekly roll conventions calculates the next weekday which matches the convention. If the passed date has been adjusted forward then it will not be identified as a valid adjusted date; the calculated unadjusted date will be the next occurrence of that weekday, rather than the previous e.g.
- For weekly roll convention  = Wednesday
- If unadjusted start date is a Wednesday which is a bad business day and adjusted start date is the following Thursday.
- Dow.adjust(startDate) will return the next Wednesday, which, when adjusted, will never equal the passed adjustedDate. Hence unadjusted date will not be calculated

See comments in code for (even) more detail.

An alternative approach could be to associate the logic for converting to unadjusted with each roll convention by adding a new method to the RollConvention interface; this would however be an intrusive and breaking change.

This is still a bit of a hack, so open to suggestions for alternatives!

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
@@ -841,7 +841,9 @@ public final class PeriodicSchedule
     // or  
     // StandardDayConvention is used and the day is not a valid roll date
 
-    if (refData != null && BusinessDayAdjustment.NONE.equals(startDateBusinessDayAdjustment)) {
+    if (refData != null && 
+        rollConvention != null && 
+        BusinessDayAdjustment.NONE.equals(startDateBusinessDayAdjustment)) {
         return calculatedUnadjustedDateFromAdjusted(startDate, rollConvention, businessDayAdjustment, refData);
     }
     return startDate;
@@ -849,7 +851,7 @@ public final class PeriodicSchedule
 
   // calculates the applicable end date
   private LocalDate calculatedUnadjustedEndDate(ReferenceData refData) {
-    if (refData != null) {
+    if (refData != null && rollConvention != null) {
       return calculatedUnadjustedDateFromAdjusted(endDate, rollConvention, calculatedEndDateBusinessDayAdjustment(), refData);
     }
     return endDate;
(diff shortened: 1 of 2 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
```diff
@@ -830,48 +830,82 @@ private RollConvention calculatedRollConvention(LocalDate calculatedFirstRegStar
...
-    int lengthOfMonth = baseDate.lengthOfMonth();
-    int actualDom = Math.min(rollDom, lengthOfMonth);
-    // startDate is already the expected day, then nothing to do
-    if (baseDate.getDayOfMonth() != actualDom) {
-      LocalDate rollImpliedDate = baseDate.withDayOfMonth(actualDom);
-      LocalDate adjDate = businessDayAdjustment.adjust(rollImpliedDate, refData);
-      if (adjDate.equals(baseDate)) {
-        return rollImpliedDate;
+  
+    if (rollConvention != null) {
```
jodastephen: Worth checking this in the calling code I think

modules/basics/src/test/java/com/opengamma/strata/basics/schedule/PeriodicScheduleTest.java
```diff
@@ -217,254 +221,272 @@ private PeriodicSchedule createDates(LocalDate start, LocalDate end, LocalDate f
...
-        {date(2014, 9, 17), date(2014, 10, 1), Frequency.ofDays(2), STUB_NONE, IMM, null, null, null,
+        {date(2014, 9, 17), date(2014, 10, 1), Frequency.ofDays(2), STUB_NONE, IMM, BDA, null, null, null,
             ImmutableList.of(date(2014, 9, 17), date(2014, 10, 1)),
             ImmutableList.of(date(2014, 9, 17), date(2014, 10, 1)), IMM},
 
+        //IMM with adjusted start dates and various conventions
+        {date(2018, 3, 22), date(2020, 03, 18), P6M, STUB_NONE, IMM, BDA_JPY_MF, null, null, BDA_NONE,
+            ImmutableList.of(date(2018, 3, 21), date(2018, 9, 19), date(2019, 3, 20), date(2019, 9, 18), date(2020, 3, 18)),
+            ImmutableList.of(date(2018, 3, 22), date(2018, 9, 19), date(2019, 3, 20), date(2019, 9, 18), date(2020, 3, 18)), IMM},
+        {date(2018, 3, 20), date(2019, 03, 20), P6M, STUB_NONE, IMM, BDA_JPY_P, null, null, BDA_NONE,
```
jodastephen: a test without STUB_NONE?

modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
```diff
@@ -830,48 +830,82 @@ private RollConvention calculatedRollConvention(LocalDate calculatedFirstRegStar
...
+          //adjust() method for "day of week" roll conventions will roll forward from the passed date; hence this logic
+          //will not work for "day of week" conventions if the passed baseDate has been adjusted to be after the original 
+          //unadjusted date (i.e. has been rolled forward).
+        
+          //Calculate the expected adjusted roll date, based on the valid unadjusted roll date
+          LocalDate adjustedRollDate = businessDayAdjustment.adjust(unadjustedValidRollDate, refData);
+        
+          //If the adjusted roll date equals the original base date then that the base date is in fact an adjusted date
+          //and hence return the unadjusted date for building the schedule.
+          if (adjustedRollDate.equals(baseDate)) {
```
jodastephen: It would be helpful to use the same local variable names for similar concepts, such as `rollInpliedDate` and `adjDate`.

PR #1581, opened 2017-10-23
Provide the ability to control payment periods

jodastephen:
Add firstRegularStartDate and lastRegularEndDate to PaymentSchedule
Update FpML and CSV trade parser


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/Schedule.java
```diff
@@ -224,6 +247,23 @@ private boolean isFinalStub() {
     return (startStub == 0 && endStub == 0 ? periods : periods.subList(startStub, periods.size() - endStub));
   }
 
+  /**
+   * Gets the complete list of unadjusted dates.
+   * <p>
+   * This returns a list including all the unadjusted period boundary dates.
+   * This is the same as the unadjusted start date of each period plus the final unadjusted end date.
```
brianweller89: Not a big deal but it reads a little odd that the comment refers to the list as start date of all periods + final end date whereas the code works off the opposite logic (start date of first period + all end dates)

PR #1584, opened 2017-10-24
Extend position CSV parsing

jodastephen:
Add mode whereby ETD positions can be parsed as `SecurityPosition` without using reference data

`CsvLoaderUtils` is existing code that has been moved.


modules/loader/src/main/java/com/opengamma/strata/loader/csv/PositionCsvInfoResolver.java
```diff
@@ -114,19 +133,109 @@ public default SecurityPosition completePosition(CsvRow row, SecurityPosition po
...
     EtdContractSpecId specId = EtdIdUtils.contractSpecId(type, exchangeId, contractCode);
     return getReferenceData().findValue(specId).orElseThrow(
         () -> new IllegalArgumentException("ETD contract specification not found in reference data: " + specId));
   }
 
+  /**
+   * Parses an ETD future position from the CSV row.
+   * <p>
+   * In most cases this will return {@link EtdFuturePosition}, however for flexibility
+   * it is allowed to return any position, such as {@link SecurityPosition}.
```
cjkent: The comment doesn't match the return type. Should it actually return `Position` like the method below?

PR #1586, opened 2017-10-25
Extend quoting of CSV output

jodastephen:
If a CSV entry starts with an expression prefix character
then prefix the quoted section by equals
This is invalid pure CSV, but accepted by Excel/Google
to indicate a test formatted cell
Update CSV input to parse equals prefix format

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
@@ -23,6 +23,7 @@ import com.opengamma.strata.collect.Unchecked;
  * Each entry may be quoted using a double quote.
  * If an entry contains a double quote, comma or trimmable whitespace, it will be quoted.
  * Two double quotes will be used to escape a double quote.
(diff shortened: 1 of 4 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
```diff
@@ -132,17 +133,28 @@ private String formatEntry(String entry, boolean alwaysQuote) {
...
+  // quoting is required if entry contains quote, comma, trimmable whitespace, or starts with an expression character
+  private boolean isQuotingRequired(String entry) {
+    return entry.indexOf('"') >= 0 ||
+        entry.indexOf(',') >= 0 ||
+        entry.trim().length() != entry.length() ||
+        isExpressionPrefix(entry);
+  }
+
+  private boolean isExpressionPrefix(String entry) {
+    char first = entry.length() == 0 ? ' ' : entry.charAt(0);
```
brianweller89: Feels odd to be assigning empty char here given we know empty char will always return false in the subsequent check. Admittedly the alternatives would require more code but this is abit random IMO.

PR #1590, opened 2017-10-27
Use actual start month and end month rather than proleptic months to determine EOM rolls.

brianweller89:
This fixes an edge case for February. For EOM trades, If start and end date are in February and one (but not both) was in a a leap year then the roll convention was calculated as 29 rather than EOM.

In terms of existing behaviour, only other change is to the below use case:
* If start and end date are in same month but not the same year
* With different day of month
* End date day of month is the last day of the month

This case previously returned EOM roll day but now returns start day of month as the roll day. A trade with this combination of regular period start/end date should only be valid if it is a single period swap, in which case roll day would not be relevant.

Also added new tests for existing use cases.

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/StubConvention.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/StubConvention.java
@@ -5,6 +5,8 @@
  */
 package com.opengamma.strata.basics.schedule;
 
+import static java.time.temporal.ChronoField.PROLEPTIC_MONTH;
+
 import java.time.LocalDate;
 
 import org.joda.convert.FromString;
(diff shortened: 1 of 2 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/StubConvention.java
```diff
@@ -259,7 +257,7 @@ public RollConvention toRollConvention(
     // dates are at the end of the month, and in different months
     if (this == NONE) {
       if (start.getDayOfMonth() != end.getDayOfMonth() &&
-          start.getLong(PROLEPTIC_MONTH) != end.getLong(PROLEPTIC_MONTH) &&
+          start.getMonth() != end.getMonth() &&
```
jodastephen: This seems like an even odder test than before. A schedule from July to July will handled differently to one from July to August.

Maybe keep the existing code, but inside the block check `preferEndOfMonth` and return `EOM` if true?

brianweller89: Per the existing comment it looks like the original intention here was to handle the case where the months are different, so varying behaviour when months are different vs the same doesn't seem unexpected?

An effect of the above suggestion would be that an EOM trade with start on 28 February and end 30th April would change to return EOM roll convention rather than 30. This sounds fine in the context of "preferEndOfMonth" being true but does feel like a slightly more intrusive change in behaviour than currently proposed.

PR #1595, opened 2017-11-06
Conventional currency pair for EUR/GBP swap

yukiiwashita:
The currency pair in EUR/GBP swap should be conventional.

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConventions.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConventions.java
@@ -27,12 +27,12 @@ public final class FxSwapConventions {
       FxSwapConvention.of(StandardFxSwapConventions.EUR_USD.getName());
 
   /**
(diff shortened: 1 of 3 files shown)


modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConventions.java
```diff
@@ -27,12 +27,12 @@
...
   /**
-   * The "GBP/EUR" FX Swap convention.
+   * The "EUR/GBP" FX Swap convention.
    * <p>
-   * GBP/EUR convention with 2 days spot date.
+   * EUR/GBP convention with 2 days spot date.
    */
-  public static final FxSwapConvention GBP_EUR =
-      FxSwapConvention.of(StandardFxSwapConventions.GBP_EUR.getName());
+  public static final FxSwapConvention EUR_GBP =
```
jodastephen: You'll need to keep the old one, but deprecated.

PR #1627, opened 2017-12-15
Provide human-readable summaries

jodastephen:
Summaries for Positions


modules/product/src/main/java/com/opengamma/strata/product/etd/EtdFutureSecurity.java
```diff
@@ -123,6 +128,15 @@ public Trade createTrade(TradeInfo tradeInfo, double quantity, double tradePrice
         .build();
   }
 
+  /**
+   * Summarizes this ETD future into string form.
+   *
+   * @return the summary description
+   */
+  public String summaryDescription() {
+    return expiry.format(YM_FORMAT) + variant.getCode();
```
brianweller89: Seems odd that the summary description for a future security doesn't include some form of contract identifier?

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdOptionSecurity.java
```diff
@@ -157,6 +163,27 @@ public Trade createTrade(TradeInfo tradeInfo, double quantity, double tradePrice
...
+    f.setGroupingUsed(false);
+    f.setMaximumFractionDigits(8);
+    String strikeStr = f.format(strikePrice).replace('-', 'M');
+
+    return expiry.format(YM_FORMAT) +
+        variant.getCode() + " " +
+        versionCode +
+        putCallStr +
+        strikeStr;
+  }
```
brianweller89: Similar to above, why doesn't this include the id?

modules/product/src/main/java/com/opengamma/strata/product/etd/EtdFutureSecurity.java
jodastephen: The full description includes the ID - "<ID> x <quantity>, <etddescription>. So this is only part of the story.

PR #1637, opened 2018-01-03
SyntheticCurveCalibrator: market data without time series

marc-henrard:
Added method to SyntheticCurveCalibrator which does not return the time-series from the input RatesProvider.


jodastephen:
This was replaced by #1638

PR #1648, opened 2018-01-24
Provide a factory method to configure decomposition function for CurveCalibrator

jurna:
Currently there is no possibility to provide a custom decomposition function to use in NewtonVectorRootFinder . Current SV decomposition implementation is based on apache commons math library, which is very inefficient. For bigger curve groups more efficient implementation is required. This pull request adds a new factory method (and new parameter in private constructor), which allows to configure a custom decomposition function.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/CurveCalibrator.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/CurveCalibrator.java
@@ -5,9 +5,6 @@
  */
 package com.opengamma.strata.pricer.curve;
 
-import static com.opengamma.strata.collect.Guavate.toImmutableList;
-import static com.opengamma.strata.collect.Guavate.toImmutableMap;
-
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


jodastephen:
Thanks for this PR. You are correct that there is no way to control the decomposition. However, your change uses an internal interface `Decomposition` on the public API, so another way will have to be found.

Looking at the change, we at OpenGamma are thinking that it may be better to allow the whole root finder to be passed into the calibrator, giving much more flexibility. We need to spend a little time evaluating the impact of that, so please bear with us.

jurna:
Yes my initial idea was also to allow root finder to be passed into calibrator. But you are already passing parameters (toleranceAbs, toleranceRel, stepMaximum) into BroydenVectorRootFinder, so adding another parameter into factory method seemed like a better idea. But I've overlooked that Decomposition is a class in the internal the api.

jodastephen:
Maybe something [like this](https://github.com/OpenGamma/Strata/commit/5b12f72819b4bfd20eec8e9938d8e13d826b8956). Not 100% sure yet.

jurna:
The other solution would be to make Decomposition class a public interface. Because at the moment I would have to implement the whole new NewtonVectorRootFinder. It's not a big deal, just a thought.

jodastephen:
Fixed by #1653, thanks

PR #1649, opened 2018-01-26
Introducing a Messages.formatWithAttributes() method + unit tests

andreiruse:
This new method will not only return the formatted message with values instead of placeholders but also a map of attributes.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
@@ -125,7 +125,7 @@ public final class Messages {
    * <p>
    * Typical template would look like:
    * <pre>
-   * "Messages.formatWithAttributes("Foo={foo}, Bar={}", "abc", 123)".
+   * Messages.formatWithAttributes("Foo={foo}, Bar={}", "abc", 123)
    * </pre>
    * This will return a {@link Pair} with a String and a Map.
    * The String will be the message, will look like: "Foo=abc, Bar=123"
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
@@ -5,6 +5,7 @@
  */
 package com.opengamma.strata.collect;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -147,12 +148,12 @@ public final class Messages {
     if (args == null) {
       return formatWithAttributes(messageTemplate, new Object[0]);
     }
+
     // try to make builder big enough for the message and the args
-
     StringBuilder outputStringBuilder = new StringBuilder(messageTemplate.length() + args.length * 20);
-    ImmutableMap.Builder<String, String> attributesMap = ImmutableMap.builder();
+    Map<String, String> attributesMap = new HashMap<>();
 
-    String pattern = "\\{(\\w+)\\}";
+    String pattern = "\\{(\\w*)\\}"; //This will match both {}, and {anything}
     Pattern regexPattern = Pattern.compile(pattern);
     Matcher matcher = regexPattern.matcher(messageTemplate);
     int groupIndex = 0;
@@ -163,7 +164,10 @@ public final class Messages {
     }
     if (matchesCount > args.length) {
       throw new IllegalArgumentException(
-          Messages.format("You have included {} placeholders, however only provided {} arguments.", matchesCount, args.length));
+          Messages.format(
+              "You have included {} placeholders, however only provided {} arguments.",
+              matchesCount,
+              args.length));
     }
     matcher.reset(); //Since we have already called .find(), and that state needs to be reset
     while (matcher.find()) {
@@ -172,7 +176,9 @@ public final class Messages {
       int endIndex = matcher.end();
 
       String replacement = args[groupIndex].toString();
-      attributesMap.put(attributeName, replacement);
+      if (!attributeName.isEmpty()) {
+        attributesMap.put(attributeName, replacement);
+      }
       String unmodifiedText = messageTemplate.substring(lastAddedStringEndIndex, startIndex);
       outputStringBuilder
           .append(unmodifiedText)
@@ -185,7 +191,7 @@ public final class Messages {
       outputStringBuilder.append(messageTemplate.substring(lastAddedStringEndIndex, messageTemplate.length()));
     }
 
-    return Pair.of(outputStringBuilder.toString(), attributesMap.build());
+    return Pair.of(outputStringBuilder.toString(), ImmutableMap.copyOf(attributesMap));
   }
 
 }
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
@@ -151,25 +151,22 @@ public final class Messages {
     //Do not use an ImmutableMap, as we avoid throwing exceptions in case of duplicate keys.
     Map<String, String> attributes = new HashMap<>();
     Matcher matcher = REGEX_PATTERN.matcher(messageTemplate);
-    int groupIndex = 0;
+    int argIndex = 0;
 
     StringBuffer outputMessageBuffer = new StringBuffer();
-    int matchesCount = 0;
     while (matcher.find()) {
-      matchesCount++;
-
       //If the number of placeholders is greater than the number of arguments, then not all placeholders are replaced.
-      if (matchesCount > args.length) {
+      if (argIndex >= args.length) {
         continue;
       }
 
       String attributeName = matcher.group(1); //Extract the attribute name
-      String replacement = args[groupIndex].toString();
+      String replacement = args[argIndex].toString();
       matcher.appendReplacement(outputMessageBuffer, replacement);
       if (!attributeName.isEmpty()) {
         attributes.put(attributeName, replacement);
       }
-      groupIndex++;
+      argIndex++;
     }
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+    int matchesCount = 0;
+    while (matcher.find()) {
+      matchesCount++;
+    }
+    if (matchesCount > args.length) {
+      throw new IllegalArgumentException(
+          Messages.format("You have included {} placeholders, however only provided {} arguments.", matchesCount, args.length));
+    }
+    matcher.reset(); //Since we have already called .find(), and that state needs to be reset
+    while (matcher.find()) {
```
cjkent: Can you use `Matcher.appendReplacement()` to simplify this?

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#appendReplacement-java.lang.StringBuffer-java.lang.String-

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
     return builder.toString();
   }
 
+  /**
+   * Formats a templated message inserting named arguments.
```
cjkent: I don't think these Javadoc make it clear enough what the method does. An example would be worth a thousand word.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+    String pattern = "\\{(\\w+)\\}";
+    Pattern regexPattern = Pattern.compile(pattern);
+    Matcher matcher = regexPattern.matcher(messageTemplate);
+    int groupIndex = 0;
+    int lastAddedStringEndIndex = 0;
+    int matchesCount = 0;
+    while (matcher.find()) {
+      matchesCount++;
+    }
+    if (matchesCount > args.length) {
```
cjkent: Could you roll this check into the replacement loop?

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+   * not throw another exception, which might hide the intended exception.
+   *
+   * @param messageTemplate  the message template with "{}" placeholders, null returns empty string
+   * @param args  the message arguments, null treated as empty array
+   * @return the formatted message
+   * @throws IllegalArgumentException if you have provided less arguments than placeholders
+   */
+  public static Pair<String, Map<String, String>> formatWithAttributes(String messageTemplate, Object... args) {
+    if (messageTemplate == null) {
+      return formatWithAttributes("", args);
```
cjkent: Would this achieve the same thing?

    return "";

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+    }
+    if (args == null) {
+      return formatWithAttributes(messageTemplate, new Object[0]);
```
cjkent: Is this the same as:

    return messageTemplate;

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+    }
+    // try to make builder big enough for the message and the args
+
+    StringBuilder outputStringBuilder = new StringBuilder(messageTemplate.length() + args.length * 20);
+    ImmutableMap.Builder<String, String> attributesMap = ImmutableMap.builder();
+
+    String pattern = "\\{(\\w+)\\}";
+    Pattern regexPattern = Pattern.compile(pattern);
```
cjkent: Compiling a pattern is relatively expensive. Do it once and put it in a static final field. `Pattern` is thread safe and can be shared.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +118,74 @@ public static String format(String messageTemplate, Object... args) {
...
+    Matcher matcher = regexPattern.matcher(messageTemplate);
```
cjkent: Using a regex might be a performance problem. But manually parsing is likely to be fiddly. I'd use a regex for simplicity until it's obviously a problem.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +119,79 @@ public static String format(String messageTemplate, Object... args) {
...
+  /**
+   * Formats a templated message inserting named arguments.
+   * <p>
+   * This method combines a template message with a list of specific arguments.
+   * It can be useful to delay string concatenation, which is sometimes a performance issue.
+   * The approach is similar to SLF4J MessageFormat, Guava Preconditions and String format().
+   * <p>
+   * The message template contains zero to many "{name}" placeholders.
+   * Each placeholder is replaced by the next available argument.
+   * Empty "{}" placeholders do not get replaced, and will be present in the output in the same form.
```
jodastephen: No. `{}` is still a placeholder, just an unamed one.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +119,79 @@ public static String format(String messageTemplate, Object... args) {
...
+    if (messageTemplate == null) {
+      return formatWithAttributes("", args);
+    }
+    if (args == null) {
+      return formatWithAttributes(messageTemplate, new Object[0]);
+    }
+
+    // try to make builder big enough for the message and the args
+    StringBuilder outputStringBuilder = new StringBuilder(messageTemplate.length() + args.length * 20);
+    Map<String, String> attributesMap = new HashMap<>();
```
jodastephen: Add a comment to not use `ImmutableMap` as duplicate keys must be allowed even though there is data loss (to avoid exceptions)

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
andreiruse: Logically, yes. However, the method has to return `Pair<String, Map<String, String>>` - returning a String would not work.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
andreiruse: Logically, yes. However, the method has to return `Pair<String, Map<String, String>>` - returning a String would not work

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +119,79 @@ public static String format(String messageTemplate, Object... args) {
...
+
+    String pattern = "\\{(\\w*)\\}"; //This will match both {}, and {anything}
+    Pattern regexPattern = Pattern.compile(pattern);
+    Matcher matcher = regexPattern.matcher(messageTemplate);
+    int groupIndex = 0;
+    int lastAddedStringEndIndex = 0;
+    int matchesCount = 0;
+    while (matcher.find()) {
```
jodastephen: This loop won't be needed once you get rid of the exception

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
cjkent: My point was that it's probably clearer to return immediately rather than calling back into the method and asking the reader to work out how that would behave. Like this:

    return Pair.of("", ImmutableMap.of());

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
andreiruse: That argument is outdated. Both named and unnamed placeholders are not replaced, as expected

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -146,52 +148,34 @@ public static String format(String messageTemplate, Object... args) {
...
-              args.length));
-    }
-    matcher.reset(); //Since we have already called .find(), and that state needs to be reset
-    while (matcher.find()) {
-      String attributeName = matcher.group(1); //Extract the attribute name
-      int startIndex = matcher.start();
-      int endIndex = matcher.end();
 
+      if (matchesCount > args.length) {
+        throw new IllegalArgumentException(
```
cjkent: Don't throw an exception. Just ignore the extra placeholders so they appear in the output as `{}` or `{foo}`.

modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -110,4 +120,57 @@ public static String format(String messageTemplate, Object... args) {
     return builder.toString();
   }
 
+  /**
+   * Formats a templated message inserting named arguments. Typical template would look like "Hello, {attributeName}".
```
cjkent: I still think this needs an example. I don't think it's clear from the docs what it does. Something like:

    Messages.formatWithAttributes("Foo={foo}, Bar={}", "abc", 123);

returns the message 

    "Foo=abc, Bar=123" 

and the map 

    {"foo": "123"}

Feel free to copy and paste :)

PR #1651, opened 2018-02-01
Add structured logging to FailureItem

andreiruse:
Making changes to FailureItem in order to use structured logging, provided by `Messages.formatWithAttributes()`.

Added a unit test for this use case.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
@@ -124,23 +124,6 @@ public final class FailureItem
     return new FailureItem(reason, message, ImmutableMap.of(), stackTrace, null);
   }
 
-  /**
-   * Obtains a failure from a reason and message.
-   * <p>
-   * The failure will still have a stack trace, but the cause type will not be present.
-   * 
-   * @param reason  the reason
-   * @param message  the failure message, not empty
-   * @param attributes the attributes associated with this failure
-   * @return the failure
-   */
-  public static FailureItem of(FailureReason reason, String message, Map<String, String> attributes) {
-    ArgChecker.notNull(reason, "reason");
-    ArgChecker.notEmpty(message, "message");
-    String stackTrace = localGetStackTraceAsString(message, 1);
-    return new FailureItem(reason, message, attributes, stackTrace, null);
-  }
-
   private static String localGetStackTraceAsString(String message, int skipFrames) {
     StringBuilder builder = new StringBuilder();
     StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
@@ -224,7 +224,7 @@ public final class FailureItem
     if (firstLine.endsWith(": " + message)) {
       return reason + ": " + message + ": " + firstLine.substring(0, firstLine.length() - message.length() - 2);
     }
-    if(!attributes.isEmpty()) {
+    if (!attributes.isEmpty()) {
       return reason + ": " + message + ": " + firstLine + " : " + attributes;
     }
     return reason + ": " + message + ": " + firstLine;
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
@@ -26,10 +26,12 @@ import org.joda.beans.impl.direct.DirectPrivateBeanBuilder;
 
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Interner;
 import com.google.common.collect.Interners;
 import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.collect.Messages;
+import com.opengamma.strata.collect.tuple.Pair;
 
 /**
  * Details of a single failed item.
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -78,8 +85,8 @@
   /**
    * Obtains a failure from a reason and message.
    * <p>
-   * The message is produced using a template that contains zero to many "{}" placeholders.
-   * Each placeholder is replaced by the next available argument.
+   * The message is produced using a template that contains zero to many "{}" or "{abc}" placeholders.
+   * Each placeholder is replaced by the next available argument. If the placeholder is name, it will be added to the attributes map.
```
cjkent: Typo "is name" should be "is named" or "has a name". Also the rest of the sentence could be clearer. Maybe something like "it's value is added to the attribute map with the name as the key".

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -61,6 +63,11 @@
    */
   @PropertyDefinition(validate = "notEmpty")
   private final String message;
+  /**
+   * The attributes associated with this failure.
```
cjkent: A bit more detail would be nice here. Something like "Attributes can contain additional information about the failure. For example, a line number in a file or the ID of a trade".

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -112,7 +119,25 @@ static FailureItem of(FailureReason reason, String message, int skipFrames) {
...
+  }
+
+  /**
+   * Obtains a failure from a reason and message.
+   * <p>
+   * The failure will still have a stack trace, but the cause type will not be present.
+   * 
+   * @param reason  the reason
+   * @param message  the failure message, not empty
+   * @param attributes the attributes associated to this failure
```
cjkent: "associated with"

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -112,7 +119,25 @@ static FailureItem of(FailureReason reason, String message, int skipFrames) {
...
+   * @param skipFrames  the number of caller frames to skip, not including this one
+   * @return the failure
+   */
+  static FailureItem of(FailureReason reason, String message, Map<String, String> attributes, int skipFrames) {
```
cjkent: This is package-private. I'm assuming you want it to be public. If so you shouldn't have the `skipFrames` parameter.

modules/collect/src/test/java/com/opengamma/strata/collect/result/FailureItemTest.java
```diff
@@ -75,4 +75,15 @@ public void test_of_reasonMessageExceptionNestedException() {
...
+  public void test_of_reasonMessageExceptionNestedExceptionWithAttributes() {
+    IllegalArgumentException innerEx = new IllegalArgumentException("inner");
+    IllegalArgumentException ex = new IllegalArgumentException("message", innerEx);
+    FailureItem test = FailureItem.of(FailureReason.INVALID, ex, "my {foo} {bar} failure", "big", "bad");
+    assertEquals(test.getReason(), FailureReason.INVALID);
+    assertEquals(test.getMessage(), "my big bad failure");
+    assertEquals(test.getCauseType().isPresent(), true);
+    assertEquals(test.getCauseType().get(), IllegalArgumentException.class);
+    assertEquals(test.getStackTrace().contains(".test_of_reasonMessageExceptionNestedExceptionWithAttributes("), true);
+    assertEquals(test.toString(), "INVALID: my big bad failure: java.lang.IllegalArgumentException: message : {bar=bad, foo=big}");
```
cjkent: It would be good to test the attributes map directly rather than doing it via the `toString()` method.

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -129,11 +131,11 @@ static FailureItem of(FailureReason reason, String message, int skipFrames) {
    * 
    * @param reason  the reason
    * @param message  the failure message, not empty
-   * @param attributes the attributes associated to this failure
+   * @param attributes the attributes associated with this failure
    * @param skipFrames  the number of caller frames to skip, not including this one
    * @return the failure
    */
-  static FailureItem of(FailureReason reason, String message, Map<String, String> attributes, int skipFrames) {
+  public static FailureItem of(FailureReason reason, String message, Map<String, String> attributes, int skipFrames) {
```
cjkent: Get rid of `skipFrames`, it's in internal implementation detail. Have a look at the other public `of` methods to see what they do.

modules/collect/src/main/java/com/opengamma/strata/collect/result/FailureItem.java
```diff
@@ -132,13 +132,12 @@ static FailureItem of(FailureReason reason, String message, int skipFrames) {
    * @param reason  the reason
    * @param message  the failure message, not empty
    * @param attributes the attributes associated with this failure
-   * @param skipFrames  the number of caller frames to skip, not including this one
    * @return the failure
    */
-  public static FailureItem of(FailureReason reason, String message, Map<String, String> attributes, int skipFrames) {
+  public static FailureItem of(FailureReason reason, String message, Map<String, String> attributes) {
```
cjkent: I've had second thoughts about including this method. I think it might be confusing. What if the user includes named placeholders in the message? Will they expect the attribute values to be substituted? Also, for symmetry you might expect the attribute map to be added to other factory methods. Better to keep things simple and get rid of it seeing as we said we didn't have a use case for it now. We can always add it later if we need to.

P.S. Sorry for changing my mind.

andreiruse: Done

PR #1654, opened 2018-02-13
Adding a failureMessage field to Result

andreiruse:
This is needed in order to provide transparency into the exception message and enable users to extract it in the UI and avoid Java-specific exception classes in when parsed from the Exception.


jodastephen:
See #1655

PR #1660, opened 2018-02-15
Add method to filter CurveGroupEntry

jodastephen:
Filter by index type


modules/market/src/main/java/com/opengamma/strata/market/curve/CurveGroupEntry.java
```diff
@@ -83,6 +85,24 @@ CurveGroupEntry merge(CurveGroupEntry newEntry) {
...
+  //-------------------------------------------------------------------------
+  /**
+   * Gets the subset of indices matching the specified type for which the curve provides forward rates.
+   * <p>
+   * The set of indices is filtered and cast using the specified type.
+   * This is empty if the curve is not used for forward rates.
+   * 
+   * @param <T>  the type of index
+   * @param indexType  the type of index required
+   * @return the set of indices as instances of {@code FloatingRateIndex}
```
brianweller89: I don't see anything referencing FloatingRateIndex in particular. Should this just be Index?

PR #1666, opened 2018-03-16
Control FX more effectively

jodastephen:
Move `FxRateLookup` to strata-calc and extend `CakculationParameter`
Change `CalculationTask` to use parameter
Extend `CalculationParameters` to search parent interfaces

This is a backwards incompatible change, however serialized instances of `RatesMarketDataLookup` will deserialize fine. Fixing the incompatibility simply requires an organize imports.

### DIFF
--- a/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
+++ b/modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
@@ -78,6 +78,9 @@ public final class CalculationParameters implements ImmutableBean, Serializable 
    * <p>
    * The list will be converted to a {@code Map} using {@link CalculationParameter#queryType()}.
    * Each parameter must refer to a different query type.
+   * <p>
+   * If a parameter implements an interface that also extends {@link CalculationParameter},
+   * that type will also be able to be searched for (unless it has been directly registered).
    * 
    * @param parameters  the parameters
    * @return the calculation parameters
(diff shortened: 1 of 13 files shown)


modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationParameters.java
```diff
@@ -110,6 +115,27 @@ private CalculationParameters of(Map<Class<? extends CalculationParameter>, Calc
...
+    ImmutableMap.Builder<Class<? extends CalculationParameter>, Class<? extends CalculationParameter>> aliases =
+        ImmutableMap.builder();
+    for (Class<? extends CalculationParameter> type : parameters.keySet()) {
+      Class<?>[] interfaces = type.getInterfaces();
+      for (Class<?> iface : interfaces) {
+        if (iface != CalculationParameter.class &&
+            CalculationParameter.class.isAssignableFrom(iface) &&
+            !parameters.containsKey(iface)) {
+
+          aliases.put(iface.asSubclass(CalculationParameter.class), type);
```
cjkent: This will fail if there are two values in `parameters` of different classes that both implement the same interface that is a subtype of `CalculationParameter`. It's ambiguous which one should be used so it should probably fail, but it might be nice to provide some explanation.

modules/calc/src/main/java/com/opengamma/strata/calc/runner/LookupScenarioFxRateProvider.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  private final FxRateLookup lookup;
+
+  // obtains an instance, returning the interface type to make type system happy at call site
+  static ScenarioFxRateProvider of(ScenarioMarketData marketData, FxRateLookup lookup) {
+    return new LookupScenarioFxRateProvider(marketData, lookup);
+  }
+
+  private LookupScenarioFxRateProvider(ScenarioMarketData marketData, FxRateLookup lookup) {
+    this.marketData = marketData;
```
cjkent: Null checks

PR #1667, opened 2018-03-20
Security quantity trade

yukiiwashita:
* Allow for creating a trade with quantity altered for any trades with quantity field.


modules/product/src/main/java/com/opengamma/strata/product/SecurityQuantityTrade.java
```diff
@@ -0,0 +1,35 @@
...
+    extends Trade, SecurityQuantity {
+
+  /**
+   * Gets the price that was traded.
+   * <p>
+   * This is the unit price agreed when the trade occurred.
+   * 
+   * @return the price
+   */
+  public abstract double getPrice();
```
jodastephen: For completeness, a `withPrice()` would make sense

PR #1676, opened 2018-04-30
Combine map-based futures

jodastephen:
Similar to list-based future combination

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -778,7 +778,7 @@ public final class Guavate {
    * Converts a map of futures to a single future.
    * <p>
    * This is similar to {@link #combineFuturesAsList(List)} but for maps.
-   * Effectively, this converts {@code Map<K, CompletableFuture<T>>} to {@code CompletableFuture<Map<K, T>>}.
+   * Effectively, this converts {@code Map<K, CompletableFuture<V>>} to {@code CompletableFuture<Map<K, V>>}.
    * <p>
    * If any input future completes exceptionally, the result will also complete exceptionally.
    *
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -772,6 +773,63 @@ public R next() {
     return collectingAndThen(toImmutableList(), Guavate::combineFuturesAsList);
   }
 
+  //-------------------------------------------------------------------------
+  /**
+   * Converts a map of futures to a single future.
+   * <p>
+   * This is similar to {@link #combineFuturesAsList(List)} but for maps.
+   * Effectively, this converts {@code Map<K, CompletableFuture<T>>} to {@code CompletableFuture<Map<K, T>>}.
```
cjkent: Probably worth using the same type param names in the Javadoc and the method signature (`K` and `T` vs `K` and `V`)

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -772,6 +773,63 @@ public R next() {
...
+          ImmutableMap.Builder<K, V> builder = ImmutableMap.builderWithExpectedSize(size);
+          for (int i = 0; i < size; i++) {
+            builder.put(keyArray[i], futuresArray[i].join());
+          }
+          return builder.build();
+        });
+  }
+
+  /**
+   * Collector used at the end of a stream to convert a list of futures to a single future,
```
cjkent: Copy / paste error - still says "list"

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -772,6 +773,63 @@ public R next() {
...
+   * combining the values into a list.
+   * <p>
+   * A collector is used to gather data at the end of a stream operation.
+   * This method returns a collector allowing a stream of futures to be combined into a single future.
+   * This converts {@code Map<K, CompletableFuture<T>>} to {@code CompletableFuture<Map<K, T>>}.
```
cjkent: Use same param names as signature

PR #1680, opened 2018-05-04
Gamma calculation with combined curve

yukiiwashita:
* Split a curve into underlying curves in the curve gamma calculator

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/sensitivity/CurveGammaCalculator.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/sensitivity/CurveGammaCalculator.java
@@ -13,18 +13,17 @@ import java.util.Map.Entry;
 import java.util.function.Function;
 import java.util.stream.IntStream;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.primitives.Doubles;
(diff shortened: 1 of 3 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/sensitivity/CurveGammaCalculator.java
```diff
@@ -124,10 +123,24 @@ public CrossGammaParameterSensitivities calculateCrossGammaIntraCurve(
       Currency currency = entry.getKey();
       Curve curve = entry.getValue();
       if (baseDelta.findSensitivity(curve.getName(), currency).isPresent()) {
-        NodalCurve nodalCurve = getNodalCurve(curve);
         CrossGammaParameterSensitivity gammaSingle = computeGammaForCurve(
-            nodalCurve, currency, c -> immProv.toBuilder().discountCurve(currency, c).build(), sensitivitiesFn);
+            curve, currency, c -> immProv.toBuilder().discountCurve(currency, c).build(), sensitivitiesFn);
         result = result.combinedWith(gammaSingle);
+      } else if (curve.split().size() > 1) {
```
jodastephen: Can this just be a foreach loop of the split curves? Won't that handle the non-combined case?

PR #1686, opened 2018-05-11
minor: update checkstyle to 8.0 and intoroduce checkstyle.version variable

romani:
first steps to use latest version of checkstyle , base on conversation at https://groups.google.com/forum/#!topic/checkstyle/KYF-9EEzsbs

We will use variable value to substitute it to stapshot version during regression testing.

### DIFF
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -118,7 +118,7 @@
           <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>7.4</version>
+            <version>${checkstyle.version}</version>
           </dependency>
           <dependency>
             <groupId>com.opengamma</groupId>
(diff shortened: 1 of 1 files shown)


romani:
@jodastephen ,

to proceed further with versions update we need to change https://github.com/OpenGamma/OG-Tools/blob/master/build-config/src/main/resources/checkstyle/checkstyle-oss.xml

project that host this file does not allow me to create issue for discussion/details - https://github.com/OpenGamma/OG-Tools .

Please let me know how you want to proceed.

after 8.0 we did several "Breaking backward compatibility" updates, so config have to changed
http://checkstyle.sourceforge.net/releasenotes.html#Release_8.1
"Make SuppressionCommentFilter and SuppressWithNearbyCommentFilter children of TreeWalker"
http://checkstyle.sourceforge.net/releasenotes.html#Release_8.2
"remove FileContentsHolder module as FileContents object is available for filters on TreeWalker in TreeWalkerAudit Event"

I tried to make changes but it damage build of Strata, so I need to spend a bit more time with further updates.

**Attention**: config is common for some project, so it might affect not only single project. There might be multiple fixes required in other projects.  Let me know how you want to proceed with such update.

PR #1687, opened 2018-05-11
Update checkstyle config

jodastephen:


romani:
Eclipse-cs did an effort on weekends and released new binaries base on 8.10 , you can upgrade if you want.

PR #1688, opened 2018-05-11
minor: bump checkstyle to 8.6 version

romani:
as config is updated to 8.6 state at https://github.com/OpenGamma/OG-Tools/pull/8

binaries have to be updated too, to let build pass.


romani:
actually you can bump version up to 8.10, Strata project build is passing without any problems on 8.10 .

jodastephen:
Fixed by #1687 thanks, let me know if anything else needs doing!

PR #1710, opened 2018-06-04
Resolve targets in calculation runner

jodastephen:
When processing `SecurityTrade` it is desirable to resolve the trade
before determining the function to use for calculation
This change ensures that the target is resolved at the correct stage
This may change behaviour where reference data contains real securities


jodastephen:
Not proceeding with this approach.

PR #1721, opened 2018-06-08
Add traded price for more security products

jodastephen:


modules/pricer/src/test/java/com/opengamma/strata/pricer/index/e2e/IborFuturesJpyEnd2EndTest.java
```diff
@@ -265,24 +265,19 @@ public void priceSensitivity() {
 
   public void presentValue() {
     // March 2016
-    CurrencyAmount pvMar = TRADE_PRICER.presentValue(FUTURE_TRADE_MAR, RATES_PROVIDER,
-        FUTURE_TRADE_MAR.getPrice());
+    CurrencyAmount pvMar = TRADE_PRICER.presentValue(FUTURE_TRADE_MAR, RATES_PROVIDER, REF_PRICE_MAR_DECIMAL);
     assertEquals(pvMar.getAmount(), -9738.418878056109, TOL * NOTIONAL);
```
Mistuus: Out of curiosity, why do we use ```TOL * NOTIONAL``` instead of just doing ```0.01d``` or having a constant ```DELTA = 0.01d```?

jodastephen: It makes the tolerance proportional to the trade size, converting it to be more like a percentage of the trade value rather than an absolute amount,

PR #1729, opened 2018-06-12
Refactor FxSingle

jodastephen:
Support case where the payment date differs between the currencies such as for AED and EGP.
Support old serialized form.

### DIFF
--- a/modules/product/src/test/java/com/opengamma/strata/product/fx/FxSingleTest.java
+++ b/modules/product/src/test/java/com/opengamma/strata/product/fx/FxSingleTest.java
@@ -20,6 +20,7 @@ import static org.testng.Assert.assertEquals;
 import java.time.LocalDate;
 import java.util.Optional;
 
+import org.joda.beans.ser.JodaBeanSer;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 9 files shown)


modules/product/src/test/java/com/opengamma/strata/product/fx/FxSingleTest.java
```diff
@@ -158,9 +189,8 @@ public void test_builder_rightOrder() {
 
   public void test_builder_switchOrder() {
     FxSingle test = FxSingle.meta().builder()
-        .set(FxSingle.meta().baseCurrencyAmount(), USD_M1600)
-        .set(FxSingle.meta().counterCurrencyAmount(), GBP_P1000)
-        .set(FxSingle.meta().paymentDate(), DATE_2015_06_30)
```
headphonejack: This hasn't changed from before, but I wouldn't mind an explanation of how we end up with the same baseCurrencyAmount when we appear to set it to something different!

jodastephen: The currency pair is GBP/USD, so the facyory method needs to swap these around to get GBP as the base and USD as the counter

PR #1732, opened 2018-06-12
Alter CSV format for FxSingle

jodastephen:
Provide a more market-focussed version of the CSV

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleTradeCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleTradeCsvLoader.java
@@ -61,62 +61,65 @@ class FxSingleTradeCsvLoader {
    * @return the parsed trade, as an instance of {@link FxSingleTrade}
    */
   static FxSingleTrade parse(CsvRow row, TradeInfo info, TradeCsvInfoResolver resolver) {
-    FxSingleTrade trade = parseRow(row, info, resolver);
+    FxSingleTrade trade = parseRow(row, info);
     return resolver.completeTrade(row, trade);
   }
 
-  /**
-   * Parses the data from a CSV row.
-   *
-   * @param row  the CSV row object
-   * @param info  the trade info object
-   * @param resolver  the resolver used to parse additional information. This is not currently used in this method.
-   * @return the parsed trade, as an instance of {@link FxSingleTrade}
-   */
-  private static FxSingleTrade parseRow(CsvRow row, TradeInfo info, TradeCsvInfoResolver resolver) {
-    Optional<BusinessDayAdjustment> paymentAdj = parsePaymentDateAdjustment(row);
+  // parses the trade
+  private static FxSingleTrade parseRow(CsvRow row, TradeInfo info) {
     if (row.findValue(BUY_SELL_FIELD).isPresent()) {
-      // convention-based
-      // ideally we'd use the trade date plus "period to start" to get the spot/payment date
-      // but we don't have all the data and it gets complicated in places like TRY, RUB and AED
-      CurrencyPair pair = CurrencyPair.parse(row.getValue(CONVENTION_FIELD));
-      BuySell buySell = LoaderUtils.parseBuySell(row.getValue(BUY_SELL_FIELD));
-      Currency currency = Currency.parse(row.getValue(CURRENCY_FIELD));
-      double notional = LoaderUtils.parseDouble(row.getValue(NOTIONAL_FIELD));
-      double fxRate = LoaderUtils.parseDouble(row.getValue(FX_RATE_FIELD));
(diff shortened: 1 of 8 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleTradeCsvLoader.java
```diff
@@ -61,43 +74,66 @@ static FxSingleTrade parse(CsvRow row, TradeInfo info, TradeCsvInfoResolver reso
...
+      BuySell buySell = LoaderUtils.parseBuySell(row.getValue(BUY_SELL_FIELD));
+      Currency currency = Currency.parse(row.getValue(CURRENCY_FIELD));
+      double notional = LoaderUtils.parseDouble(row.getValue(NOTIONAL_FIELD));
+      double fxRate = LoaderUtils.parseDouble(row.getValue(FX_RATE_FIELD));
+      LocalDate paymentDate = LoaderUtils.parseDate(row.getValue(PAYMENT_DATE_FIELD));
+      CurrencyAmount amount = CurrencyAmount.of(currency, buySell.normalize(notional));
+      FxSingle fx = paymentAdj
+          .map(adj -> FxSingle.of(amount, FxRate.of(pair, fxRate), paymentDate, adj))
+          .orElse(FxSingle.of(amount, FxRate.of(pair, fxRate), paymentDate));
+      return FxSingleTrade.of(info, fx);
```
Mistuus: Would it be better to extract this into its own method for readability?

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleTradeCsvLoader.java
```diff
@@ -61,43 +74,66 @@ static FxSingleTrade parse(CsvRow row, TradeInfo info, TradeCsvInfoResolver reso
...
+      CurrencyPair pair = CurrencyPair.parse(row.getValue(CONVENTION_FIELD));
+      BuySell buySell = LoaderUtils.parseBuySell(row.getValue(BUY_SELL_FIELD));
+      Currency currency = Currency.parse(row.getValue(CURRENCY_FIELD));
+      double notional = LoaderUtils.parseDouble(row.getValue(NOTIONAL_FIELD));
+      double fxRate = LoaderUtils.parseDouble(row.getValue(FX_RATE_FIELD));
+      LocalDate paymentDate = LoaderUtils.parseDate(row.getValue(PAYMENT_DATE_FIELD));
+      CurrencyAmount amount = CurrencyAmount.of(currency, buySell.normalize(notional));
+      FxSingle fx = paymentAdj
+          .map(adj -> FxSingle.of(amount, FxRate.of(pair, fxRate), paymentDate, adj))
+          .orElse(FxSingle.of(amount, FxRate.of(pair, fxRate), paymentDate));
```
Mistuus: Probably not a huge performance gain, but using ```orElseGet(...)``` would only get the value lazily instead of computing it every time.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleTradeCsvLoader.java
```diff
@@ -61,43 +74,66 @@ static FxSingleTrade parse(CsvRow row, TradeInfo info, TradeCsvInfoResolver reso
...
+            "Detected two legs having the same direction: {}, {}.",
+            direction1.toString(),
+            direction2.toString()));
+      }
+      Payment payment1 = Payment.of(currency1, direction1.normalize(notional1), paymentDate1);
+      Payment payment2 = Payment.of(currency2, direction2.normalize(notional2), paymentDate2);
+      FxSingle fx = paymentAdj
+          .map(adj -> FxSingle.of(payment1, payment2, adj))
+          .orElse(FxSingle.of(payment1, payment2));
+      return FxSingleTrade.of(info, fx);
```
Mistuus: Can this if branch be extracted in a method?

modules/product/src/main/java/com/opengamma/strata/product/fx/FxSwap.java
```diff
@@ -96,33 +97,34 @@ public static FxSwap of(FxSingle nearLeg, FxSingle farLeg) {
...
       FxRate nearRate,
-      double forwardPoints,
+      double decimalForwardPoints,
       LocalDate nearDate,
       LocalDate farDate) {
 
     Currency currency1 = amount.getCurrency();
     ArgChecker.isTrue(
         nearRate.getPair().contains(currency1),
         Messages.format("Amount and FX rate have a currency in common: {} and {}", amount, nearDate));
```
Mistuus: Message should probably be: "Amount and FX rate **must** have a currency in common: {} and {}", 
and shouldn't last parameter be **nearRate** rather than nearDate?

PR #1740, opened 2018-06-14
Simplify calculation tasks

jodastephen:


modules/calc/src/main/java/com/opengamma/strata/calc/runner/CalculationTasks.java
```diff
@@ -72,6 +72,8 @@
    * <p>
    * The targets will typically be trades.
    * The columns represent the measures to calculate.
+   * <p>
+   * Any target that implements {@link ResolvableCalculationTarget} will result in a failed task.
```
Mistuus: Is this because there won't be any reference data to resolve those calc targets?

jodastephen: Yes

PR #1769, opened 2018-07-17
 Faster holiday calendar loading

jodastephen:
Move data from GlobalHolidayCalendars into a binary file
Seems to save around 50ms on initialization
Also try to enhance `CsvRow`
Fix Javadoc


modules/basics/src/main/java/com/opengamma/strata/basics/date/GlobalHolidayCalendarLookup.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import com.google.common.collect.ImmutableMap;
+import com.opengamma.strata.collect.io.ResourceLocator;
+import com.opengamma.strata.collect.named.NamedLookup;
+
+/**
+ * Provide that loads common global holiday calendars from binary form on the classpath.
```
cjkent: Typo. Should it be "provider"?

PR #1778, opened 2018-08-15
Additional MapStream methods

wjnicholson:
FlatMap to another MapStream.
FlatMap keys.


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -282,6 +282,29 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+  //-------------------------------------------------------------------------
+  /**
+   * Transforms the entries in the stream by applying a mapper function to each key.
+   *
+   * @param mapper  a mapper function whose return values are included in the new stream
+   * @param <R>  the type of elements in the new stream
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <R> Stream<R> flatMapKeys(Function<? super K, Stream<? extends R>> mapper) {
+    return underlying.flatMap(e -> mapper.apply(e.getKey()));
```
jodastephen: I was expecting this to return a `MapStream`. This could just be `keys().flatMap()`, so doesn't gain enough.

wjnicholson: So you'd be expecting something similar to this?
```
  public <nK, nV> MapStream<nK, nV> flatMapKeys(Function<? super K, Stream<Pair<nK, nV>>> mapper) {
    return wrap(underlying
        .flatMap(e -> mapper.apply(e.getKey()))
        .map(pair -> entry(pair.getFirst(), pair.getSecond())));
  }
```

Or if Pair implemented Map.Entry then can remove the last .map call and somebody would be able to pass in a MapStream into the function

i.e
```
  public <nK, nV> MapStream<nK, nV> flatMapKeys(Function<? super K, Stream<Map.Entry<nK, nV>>> mapper) {
    return wrap(underlying.flatMap(e -> mapper.apply(e.getKey()));
  }

MapStream.of(map)
        .filterValues(this::someValueFilter)
        .flatMapKeys(key -> MapStream.of(nestedMaps.get(key)))
```

jodastephen: I'd expect `flatMapKeys` to take each key, flat map it, and associate the original value with each new key. `MapStream<List<String>, Integer>` ->  `MapStream<String, Integer>`.

I'm not sure that `flatMapValues` is feasible.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -282,6 +282,29 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
 
+  //-------------------------------------------------------------------------
+  /**
+   * Transforms the entries in the stream by applying a mapper function to each key.
+   *
+   * @param mapper  a mapper function whose return values are included in the new stream
+   * @param <R>  the type of elements in the new stream
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <R> Stream<R> flatMapKeys(Function<? super K, Stream<? extends R>> mapper) {
```
cjkent: The `mapKeys()` method has 2 overloads - one taking the key and another taking both the key and value. Is there an argument for doing the same here?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
cjkent: @wjnicholson what's the use case that inspired these?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
wjnicholson: As stephen mentioned above, not sure how worthwhile these methods are since .keys().flatMap() can be used.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
wjnicholson: The original reasoning was flattenening nested data structures (that contained nested maps), e.g
` MapStream<Key, Map<Key2, Map<Key3, Value>>> ` -> ` Map<Key, Value>`

wjnicholson: I also had in the back of my head a want for MapStream like semantics for a PairStream, in particular the ` Guavate.zip() ` and ` Guavate.zipWithIndex() ` methods. Since we currently have no way of transforming a Stream<Pair> into a MapStream.

wjnicholson: True, I guess I'm being pernickety about the verbosity with method references.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -282,6 +282,47 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+  /**
+   * Transforms the entries in the stream by applying a mapper function to each key and value to produce a stream of
+   * entries, and then flattening the resulting stream of streams.
+   *
+   * @param mapper  a mapper function whose return values are included in the new stream
+   * @param <nK>  the type of the new keys
+   * @param <nV>  the type of the new values
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <nK, nV> MapStream<nK, nV> flatMap(BiFunction<? super K, ? super V, Stream<Map.Entry<nK, nV>>> mapper) {
```
jodastephen: This method should return `Stream<R>` not `MapStream` for consistency with `map(BiFunction)`. It should also go after the other two methods for consistency.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -282,6 +282,47 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+   *
+   * @param mapper  a mapper function whose return values are the keys in the new stream
+   * @param <R>  the type of the new keys
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <R> MapStream<R, V> flatMapKeys(BiFunction<? super K, ? super V, Stream<R>> mapper) {
+    return wrap(underlying
+        .flatMap(e -> mapper.apply(e.getKey(), e.getValue()).map(newKey -> entry(newKey, e.getValue()))));
+  }
+
```
jodastephen: `flatMapValues()` does make sense after all, so those 2 methods will need adding.

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -282,6 +282,47 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+   * @return a stream containing the values returned from the mapper function
+   */
+  public <nK, nV> MapStream<nK, nV> flatMap(BiFunction<? super K, ? super V, Stream<Map.Entry<nK, nV>>> mapper) {
+    return wrap(underlying.flatMap(e -> mapper.apply(e.getKey(), e.getValue())));
+  }
+
+  /**
+   * Transforms the keys in the stream by applying a mapper function to each key.
+   * <p>
+   * The new keys produced will be associated with the original value.
```
jodastephen: Better docs, as it gets complex:

```
For example, this could convert a `MapStream<List<String>, LocalDate>` to a `MapStream<String, LocalDate>`
```

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
wjnicholson:
```
  /**
   * Transforms the values in the stream by applying a mapper function to each value.
   * <p>
   * The new values produced will be associated with the original key.
   * <p>
   * For example this could turn a {@code MapStream<LocalDate, List<String>>} into a
   * {@code MapStream<LocalDate, String>}
   *
   * @param mapper  a mapper function whose return values are the keys in the new stream
   * @param <R>  the type of the new keys
   * @return a stream containing the values returned from the mapper function
   */
  public <R> MapStream<K, R> flatMapValues(Function<? super V, Stream<R>> mapper) {
    return wrap(underlying.flatMap(e -> mapper.apply(e.getValue()).map(newValue -> entry(e.getKey(), newValue))));
  }
```

PR #1788, opened 2018-08-24
MapStream methods

wjnicholson:
* Sorting by keys and by values
* Maximum and minimum by keys and by values

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
@@ -40,7 +40,7 @@ import com.google.common.collect.Streams;
  * This stream wraps a {@code Stream&lt;Map.Entry&gt;}, providing convenient methods for
  * manipulating the keys and values. Unlike a {@code Map}, the keys in a {@code MapStream}
  * do not have to be unique, although certain methods will fail if they are not unique.
- * 
+ *
  * @param <K>  the key type
  * @param <V>  the value type
  */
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -386,6 +386,96 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+  }
+
+  //-----------------------------------------------------------------------
+  /**
+   * Finds the minimum entry in the stream by comparing the keys using the supplied comparator.
+   *
+   * @param comparator  a comparator of keys
+   * @return the minimum entry
+   */
+  public Optional<Map.Entry<K, V>> minKeys(Comparator<? super K> comparator) {
```
cjkent: Should this be called `minKey` (singular)?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -386,6 +386,96 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+    return underlying.min((e1, e2) -> comparator.compare(e1.getKey(), e2.getKey()));
+  }
+
+  /**
+   * Finds the minimum entry in the stream by comparing the values using the supplied comparator.
+   *
+   * @param comparator  a comparator of values
+   * @return the minimum entry
+   */
+  public Optional<Map.Entry<K, V>> minValues(Comparator<? super V> comparator) {
```
cjkent: `minValue`?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -386,6 +386,96 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+    return underlying.min((e1, e2) -> comparator.compare(e1.getValue(), e2.getValue()));
+  }
+
+  /**
+   * Finds the maximum entry in the stream by comparing the keys using the supplied comparator.
+   *
+   * @param comparator  a comparator of keys
+   * @return the maximum entry
+   */
+  public Optional<Map.Entry<K, V>> maxKeys(Comparator<? super K> comparator) {
```
cjkent: `maxKey`?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -386,6 +386,96 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+    return underlying.max((e1, e2) -> comparator.compare(e1.getKey(), e2.getKey()));
+  }
+
+  /**
+   * Finds the maximum entry in the stream by comparing the values using the supplied comparator.
+   *
+   * @param comparator  a comparator of values
+   * @return the maximum entry
+   */
+  public Optional<Map.Entry<K, V>> maxValues(Comparator<? super V> comparator) {
```
cjkent: `maxValue`?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
wjnicholson: I was thinking it would be minKeys because you're taking the minimum by comparing the keys, whereas minKey makes me think that it will return the minimum key i.e Optional<K>

Happy to be wrong.

jodastephen: I think the name is fine as is, I just struggle to think that these four min/max methods would actually be used.

wjnicholson: Ah, I wanted to use them, that's why I thought of them.

PR #1793, opened 2018-09-05
Add NumberFormatter

jodastephen:
Thread-safe wrapper around `DecimalFormat`/`NumberFormat`.
`ThreadLocal` is 2.5x faster than clone() for common cases.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/NumberFormatter.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/NumberFormatter.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright (C) 2018 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.collect;
+
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.text.NumberFormat;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/NumberFormatter.java
```diff
@@ -0,0 +1,199 @@
...
+
+  //-------------------------------------------------------------------------
+  /**
+   * Obtains a standard formatter configured by grouping and decimal places.
+   * <p>
+   * The formatter will have the specified number of decimal places.
+   * The integer part will be grouped if the flag is set.
+   * The decimal part will never be grouped or truncated.
+   * The implementation uses English locale data, which uses commas as a separator and a decimal point (dot).
+   * Numbers will be rounded using {@link RoundingMode#HALF_EVEN}
```
cjkent: Why `HALF_EVEN` and not `HALF_UP`? `HALF_UP` is the "standard" rounding mode everyone learns at school. From the Javadocs for `HALF_UP`

>Note that this is the rounding mode commonly taught at school.

jodastephen: It is what `DecimalFormat` uses.

cjkent: So it does. That surprises me.

PR #1798, opened 2018-10-01
Extend MapStream

jodastephen:
Provide ability to zip two streams into a `MapStream`

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -228,7 +228,7 @@ public final class Guavate {
    * @param zipper  the function used to combine the pair of objects
    * @return a stream of pairs, one from each stream
    */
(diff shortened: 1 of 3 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -228,7 +228,7 @@ public boolean hasNext() {
    * @param zipper  the function used to combine the pair of objects
    * @return a stream of pairs, one from each stream
    */
-  private static <A, B, R> Stream<R> zip(Stream<A> stream1, Stream<B> stream2, BiFunction<A, B, R> zipper) {
+  static <A, B, R> Stream<R> zip(Stream<A> stream1, Stream<B> stream2, BiFunction<A, B, R> zipper) {
     // this is private for now, to see if it is really needed on the API
```
wjnicholson: Does it make sense to leave this comment now that it's used elsewhere in the package?

jodastephen: I think so. It explains why it isn't public. And I suspect the new method may end up being a reason to never make it public.

PR #1800, opened 2018-10-01
Map based sensitivity

jodastephen:
Provide methods to convert sensitivity to and from a map


modules/market/src/main/java/com/opengamma/strata/market/param/CurrencyParameterSensitivity.java
```diff
@@ -398,6 +429,37 @@ public CurrencyAmount total() {
...
+   * 
+   * @return a map stream containing the parameter metadata and the sensitivity value
+   */
+  public MapStream<ParameterMetadata, Double> sensivities() {
+    return MapStream.zip(parameterMetadata.stream(), sensitivity.stream().boxed());
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Converts this instance to a stream of sensitivity, keyed by the identifier type.
```
cjkent: Javadoc is for the method above.

PR #1815, opened 2018-10-12
Allow a throwable to be the cause of a FailureItem

wjnicholson:
We don't depend on any exception specific behaviour so makes sense to be more general


jodastephen:
The design of this part of the code was made on the basis that `Throwable` should not be caught, only `Exception`. Were a change to be made it would affect more classes than just `FailureItem`, including `Failure` and `Result`. But at the moment I don't think we should make the change.

wjnicholson:
While I agree with you about not catching errors, it's frustrating that `CompletableFuture.completeExceptionally()` asks you to handle a throwable. It seems off to have to either cast (risking a class cast exception if it is an error and obscuring the actual problem) or to have to do instance of checks and potentially wrapping the error to pass to FailureItem or rethrowing the error.

PR #1824, opened 2018-10-16
Enhance MapStream

jodastephen:
Add `toMapGrouping` that provides more control over the terminal group

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
@@ -6,11 +6,17 @@
 package com.opengamma.strata.collect;
 
 import static com.opengamma.strata.collect.Guavate.entry;
+import static java.util.stream.Collectors.collectingAndThen;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
 
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Spliterator;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -556,25 +576,66 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
...
+  //-------------------------------------------------------------------------
+  /**
+   * Returns an immutable map built from the entries in the stream, grouping by key.
+   * <p>
+   * Entries are grouped based on the equality of the key.
+   * <p>
+   * This is a terminal operation.
+   *
+   * @return an immutable map built from the entries in the stream
+   * @throws IllegalArgumentException if the same key occurs more than once
```
wjnicholson: Leftover from a copy paste? I would be alarmed if this were the case

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -556,25 +576,66 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
    *   MapStream.concat(mapStreamA, mapStreamB).toMap((a,b) -> a);
    * </pre>
    * </p>
+   * <p>
```
wjnicholson: Why is there an extra `</p>` above this?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -229,7 +238,7 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
    * @return a stream of the values
    */
   public Stream<V> values() {
-    return underlying.map(e -> e.getValue());
+    return underlying.map(Entry::getValue);
```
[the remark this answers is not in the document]
jodastephen: Where there is no chance of it changing, method references can be clearer. It is in code that has more business logic that method references often end up annoyingly being converted back to lambdas.

Didn't add the comment to values as users should not expect them to be unique.

PR #1851, opened 2018-11-19
adding exclude logic for holidays to handle Chinese "long holiday" weekends shift

loveyoupeng:
in China, the weekends need to be shifted for country-wide "long holidays" such as Spring Festivals, adding exclude logic to adding capability to deal this requirement.


loveyoupeng:
need further change to ImmutableHolidayCalendar

PR #1853, opened 2018-11-20
adding exclude section in holiday data ini file to handle shifted weekends

loveyoupeng:
In China, the weekends could be shift/swaped with weekdays to form a "Golden Week", adding exclude section in the holiday data ini logic to handle similar cases


jodastephen:
Thanks for this PR. I've taken the work you've done and updated it in #1958. My goal was to add something that meets you use case but in a way that works fully in the context of the class. The "Excludes" was changed to "WorkingDays" to be more positive. The `ImmutableHolidayCalendar.combined()` method was also fixed.

Thanks again for the PR, and sorry it took a while to process.

PR #1864, opened 2019-01-09
ValueWithFailures.combining() function for Stream reductions

wjnicholson:


wjnicholson:
That PR Quality review is complaining about underscores in test names - the rest of the class is written in the same way

jodastephen:
I'm not sure I understand this one. We already have `vwf.combinedWith(vwf, concatToList())`. Do you need a variation of that? I'm not a big fan of the optional returning reduce on stream.

wjnicholson:
> I'm not sure I understand this one. We already have `vwf.combinedWith(vwf, concatToList())`. Do you need a variation of that? I'm not a big fan of the optional returning reduce on stream.

It reduces boilerplate for Stream reductions - writing 
`.reduce((vwf1, vwf2) -> vwf1.combinedWith(vwf2, (v1, v2) -> v1.doSomethingWith(v2, anotherArg)))`
 introduces a lot of noise around the interesting method call.

Also that reduce is useful if one has a default value that requires computation, do you tend away from `.orElseGet()`?

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -110,6 +111,21 @@
     }
   }
 
+  //-----------------------------------------------------------------------
+  /**
+   * Returns a BinaryOperator that combines ValueWithFailures objects using the provided combiner function.
+   * <p>
+   * Useful in cases where one has a {@code Stream<ValueWithFailures>} and wants to call {@code .reduce()}.
```
jodastephen: Change this to an example in a `<pre>` block.

modules/collect/src/main/java/com/opengamma/strata/collect/result/ValueWithFailures.java
```diff
@@ -110,6 +111,21 @@
     }
   }
 
+  //-----------------------------------------------------------------------
+  /**
+   * Returns a BinaryOperator that combines ValueWithFailures objects using the provided combiner function.
```
jodastephen: Use `{@code xxx}` for the two types

wjnicholson:
Something odd here - can't see the build running as a check

PR #1866, opened 2019-01-09
Add convenience methods for working with CompletableFuture<Void>

wjnicholson:
Ran into this with a `Stream<CompletableFuture<Void>>` and wanting to collect it.
If you use `Guavate.toCombinedFuture()` then it'll throw a `NullPointerException` when collecting the results into an `ImmutableList`

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -863,7 +863,7 @@ public final class Guavate {
    *
    * @return a collector that combines the input futures as single future
    */
-  public static Collector<CompletableFuture<Void>, ?, CompletableFuture<Void>> toCombinedVoidFuture() {
+  public static Collector<? extends CompletableFuture<Void>, ?, CompletableFuture<Void>> toCombinedVoidFuture() {
     return collectingAndThen(toImmutableList(), Guavate::combineFuturesAsVoid);
   }
 
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -9,6 +9,7 @@ import static java.util.stream.Collectors.collectingAndThen;
 
 import java.time.Duration;
 import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -831,6 +832,41 @@ public R next() {
...
+   * <p>
+   * If any input future completes exceptionally, the result will also complete exceptionally.
+   *
+   * @param futures the void futures, may be empty
+   * @return a void future that combines the input future
+   */
+  public static CompletableFuture<Void> combineFuturesAsVoid(List<? extends CompletableFuture<Void>> futures) {
+    int size = futures.size();
+    CompletableFuture[] futuresArray = futures.toArray(new CompletableFuture[size]);
+    return CompletableFuture.allOf(futuresArray);
```
jodastephen: Could this not just be the default implementation? I don't see anything in the signature that requires ImmutableList

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -967,7 +1004,7 @@ public static ThreadFactoryBuilder namedThreadFactory(String threadNamePrefix) {
    * <p>
    * This takes an argument which is the number of stack levels to look back.
    * This will be 2 to return the caller of this method, 3 to return the caller of the caller, and so on.
-   * 
+   *
```
jodastephen: Really need an IntelliJ setting to avoid this

wjnicholson: It's frustrating - I tried to add the space and commit it back, but intellij refused to see it as a  change it could commit!

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
wjnicholson: So change `combineFuturesAsList()` to allow null returned values, and get rid of this?
Could make sense, you'd be building up a list of `null` simply to discard it, but that's not the end of the world.

wjnicholson: Could that reasoning apply to changing `combineFuturesAsMap()`? Not sure it translates as well, having null values in a map I guess is a kind of failure, and if you'd want to wait on the results you could simply collect the values using the `combineFuturesAsList()`

PR #1867, opened 2019-01-14
Add Guavate filtering/casting

jodastephen:
Helper methods for streams that allow filtering and casting by type

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -21,7 +21,6 @@ import java.util.Spliterators;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.BinaryOperator;
@@ -288,9 +287,9 @@ public final class Guavate {
    * or {@link Stream#findAny()}, this approach ensures an exception is thrown if there
    * is more than one element in the stream.
    * <p>
-   * This would be used as follows:
+   * This would be used as follows (with a static import):
    * <pre>
-   *   stream.filter(...).reduce(Guavate.ensureOnlyOne()).get();
+   *   stream.filter(...).reduce(ensureOnlyOne()).get();
    * </pre>
    *
    * @param <T>  the type of element in the stream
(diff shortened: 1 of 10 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -303,6 +302,58 @@ public R next() {
...
+   * <pre>
+   *   stream.map(Foo.class::cast);
+   * </pre>
+   *
+   * @param <T>  the type of element in the input stream
+   * @param <R>  the type of element in the output stream
+   * @param cls  the type of element in the output stream
+   * @return the function
+   */
+  public static <T, R extends T> Function<T, R> casting(Class<R> cls) {
```
wjnicholson: What's the benefit of using this over the `Foo.class::cast` style?

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -303,6 +302,58 @@ public R next() {
...
+   * to filter elements in a stream to a particular type.
+   * <p>
+   * This would be used as follows (with a static import):
+   * <pre>
+   *   stream.flatMap(filtering(Foo.class));
+   * </pre>
+   * <p>
+   * This replaces code of the form:
+   * <pre>
+   *   stream.filtering(Foo.class::isInstance).map(Foo.class::cast);
```
wjnicholson:
```suggestion
   *   stream.filter(Foo.class::isInstance).map(Foo.class::cast);
```

modules/collect/src/test/java/com/opengamma/strata/collect/GuavateTest.java
```diff
@@ -180,6 +181,26 @@ public void test_ensureOnlyOne() {
     assertThrowsIllegalArg(() -> Stream.of("a", "b").reduce(Guavate.ensureOnlyOne()));
   }
 
+  //-------------------------------------------------------------------------
+  public void test_casting() {
+    assertEquals(Stream.empty().map(Guavate.casting(Integer.class)).collect(toList()), ImmutableList.of());
+    List<Number> baseList = Arrays.asList(1, 2, 3);
```
wjnicholson: Why is `Arrays.asList()` popping up here?

modules/collect/src/test/java/com/opengamma/strata/collect/GuavateTest.java
```diff
@@ -180,6 +181,26 @@ public void test_ensureOnlyOne() {
...
+    List<Integer> castList = baseList.stream().map(Guavate.casting(Integer.class)).collect(toList());
+    assertEquals(castList, ImmutableList.of(1, 2, 3));
```
wjnicholson: I'd prefer to have the assertion be `assertEquals(castList, baseList)` rather than writing out the contents of `baseList`  but happy to leave it.

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -303,6 +302,58 @@ public R next() {
...
+   * </pre>
+   * <p>
+   * This replaces code of the form:
+   * <pre>
+   *   stream.map(Foo.class::cast);
+   * </pre>
+   *
+   * @param <T>  the type of element in the input stream
+   * @param <R>  the type of element in the output stream
+   * @param cls  the type of element in the output stream
```
wjnicholson: Weird seeing that repeated twice, although I guess it's correct?

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
jodastephen: Teaching people to use methods in Guavate seems to be more successful than teaching them this particular method reference approach. (Many don't appreciate that the LHS of a method ref can be an expression)

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
jodastephen: Yep, writing docs for generics is a pain

PR #1868, opened 2019-01-14
Add Guavate filteringOptional

jodastephen:
Helper method for streams that allow filtering of optionals


wjnicholson:
What does this add over `Guavate.stream(Optional<T>)` ?

jodastephen:
A name suitable for use in a stream flatMap.

PR #1875, opened 2019-01-21
Additional ValueWithFailures methods

jodastephen:
Provide more control over the value and failures.
Avoid the need to use `map` and `flatMap` in certain circumstances.

I don't love the method name, but other names didn't seem that much better.


jodastephen:
Copy of #1872

PR #1879, opened 2019-01-25
Historic index support

jodastephen:
Provide a mechanism to support historic indices where the time-series
needs to be queried but the forward curve is not present


modules/measure/src/main/java/com/opengamma/strata/measure/rate/DefaultLookupRatesProvider.java
```diff
@@ -207,32 +210,56 @@ public FxForwardRates fxForwardRates(CurrencyPair currencyPair) {
...
+      return historicCurve(index);
+    }
+    return IborIndexRates.of(index, getValuationDate(), marketData.getValue(curveId), timeSeries(index));
+  }
+
+  // creates a historic rates instance if index is inactive and time-series is available
+  private IborIndexRates historicCurve(IborIndex index) {
+    LocalDateDoubleTimeSeries fixings = timeSeries(index);
+    if (index.isActive() || fixings.isEmpty()) {
       throw new MarketDataNotFoundException(lookup.msgIndexNotFound(index));
```
brianweller89: Not sure this is MarketDataNotFoundException in the case where index is active. it's more like MarketDataFoundException!

jodastephen: This replicates the original exception. It occurs when the market data (curve) is not found for an active index.

PR #1893, opened 2019-02-18
Add default mapping for Thai Baht and Taiwan Dollar holiday calendars

wjnicholson:


modules/basics/src/main/resources/META-INF/com/opengamma/strata/config/base/HolidayCalendarDefaultData.ini
```diff
@@ -31,3 +31,6 @@ HKD = HKHK
 INR = INMU
 KRW = KRSE
 SGD = SGSI
+
```
jodastephen: No blank line :-)

PR #1902, opened 2019-03-05
New cash settlement method for swaptions: Collateralized cash price

marc-henrard:
Method is now standard convention for EUR swaptions


modules/product/src/main/java/com/opengamma/strata/product/swaption/CashSwaptionSettlementMethod.java
```diff
@@ -35,7 +35,14 @@
    * <p>
    * The settlement amount is computed with the discount factor based on the agreed zero coupon curve.
    */
-  ZERO_COUPON_YIELD;
+  ZERO_COUPON_YIELD,
+  /**
+   * The Collateralized Cash Price
+   * <p>
+   * The settlement amount is computed  with the discount factor based where the parties are deemed to have a 
+   * bilateral, zero-threshold ISDA CSA. (ISDA definition 18.3 (g)
```
jodastephen: Can we use the format "Defined by the 2006 ISDA definitions 4.16a."

PR #1912, opened 2019-03-13
Add businessDayAdjustment default for AccrualSchedule

Mistuus:
This change is driven by the finding that when we parse a Strata Trade from CSV format without specifying a leg 'Date Calendar' and 'Date Convention', it will create a Strata trade with an AccrualSchedule with BusinessDayAdjustment.NONE. 

This leads to computing sensitivities/IM by (potentially) taking into consideration days which fall on a weekend (i.e. non-business days). A consequence of this is that IM numbers might not match CCP IM numbers.


Mistuus:
There are also a few places where we use the below (i.e. defaulting to NONE). Not sure if we need to use a more sensible default like ModFollowing w/ default currency calendar. 

```java
BusinessDayAdjustment bda = parseBusinessDayAdjustment(row, leg, cnvField, calField)
        .orElse(BusinessDayAdjustment.NONE);
```

modules/loader/src/test/java/com/opengamma/strata/loader/csv/TradeCsvLoaderTest.java
```diff
@@ -599,7 +600,9 @@ public void test_load_swap_full5() {
                     .startDate(date(2017, 8, 1))
                     .endDate(date(2022, 8, 1))
                     .frequency(Frequency.P3M)
-                    .businessDayAdjustment(BusinessDayAdjustment.NONE)
+                    .businessDayAdjustment(BusinessDayAdjustment.of(
```
jodastephen: Is it still possible to get `NONE`? Is there a test for that? I suspect the code doesn't allow for it.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -660,6 +663,23 @@ private static PriceIndexCalculationMethod parseInflationMethod(Optional<String>
         .map(cal -> BusinessDayAdjustment.of(dateCnv, cal));
   }
 
+  // days adjustment, defaulting business day convention and holiday calendar
+  private static BusinessDayAdjustment parseBusinessDayAdjustmentWithFallback(
```
jodastephen: Probably should be `WithDefault` because "fallback" has a different meaning within this class.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -660,6 +663,23 @@ private static PriceIndexCalculationMethod parseInflationMethod(Optional<String>
...
+      String cnvField,
+      String calField) {
+
+    Currency currency = Currency.of(getValueWithFallback(row, leg, CURRENCY_FIELD));
+    BusinessDayConvention dateCnv = findValue(row, leg, cnvField)
+        .map(s -> LoaderUtils.parseBusinessDayConvention(s))
+        .orElse(BusinessDayConventions.MODIFIED_FOLLOWING);
+    HolidayCalendarId holidayCalendarId = findValue(row, leg, calField)
+        .map(s -> HolidayCalendarId.of(s))
+        .orElseGet(() -> HolidayCalendarId.defaultByCurrency(currency));
```
jodastephen: These defaults are not documented anywhere AFAIK. eg. [here](http://strata.opengamma.io/holiday_data/) or [here](http://strata.opengamma.io/trade_loader/)

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -267,8 +267,11 @@ private static PeriodicSchedule parseAccrualSchedule(CsvRow row, String leg) {
     builder.endDate(LoaderUtils.parseDate(getValueWithFallback(row, leg, END_DATE_FIELD)));
     builder.frequency(Frequency.parse(getValue(row, leg, FREQUENCY_FIELD)));
     // adjustments
-    BusinessDayAdjustment dateAdj = parseBusinessDayAdjustment(row, leg, DATE_ADJ_CNV_FIELD, DATE_ADJ_CAL_FIELD)
-        .orElse(BusinessDayAdjustment.NONE);
+    BusinessDayAdjustment dateAdj = parseBusinessDayAdjustmentWithFallback(
```
jodastephen: Would it not be necessary to update all uses of `parseBusinessDayAdjustment`, not just this one?

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
```diff
@@ -660,6 +663,23 @@ private static PriceIndexCalculationMethod parseInflationMethod(Optional<String>
...
   }
 
+  // days adjustment, defaulting business day convention and holiday calendar
+  private static BusinessDayAdjustment parseBusinessDayAdjustmentWithFallback(
+      CsvRow row,
+      String leg,
+      String cnvField,
+      String calField) {
+
+    Currency currency = Currency.of(getValueWithFallback(row, leg, CURRENCY_FIELD));
```
jodastephen: I don't think this can result in `BusinessDayAdjustment.NONE`. If both fields are empty, `NONE` should be the result.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
Mistuus: How does one modify the documentation?

jodastephen: https://github.com/OpenGamma/StrataDocs

modules/loader/src/test/java/com/opengamma/strata/loader/csv/TradeCsvLoaderTest.java
Mistuus: No longer possible to get NONE for BusinessDayAdjustment. There should always be a default from now on.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
Mistuus: Removed this method completely.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvLoader.java
Mistuus: We should preserve the logic for the other ones: startDate & endDate.

Mistuus:
This last commit made the following changes:
- `businessDayAdjustment` for `accrualSchedule` will always default to ModFollowing with the default currency holiday calendar => TODO: Document in strata docs
-  Always provide a default `businessDayAdjustment` (as specified above) when a days offset is specified without any businessDayAdjustment information => TODO: Document in strata docs 
- Same as above for any adjustable date => TODO: Document in strata docs

Mistuus:
Documentation PR raised: https://github.com/OpenGamma/StrataDocs/pull/48

Mistuus:
Closed as more work than initially thought is required.

PR #1917, opened 2019-03-18
Add LongArray

cjkent:
Adds a `LongArray` class that is the equivalent of `IntArray` and `DoubleArray` for `long` values.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/array/LongArray.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/array/LongArray.java
@@ -278,7 +278,7 @@ public final class LongArray
 
   //-----------------------------------------------------------------------
   /**
-   * Obtains an instance from a collection of {@code Integer}.
+   * Obtains an instance from a collection of {@code Long}.
    * <p>
    * The order of the values in the returned array is the order in which elements are returned
    * from the iterator of the collection.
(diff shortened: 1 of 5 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/array/LongArray.java
```diff
@@ -0,0 +1,1297 @@
...
+
+  /**
+   * Returns an instance with each value divided by the specified divisor.
+   * <p>
+   * This is used to divide the contents of this array, returning a new array.
+   * <p>
+   * This is a special case of {@link #map(LongUnaryOperator)}.
+   * This instance is immutable and unaffected by this method.
+   *
+   * @param divisor  the value by which the array applyAsLong() divided
```
jodastephen: Should be "array values are divided"

modules/collect/src/main/java/com/opengamma/strata/collect/array/LongArray.java
```diff
@@ -0,0 +1,1297 @@
...
+   * <pre>
+   *   result = base.mapWithIndex((index, value) -&gt; index * value);
+   * </pre>
+   * <p>
+   * This instance is immutable and unaffected by this method.
+   *
+   * @param function  the function to be applied
+   * @return a copy of this array with the operator applied to the original values
+   */
+  public LongArray mapWithIndex(LongBinaryOperator function) {
```
jodastephen: Wrong type. Should be `IntLongToLongFunction` or similar

modules/collect/src/main/java/com/opengamma/strata/collect/function/IntLongConsumer.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package com.opengamma.strata.collect.function;
+
+/**
+ * An operation consuming two arguments - {@code int} and {@code long}.
+ * <p>
+ * Implementations of this interface will operate using side-effects.
+ */
+@FunctionalInterface
+public interface IntLongConsumer {
```
jodastephen: Worth checking if these functional interfaces are the only ones you are going to need for `long`.

modules/collect/src/main/java/com/opengamma/strata/collect/array/LongArray.java
cjkent: Oops. I hand a find-and-replace operation that went rogue. I thought I'd fixed everything but missed this one.

PR #1927, opened 2019-03-25
Parse bullet payments from CSV

jodastephen:
Additional support in CSV trade parser

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/BulletPaymentTradeCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/BulletPaymentTradeCsvLoader.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2019 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.loader.csv;
+
+import static com.opengamma.strata.basics.date.BusinessDayConventions.FOLLOWING;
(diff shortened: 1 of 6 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/BulletPaymentTradeCsvLoader.java
```diff
@@ -0,0 +1,67 @@
...
+    double notional = LoaderUtils.parseDouble(row.getValue(NOTIONAL_FIELD));
+    LocalDate paymentDate = LoaderUtils.parseDate(row.getValue(PAYMENT_DATE_FIELD));
+    BusinessDayAdjustment paymentAdj = FxSingleTradeCsvLoader.parsePaymentDateAdjustment(row)
+        .orElseGet(() -> BusinessDayAdjustment.of(FOLLOWING, HolidayCalendarId.defaultByCurrency(currency)));
+    CurrencyAmount amount = CurrencyAmount.of(currency, notional);
+    BulletPayment.Builder builder = BulletPayment.builder()
+        .payReceive(payReceive)
+        .value(amount)
+        .date(AdjustableDate.of(paymentDate, paymentAdj));
+    return BulletPaymentTrade.of(info, builder.build());
```
wjnicholson: Why not call `.build()` on the previous line?

jodastephen: It is copied from other code, which probably had logic between those two lines.

PR #1934, opened 2019-04-01
Output CSV using headers

jodastephen:
Provide a more convenient way to output CSV files using headers.

Note that this change includes some mutability (!), but that is fairly typical of IO code.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
@@ -185,7 +185,9 @@ public final class CsvOutput {
    * cells to be written by header name.
    * <p>
    * This writes the header line to the underlying.
-   * The returned instance should then be used to write additional rows.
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
```diff
@@ -171,6 +179,26 @@ private CsvOutput(Appendable underlying, String newLine, String separator, boole
...
+   * <p>
+   * This writes the header line to the underlying.
+   * The returned instance should then be used to write additional rows.
+   *
+   * @param headers  the list of headers
+   * @param alwaysQuote  when true, each column will be quoted, when false, quoting is selective
+   * @return the writer to use from now on
+   * @throws UncheckedIOException if an IO exception occurs
+   */
+  public CsvRowOutputWithHeaders withHeaders(List<String> headers, boolean alwaysQuote) {
```
wjnicholson: Does this need some documentation saying that it needs to be called before any other outputting method, and should only be called once?
Or can we assume the best from users?

PR #1942, opened 2019-04-10
Add ? extends to generics in MapStream#flatMap methods

wjnicholson:
Also change from using `BiFunction` to using `BiPredicate` on `filter`

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
@@ -251,8 +251,8 @@ public final class MapStream<K, V>
    * @param predicate  a predicate function applied to each key and value in the stream
    * @return a stream including the entries for which the predicate function returned true
    */
-  public MapStream<K, V> filter(BiFunction<? super K, ? super V, Boolean> predicate) {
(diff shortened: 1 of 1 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -383,7 +383,7 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
    * @param <R>  the type of the new keys
    * @return a stream of entries with new keys from the mapper function assigned to the values
    */
-  public <R> MapStream<R, V> flatMapKeys(Function<? super K, Stream<R>> mapper) {
+  public <R> MapStream<R, V> flatMapKeys(Function<? super K, Stream<? extends R>> mapper) {
```
jodastephen: The JDK uses `<R> Stream<R> flatMap(Function<? super T, ? extends Stream<? extends R>> mapper);` (one extra `? extends`)

My concern is whether adding the extra degree of freedom actually makes the job of the compiler harder, such that it might not resolve the desired type at all. Is there a particular use case where it is valuable to have the resulting stream be a supertype of the type of the stream returned from the function?

PR #1976, opened 2019-05-21
Synthetic calibrator extended to support Cross-currency swaps.

marc-henrard:
Synthetic calibrator did not support the cross-currency instruments (FX swaps and XCcy swaps) because the way the FxRateId was treated. Code has been improved to deal with QuoteId and FxRateId separetely.


modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/SyntheticRatesCurveCalibrator.java
```diff
@@ -158,37 +160,44 @@ public ImmutableMarketData marketData(
...
-        mapIdSy.put(k, mq);
-      }
-    }
     // Generate quotes for FX pairs. The first currency is arbitrarily selected as starting point. 
     // The crosses are automatically generated by the MarketDataFxRateProvider used in calibration.
     for (int loopccy = 1; loopccy < ccyRequired.size(); loopccy++) {
       CurrencyPair ccyPair = CurrencyPair.of(ccyRequired.get(0), ccyRequired.get(loopccy));
       FxRateId fxId = FxRateId.of(ccyPair);
       mapIdSy.put(fxId, FxRate.of(ccyPair, inputProvider.fxRate(ccyPair)));
+      mapId0.put(fxId, FxRate.of(ccyPair, inputProvider.fxRate(ccyPair)));  // Required to generate the trade
```
yukiiwashita: Because `MarketData` created this map is used in `CurveNode` for cross currency products without `MarketDataFxRateProvider`, FX crosses are not generated. Thus the exact `FxRateId` is required.

PR #1980, opened 2019-05-23
Adding two methods on Result: ifSuccess and ifFailure

andreiruse:
The consumer in `ifSuccess` will receive the value of the Result. It will only execute if the Result is successful.

The BiConsumer in `ifFailure` will receive both the value and the failure of the Result. It will only execute if the Result has a failure.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
@@ -10,7 +10,9 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
```diff
@@ -583,6 +585,19 @@ public boolean isSuccess() {
     return value != null;
   }
 
+  /**
+   * Executes the given consumer if the result represents a successful call and has a result available.
+   *
+   * This is the opposite of {@link #ifFailure(BiConsumer)}.
```
jodastephen: No need for this - it is self explanatory. And if it were here, it would need `<p>` above it.

modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
```diff
@@ -594,6 +609,19 @@ public boolean isFailure() {
     return failure != null;
   }
 
+  /**
+   * Executes the given consumer if the result represents a failure.
+   *
+   * This is the opposite of {@link #ifSuccess(Consumer)}.
```
jodastephen: No need for this - it is self explanatory. And if it were here, it would need `<p>` above it.

modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
```diff
@@ -594,6 +609,19 @@ public boolean isFailure() {
...
+   *
+   * @param consumer the consumer to be decorated
+   */
+  public void ifFailure(BiConsumer<? super T, Failure> consumer) {
```
jodastephen: This should just be a `Consumer` as value will always be null.

PR #1982, opened 2019-05-24
Check if valid FpML document

jodastephen:
Provide method to check if document is FpML.
Build on top of filtered parse method in `XmlFile`.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
@@ -9,6 +9,7 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.ToIntFunction;
 
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLStreamConstants;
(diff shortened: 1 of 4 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -101,6 +102,35 @@ public static XmlFile of(ByteSource source, String refAttrName) {
     });
   }
 
+  /**
+   * Parses the element names and structure from the specified XML, filtering to reduce memory usage.
+   * <p>
+   * This parses the specified byte source expecting an XML file format.
+   * The filter function should return the number of levels below the current one
```
headphonejack: `The filter function operates on the element name and...` maybe?

modules/collect/src/main/java/com/opengamma/strata/collect/io/XmlFile.java
```diff
@@ -101,6 +102,35 @@ public static XmlFile of(ByteSource source, String refAttrName) {
...
+   * that should be included in the response.
+   * Always returning {@code Integer.MAX_VALUE} will not filter the children.
```
headphonejack: So the range of `Integer.MIN_VALUE <= x < 0` is invalid values? Talking of which `Integer.MIN_VALUE` could cause interesting behaviour underflowing back to MAX_VALUE, which would remain for all children, and have the same effect as MAX_VALUE!

modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
```diff
@@ -175,6 +201,33 @@ private FpmlDocumentParser(
...
 
   //-------------------------------------------------------------------------
+  /**
+   * Checks if the source can be parsed as FpML.
+   * <p>
+   * This parses the specified byte source to determine if it appears to be FpML.
+   * <p>
+   * Sometimes, the FpML document is embedded in a non-FpML wrapper.
+   * This method will intelligently find the FpML document at the root or within one or two levels
+   * of wrapper by searching for an element that contains both {@code <trade>} and {@code <party>}.
```
headphonejack: `of wrapper` missing a word?

modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
```diff
@@ -216,7 +272,7 @@ private static XmlElement findFpmlRoot(XmlElement root) {
         }
       }
     }
-    throw new FpmlParseException("Unable to find FpML root element");
+    return null;
```
headphonejack: Is this a deliberate decision to avoid `Optional` for performance?

jodastephen: There is no need to use `Optional` between two private methods of the same class. Sometimes it may make sense for other reasons, but didn't seem to do so here.

PR #1984, opened 2019-05-29
Add helper methods to CsvOutput to bridge writing CSV Files and Iterators

ashleyheath:
Sometimes you already have an instance of a CsvFile or CsvIterator that you want to write out somewhere. This PR adds some static helper methods to CsvOutput to help support that.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
@@ -223,22 +223,6 @@ public final class CsvOutput {
   }
 
   /**
-   * Writes multiple {@code CsvRow}s to the underlying.
-   * <p>
-   * The boolean flag controls whether each entry is always quoted or only quoted when necessary.
-   *
-   * @param rows  the rows to write
-   * @param alwaysQuote  when true, each column will be quoted, when false, quoting is selective
-   * @throws UncheckedIOException if an IO exception occurs
-   */
-  public void writeRows(Iterable<CsvRow> rows, boolean alwaysQuote) {
-    ArgChecker.notNull(rows, "rows");
-    for (CsvRow row : rows) {
-      writeRow(row, alwaysQuote);
-    }
-  }
-
-  /**
    * Writes a single CSV line to the underlying, only quoting if needed.
    * <p>
    * This can be used as a method reference from a {@code Stream} pipeline from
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
```diff
@@ -222,6 +434,22 @@ public void writeLines(Iterable<? extends List<String>> lines, boolean alwaysQuo
...
+  /**
+   * Writes multiple {@code CsvRow}s to the underlying.
+   * <p>
+   * The boolean flag controls whether each entry is always quoted or only quoted when necessary.
+   *
+   * @param rows  the rows to write
+   * @param alwaysQuote  when true, each column will be quoted, when false, quoting is selective
+   * @throws UncheckedIOException if an IO exception occurs
+   */
+  public void writeRows(Iterable<? extends CsvRow> rows, boolean alwaysQuote) {
```
jodastephen: * Method should be next to `writeRow` (after `writeLine`)
* `CsvRow` is final, so `? extends` won't add much.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
```diff
@@ -174,6 +174,218 @@ public static CsvOutput safe(Appendable underlying, String newLine, String separ
...
+  /**
+   * Writes the provided {@code CsvFile} to the underlying, using the system default line separator and using a
+   * comma separator.
+   * <p>
+   * See the standard quoting rules in the class-level documentation.
+   *
+   * @param file  the file whose contents to output
+   * @param underlying  the destination to write to
+   */
+  public static void writeStandard(CsvFile file, Appendable underlying) {
```
jodastephen: I think instance methods would be the better choice, as 12 methods should reduce to 2.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
ashleyheath: Keep writeLine_s_ and writeRow_s_ next to each other?

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
ashleyheath: Add an 'alwaysQuote' param to each (for consistency) and make that four?

ashleyheath: Actually, ignore that, safe/standard is already on the instance, isn't it.

modules/collect/src/main/java/com/opengamma/strata/collect/io/CsvOutput.java
```diff
@@ -222,6 +222,22 @@ public void writeLines(Iterable<? extends List<String>> lines, boolean alwaysQuo
     }
   }
 
+  /**
```
jodastephen: Move method below lines methods. ie. group all the lines methods, then all the rows methods, then all the cell methods, then the file/iterator ones

ashleyheath: Done.

PR #1988, opened 2019-05-30
Utility combining methods on tuples

wjnicholson:
Useful for when you want to stream and combine `Collection<Pair<?>>` or `Collection<Triple<?>>`
Would appreciate opinions from @cjkent @jodastephen (appeal to authority)

Just realised I missed tests, woopsy


wjnicholson:
> Not overly convinced this will make for terribly readable code. But its probably better than doing this manually.

Agreed. 
It's useful for a single convenience call, using it more than once is a sign you need a proper class.

PR #1995, opened 2019-06-04
Pluggable trade CSV parsing

jodastephen:
Allow additional product types to be parsed
Allow the existing parsers to be replaced

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvLoader.java
@@ -14,6 +14,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharSource;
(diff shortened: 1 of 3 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvLoader.java
```diff
@@ -510,32 +512,39 @@ public boolean isKnownFormat(CharSource charSource) {
...
             break;
           default:
-            failures.add(FailureItem.of(
-                FailureReason.PARSING,
-                "CSV file trade type '{tradeType}' is not known at line {lineNumber}",
-                typeRaw,
-                row.lineNumber()));
+            Optional<Trade> parsedOpt = resolver.parseOtherTrade(typeUpper, row, info);
+            if (parsedOpt.isPresent()) {
+              if (tradeType.isInstance(parsedOpt.get())) {
```
wjnicholson: If it isn't an instance should this add a failure?

jodastephen: No, the type here is a filter, as per the other similar lines above,

PR #2004, opened 2019-06-19
Add Attributes.hasAttribute() shorthand method

ashleyheath:
Add a simple helper method to Attributes - `thing.hasAttribute(foo)` reads more clearly than `thing.findAttribute(foo).isPresent()` and is also nicer to use in functional code.

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/Attributes.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/Attributes.java
@@ -68,7 +68,7 @@ public interface Attributes {
    * @param type  the type to find
    * @return true if a matching attribute is present
    */
-  public default <T> boolean hasAttribute(AttributeType<T> type) {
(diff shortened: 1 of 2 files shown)


modules/product/src/main/java/com/opengamma/strata/product/Attributes.java
```diff
@@ -61,6 +61,17 @@ public static Attributes empty() {
...
   }
 
+  /**
+   * Determines if an attribute associated with the specified type is present.
+   *
+   * @param <T>  the type of the attribute value
+   * @param type  the type to find
+   * @return true if a matching attribute is present
+   */
+  public default <T> boolean hasAttribute(AttributeType<T> type) {
```
jodastephen: Looking around, we tend to use `containsXxx()` for methods like this.

ashleyheath: Updated.

PR #2007, opened 2019-06-20
Remove TestNG from TestHelper

jodastephen:
Allows the class to be used from tests in either JUnit or TestNG.
Without this TestNG and JUnit are both seen by maven when running tests.

### DIFF
--- a/modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
+++ b/modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
@@ -5,11 +5,6 @@
  */
 package com.opengamma.strata.collect;
 
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertNotNull;
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
-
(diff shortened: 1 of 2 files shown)


modules/collect/src/test/java/com/opengamma/strata/collect/TestHelper.java
```diff
@@ -521,14 +518,14 @@ public void flush() {
    */
   public static void assertUtilityClass(Class<?> clazz) {
     assertNotNull(clazz, "assertUtilityClass() called with null class");
-    assertTrue(Modifier.isFinal(clazz.getModifiers()), "Utility class must be final");
+    assertEquals(Modifier.isFinal(clazz.getModifiers()), true, "Utility class must be final");
```
cjkent: Is it worth defining `assertTrue`? It's such a standard method. `assertEquals(..., true)` seems a bit odd.

jodastephen: These are private methods, and since they are only used in a couple of places, it didn't seem worthwhile.

PR #2009, opened 2019-06-26
Optimise CurrencyAmount#parse

wjnicholson:
Using Splitter requires us to traverse the string twice, but since we know the format should always be an ISO 3 letter code then a double we can take a shortcut and split ourselves and then let Currency.parse() or Double.parse() throw if they have any issues.

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/CurrencyAmount.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/CurrencyAmount.java
@@ -50,6 +50,9 @@ public final class CurrencyAmount
   /** Serialization version. */
   private static final long serialVersionUID = 1L;
 
+  /** Splitter on single space for parsing from string. */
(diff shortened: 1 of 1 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/currency/CurrencyAmount.java
```diff
@@ -117,14 +115,19 @@ public static CurrencyAmount of(String currencyCode, double amount) {
    */
   @FromString
   public static CurrencyAmount parse(String amountStr) {
-    ArgChecker.notNull(amountStr, "amountStr");
-    List<String> split = Splitter.on(' ').splitToList(amountStr);
-    if (split.size() != 2) {
+    if (amountStr == null || amountStr.length() <= 4 || amountStr.charAt(3) != ' ') {
```
jodastephen: Add a comment to indicate that the method has been optimised, eg `// this method has had some performance optimisations applied`

PR #2012, opened 2019-06-27
Handle observable source when creating lookups

### LINKED ISSUE #2011: RatesMarketDataLookup constructs CurveIds with no observable source
The method `RatesMarketDataLookup.of(RatesCurveGroupDefinition, ObservableSource, FxRateLookup)` doesn't use the `ObservableSource` when constructing the `CurveId` instances.

The result of this behaviour is that `DefaultLookupRatesProvider.findData()` always returns the curve with the default observable source if there are multiple curves in the market data with the same name but different observable sources.

cjkent:
Use the `ObservableSource` when creating `CurveId` instance in `RatesMarketDataLookup`. This fixes a bug where curves with the default `ObservableSource` are returned from a rates provider that uses a lookup with a non-default source.

Fixes #2011


modules/measure/src/test/java/com/opengamma/strata/measure/rate/DefaultRatesMarketDataLookupTest.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (C) 2019 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+
```
jodastephen: No blank line after license

PR #2017, opened 2019-07-01
Added TenorTenorParameterMetadata

barrell12:


modules/market/src/main/java/com/opengamma/strata/market/param/TenorTenorParameterMetadata.java
```diff
@@ -0,0 +1,492 @@
...
+
+import com.opengamma.strata.basics.date.Tenor;
+import com.opengamma.strata.collect.tuple.Pair;
+
+/**
+ * Parameter metadata based on an expiry tenor, an underlying tenor and their respective year fractions.
+ */
+@BeanDefinition(builderScope = "private")
+public final class TenorTenorParameterMetadata
+    implements ParameterMetadata, ImmutableBean, Serializable {
```
wjnicholson: Maybe implement TenoredParameterMetadata? Although perhaps @jodastephen can give more insight into that

modules/market/src/main/java/com/opengamma/strata/market/param/TenorTenorParameterMetadata.java
```diff
@@ -0,0 +1,492 @@
...
+
+  /**
+   * The expiry tenor associated with the parameter.
+   */
+  @PropertyDefinition
```
wjnicholson: Tenor properties need to be validated not null

yukiiwashita:
Need a unit test?

PR #2022, opened 2019-07-08
MINOR: Additional argchecker usage

wjnicholson:


modules/market/src/main/java/com/opengamma/strata/market/param/YearMonthDateParameterMetadata.java
```diff
@@ -69,7 +69,6 @@
    * @return the parameter metadata based on the year-month
    */
   public static YearMonthDateParameterMetadata of(LocalDate date, YearMonth yearMonth) {
-    ArgChecker.notNull(date, "date");
```
jodastephen: I think this should be left in - when one arg checker is needed, I would usually check everything

modules/market/src/main/java/com/opengamma/strata/market/param/TenorDateParameterMetadata.java
```diff
@@ -62,7 +62,6 @@
    * @return the parameter metadata based on the tenor
    */
   public static TenorDateParameterMetadata of(LocalDate date, Tenor tenor) {
-    ArgChecker.notNull(date, "date");
```
jodastephen: I think this should be left in - when one arg checker is needed, I would usually check everything

PR #2027, opened 2019-07-11
Enhance performance by avoiding Math.pow

jodastephen:
Math.pow is slow when the power is a small integer
This particularly affects SABR volatility adjoint

### DIFF
--- a/modules/math/src/main/java/com/opengamma/strata/math/MathUtils.java
+++ b/modules/math/src/main/java/com/opengamma/strata/math/MathUtils.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2019 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.math;
+
+/**
+ * Simple utilities for maths.
+ */
+public final class MathUtils {
+
+  /**
+   * Returns the power of 2 (square).
+   * 
+   * @param value  the value to check
+   * @return {@code value * value}
+   */
+  public static double pow2(double value) {
+    return value * value;
+  }
+
+  /**
+   * Returns the power of 3 (cube).
+   * 
+   * @param value  the value to check
+   * @return {@code value * value * value}
+   */
+  public static double pow3(double value) {
+    return value * value * value;
+  }
+
+  /**
+   * Returns the power of 4.
+   * 
+   * @param value  the value to check
+   * @return {@code value * value * value * value}
+   */
+  public static double pow4(double value) {
+    return value * value * value * value;
+  }
+
+  //-------------------------------------------------------------------------
+  /**
+   * Checks if a number is near zero.
+   * <p>
+   * NaN and infinity are not near zero.
+   * 
+   * @param value  the value to check
+   * @param tolerance  the tolerance, must be positive and not NaN/Infinite (not validated)
+   * @return true if near zero
+   */
+  public static boolean nearZero(double value, double tolerance) {
+    return value >= -tolerance && value <= tolerance;
+  }
+
+  /**
+   * Checks if a number is near one.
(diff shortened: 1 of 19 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/SabrHaganVolatilityFunctionProvider.java
```diff
@@ -287,44 +301,48 @@ public ValueDerivatives volatilityAdjoint(
...
-          rhoBar = (0.5 * Math.pow(z / (1 - z), 2) + 0.25 * (z - 4.0) * Math.pow(z / (1.0 - z), 3) / (1.0 - z) *
-              rhoStar) * xzBar;
+          double zStar = 1 - z;
+          double zDivZStar = z / zStar;
+          rhoBar = (0.5 * pow2(zDivZStar) +
+              0.25 * (z - 4.0) * pow3(zDivZStar) / zStar * rhoStar) * xzBar;
         }
       } else {
         rhoBar = (1 / (Math.sqrt(1 - 2 * rho * z + z * z) + z - rho) *
             (-Math.pow(1 - 2 * rho * z + z * z, -0.5) * z - 1) + 1 / rhoStar) * xzBar;
```
wjnicholson: This line also reuses `(z * z)`

modules/math/src/main/java/com/opengamma/strata/math/MathUtils.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * Returns the power of 4.
+   * 
+   * @param value  the value to check
+   * @return {@code value * value * value * value}
+   */
+  public static double pow4(double value) {
+    return value * value * value * value;
```
wjnicholson: Since we're doing micro optimisation, is there any performance difference between this and doing 
```
double sq = value * value;
return sq * sq;
```
Or is that equivalent to what's done in hotspot?

modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/volatility/smile/SabrHaganVolatilityFunctionProvider.java
```diff
@@ -103,13 +108,13 @@ public double volatility(
...
     } else {
-      if (DoubleMath.fuzzyEquals(beta, 0, BETA_EPS)) {
+      if (nearZero(beta, BETA_EPS)) {
         double ln = Math.log(forward / k);
         z = nu * Math.sqrt(forward * k) * ln / alpha;
         zOverChi = getZOverChi(rho, z);
         vol = alpha * ln * zOverChi *
             (1 + timeToExpiry * (alpha * alpha / forward / k + nu * nu * (2 - 3 * rho * rho)) / 24) / (forward - k);
-      } else if (DoubleMath.fuzzyEquals(beta, 1, BETA_EPS)) {
+      } else if (nearOne(beta, BETA_EPS)) {
```
wjnicholson: Do you want comments on more potentially duplicated squares etc in this class or is this simply a review on replacing Math.pow() usages?

jodastephen: The goal was not to find all squares and replace them with `pow2()` - I suspect that would be counterproductive. In general, once you go beyond replacing `Math.pow` and the repeated large divisions, you can't tell whether a change makes a difference.

jodastephen:
I've made some more changes. The goal isn't to fully optimise as there are too many code paths. Instead, I'm just looking for some key changes around pow/sqrt. Not every extraction is necessarily faster as too many local variables may well be a problem too.

PR #2031, opened 2019-07-19
Remove N^2 uniqueness checks in interpolators

wjnicholson:
Not sure this is the best implementation - still seems a little boilerplatey.
Could definitely be streamlined with more ArgChecker methods, potentially ones that assert that a double array is sorted, unique, and doesn't contain NaN or infinity, but maybe that's overstepping what ArgChecker is responsible for.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
@@ -448,7 +448,7 @@ public final class DoubleArrayMath {
     double[] result = new double[len1];
     for (int i = 0; i < len1; i++) {
       int key = positions[i];
-      ArgChecker.inRange(key, 0, len1, "key must be in range: 0 <= key < length");
+      ArgChecker.inRange(key, 0, len1, "key");
       result[i] = values[key];
     }
     return result;
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
@@ -35,7 +35,7 @@ public final class DoubleArrayMath {
   //-------------------------------------------------------------------------
   /**
    * Converts a {@code double} array to a {@code Double} array.
-   * 
+   *
    * @param array  the array to convert
    * @return the converted array
    */
@@ -54,7 +54,7 @@ public final class DoubleArrayMath {
    * Converts a {@code Double} array to a {@code double} array.
    * <p>
    * Throws an exception if null is found.
-   * 
+   *
    * @param array  the array to convert
    * @return the converted array
    * @throws NullPointerException if null found
@@ -75,7 +75,7 @@ public final class DoubleArrayMath {
    * Calculates the sum total of all the elements in the array.
    * <p>
    * The input array is not mutated.
-   * 
+   *
    * @param array  the array to sum
    * @return the sum total of all the elements
    */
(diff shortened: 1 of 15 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
```diff
@@ -420,6 +420,92 @@ private static void swap(double[] keys, double[] values, int first, int second)
...
+   */
+  public static double[] reorderedCopy(double[] values, int[] positions) {
+    int len1 = positions.length;
+    if (len1 != values.length) {
+      throw new IllegalArgumentException("Value array cannot be reordered as they differ in length");
+    }
+    double[] result = new double[len1];
+    for (int i = 0; i < len1; i++) {
+      int key = positions[i];
+      ArgChecker.inRange(key, 0, len1, "key must be in range: 0 <= key < length");
```
jodastephen:
```suggestion
      ArgChecker.inRange(key, 0, len1, "key");
```

modules/collect/src/main/java/com/opengamma/strata/collect/DoubleArrayMath.java
```diff
@@ -476,6 +562,10 @@ private static void swap(double[] keys, double[] values, int first, int second)
 
   /**
    * Return the array lengths if they are the same, otherwise throws an {@code IllegalArgumentException}.
+   *
```
jodastephen: I would have just changed this to a `//` comment

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/CubicSplineInterpolator.java
```diff
@@ -47,16 +50,8 @@ public PiecewisePolynomialResult interpolate(final double[] xValues, final doubl
...
-      for (int j = i + 1; j < nDataPts; ++j) {
-        ArgChecker.isFalse(xValues[i] == xValues[j], "Data should be distinct");
-      }
-    }
-
-    double[] xValuesSrt = new double[nDataPts];
-    double[] yValuesSrt = new double[nDataPts];
-
-    xValuesSrt = Arrays.copyOf(xValues, nDataPts);
+    double[] xValuesSrt = Arrays.copyOf(xValues, nDataPts);
```
jodastephen: I suspect `xValues.clone()` might be quicker in cases like this.

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/CubicSplineInterpolator.java
```diff
@@ -193,13 +177,11 @@ public PiecewisePolynomialResultsWithSensitivity interpolateWithSensitivity(fina
       ArgChecker.isFalse(Double.isInfinite(yValues[i]), "yData containing Infinity");
     }
 
-    for (int i = 0; i < nDataPts; ++i) {
-      for (int j = i + 1; j < nDataPts; ++j) {
-        ArgChecker.isFalse(xValues[i] == xValues[j], "Data should be distinct");
-      }
+    if (DoubleStream.of(xValues).distinct().count() != xValues.length) {
```
jodastephen: `ArgChecker`?

modules/math/src/main/java/com/opengamma/strata/math/impl/interpolation/CubicSplineInterpolator.java
wjnicholson: Apparently `.clone()` may add a cast (since it's implicitly overriden from the `Object::clone()` method)
`Arrays.copyOf()` delegates to `System.arrayCopy` which is native, and doesn't involve a cast.
I'd have to benchmark to be sure

PR #2038, opened 2019-07-29
Add bean byte sources for URI and File

jodastephen:
Provides consumers with ability to fully use `ByteSource` as beans


jodastephen:
This changes the return type of `getByteSource()` in `ResourceLocator`. This is source compatible, but is not binary compatible if you use the method.

PR #2058, opened 2019-08-12
Make EtdPositions SecuritisedProductPositions

ashleyheath:


modules/product/src/main/java/com/opengamma/strata/product/etd/EtdPosition.java
```diff
@@ -21,20 +22,11 @@
...
-   * Gets the currency of the position.
-   * <p>
-   * This is the currency of the security.
-   *
-   * @return the trading currency
-   */
-  public default Currency getCurrency() {
-    return getSecurity().getCurrency();
-  }
+public interface EtdPosition<T extends SecuritizedProduct>
```
jodastephen: Is it necessary to change this interface? Could just make the concrete classes implement `SecuritizedProduct` without changing here, then no need for generics,

PR #2064, opened 2019-08-27
Use regex possessive qualifier

jodastephen:
Avoids potential for catastrophic backtracking

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
@@ -150,18 +150,22 @@ public final class RatesCalibrationCsvLoader {
   private static final String CURVE_MIN_GAP = "Min Gap";
   private static final String CURVE_CLASH_ACTION = "Clash Action";
 
+  // these regexes use possessive ?+ and ++ to prevent unnecessary backtracking
+
   // Regex to parse FRA time string
-  private static final Pattern FRA_TIME_REGEX = Pattern.compile("P?([0-9]+)M? ?X ?P?([0-9]+)M?");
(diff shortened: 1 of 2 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/RatesCalibrationCsvLoader.java
```diff
@@ -150,18 +150,22 @@
...
   // Regex to parse FRA time string
-  private static final Pattern FRA_TIME_REGEX = Pattern.compile("P?([0-9]+)M? ?X ?P?([0-9]+)M?");
+  private static final Pattern FRA_TIME_REGEX = Pattern.compile("P?+([0-9]++)M?+ ?+X ?+P?+([0-9]++)M?+");
   // Regex to parse future time string
-  private static final Pattern FUT_TIME_REGEX = Pattern.compile("P?((?:[0-9]+D)?(?:[0-9]+W)?(?:[0-9]+M)?) ?[+] ?([0-9]+)");
+  private static final Pattern FUT_TIME_REGEX =
+      Pattern.compile("P?+((?:[0-9]++D)?+(?:[0-9]++W)?+(?:[0-9]++M)?+) ?+[+] ?+([0-9]++)");
   // Regex to parse future month string
-  private static final Pattern FUT_MONTH_REGEX = Pattern.compile("([A-Z][A-Z][A-Z][0-9][0-9])");
+  private static final Pattern FUT_MONTH_REGEX = Pattern.compile("([A-Z]{3}[0-9]{2})");
```
wjnicholson: No possessives on this?

jodastephen: Possessives only apply to `?`, `+` and `*`.

wjnicholson: Ah my bad - it seems you can apply possessives to curly braces although only when they are open ended i.e `[A-Z]{1,}+` which isn't required here since there's an exact match

PR #2071, opened 2019-09-02
Move from TestNG to JUnit 5: strata-math

jodastephen:
Mostly created by the conversion tool

### DIFF
--- a/modules/math/src/test/java/com/opengamma/strata/math/impl/interpolation/NonnegativityPreservingCubicSplineInterpolatorTest.java
+++ b/modules/math/src/test/java/com/opengamma/strata/math/impl/interpolation/NonnegativityPreservingCubicSplineInterpolatorTest.java
@@ -5,10 +5,11 @@
  */
 package com.opengamma.strata.math.impl.interpolation;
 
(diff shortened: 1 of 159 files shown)


modules/math/src/test/java/com/opengamma/strata/math/impl/interpolation/NonnegativityPreservingCubicSplineInterpolatorTest.java
```diff
@@ -268,265 +275,281 @@ public void flipMultiTest() {
...
   }
 
   /**
    * 
    */
-  @Test(expectedExceptions = IllegalArgumentException.class)
+  @Test
   public void nullXdataMultiTest() {
-    double[] xValues = new double[] {1., 2., 3., 4. };
+    double[] xValues = null;
```
ashleyheath: This deliberate?

jodastephen: Code moved to a lambda, but local variables can't be mutated, so need to nullify at source

PR #2075, opened 2019-09-03
Support PropertySet-style combination directly with IniFiles

ashleyheath:


modules/collect/src/test/java/com/opengamma/strata/collect/io/IniFileTest.java
```diff
@@ -174,6 +176,25 @@ public void test_of_ioException() {
...
+    Map<String, PropertySet> bSections = ImmutableMap.of(
+        "InB", PropertySet.of(ImmutableMap.of("BTest", "BValue")),
+        "InBoth", PropertySet.of(ImmutableMultimap.of("InBoth", "Ignored", "InBoth", "AlsoIgnored")));
+    IniFile b = IniFile.of(bSections);
+
+    IniFile combined = a.combinedWith(b);
+    assertThat(combined.sections()).containsExactlyInAnyOrder("InA", "InB", "InBoth");
+    assertThat(combined.section("InA").valueList("ATest")).isEqualTo(ImmutableList.of("AValue"));
+    assertThat(combined.section("InB").valueList("BTest")).isEqualTo(ImmutableList.of("BValue"));
+    assertThat(combined.section("InBoth").valueList("InBoth")).isEqualTo(ImmutableList.of("Override", "AlsoOverrides"));
```
jodastephen: This looks like a bug in `PropertySet`. All the keys should be retained on merge AFAICT. The bug is that it uses `valueList()` without considering the need to merge lists.

jodastephen: Scratch that, when merging a single key it must be a complete override, so there is no way to accurately merge a list.

PR #2084, opened 2019-09-19
PropertySet#overrideWith let the overriding set take order precedence.

wjnicholson:


jodastephen:
Not implementing this for now, there may be a better way to improve `ExtendedEnum` config

PR #2087, opened 2019-09-25
Risky annuity for single name CDS and CDS index

yukiiwashita:
* Added public risky annuity method, which was internally calculated.
* This is a prerequisite for default swaptions.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaCdsProductPricer.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaCdsProductPricer.java
@@ -19,7 +19,6 @@ import com.opengamma.strata.basics.currency.CurrencyAmount;
 import com.opengamma.strata.collect.ArgChecker;
 import com.opengamma.strata.collect.array.DoubleArray;
 import com.opengamma.strata.collect.tuple.Pair;
-import com.opengamma.strata.market.sensitivity.PointSensitivities;
 import com.opengamma.strata.market.sensitivity.PointSensitivityBuilder;
 import com.opengamma.strata.math.impl.util.Epsilon;
 import com.opengamma.strata.pricer.common.PriceType;
@@ -139,14 +138,14 @@ public class IsdaCdsProductPricer {
    * @param refData  the reference data
    * @return the present value sensitivity
    */
-  public PointSensitivities priceSensitivity(
+  public PointSensitivityBuilder priceSensitivity(
       ResolvedCds cds,
       CreditRatesProvider ratesProvider,
       LocalDate referenceDate,
       ReferenceData refData) {
 
     if (isExpired(cds, ratesProvider)) {
-      return PointSensitivities.empty();
+      return PointSensitivityBuilder.none();
     }
     LocalDate stepinDate = cds.getStepinDateOffset().adjust(ratesProvider.getValuationDate(), refData);
     LocalDate effectiveStartDate = cds.calculateEffectiveStartDate(stepinDate);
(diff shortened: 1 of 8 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaCdsProductPricer.java
```diff
@@ -347,6 +347,32 @@ public double riskyAnnuity(
...
+   * @param refData  the reference data
+   * @return the risky annuity sensitivity
+   */
+  public PointSensitivityBuilder riskyAnnuitySensitivity(
+      ResolvedCds cds,
+      CreditRatesProvider ratesProvider,
+      LocalDate referenceDate,
+      ReferenceData refData) {
+
+    ArgChecker.isTrue(cds.getProtectionEndDate().isAfter(ratesProvider.getValuationDate()), "CDS already expired");
```
jodastephen: Would it not be better to return an empty sensitivity? In either case, this case should be in the Javadoc.

modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaCdsProductPricer.java
```diff
@@ -347,6 +347,32 @@ public double riskyAnnuity(
...
+   * The risky annuity sensitivity of the product is the sensitivity of risky annuity to the underlying curves.
+   * The resulting sensitivity is based on the currency of the CDS product.
+   * 
+   * @param cds  the product
+   * @param ratesProvider  the rates provider
+   * @param referenceDate  the reference date
+   * @param refData  the reference data
+   * @return the risky annuity sensitivity
+   */
+  public PointSensitivityBuilder riskyAnnuitySensitivity(
```
jodastephen: Is returning the builder the right choice for a public method?

modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaHomogenousCdsIndexProductPricer.java
```diff
@@ -265,6 +265,64 @@ public CurrencyAmount rpv01(
...
+   */
+  public double riskyAnnuity(
+      ResolvedCdsIndex cdsIndex,
+      CreditRatesProvider ratesProvider,
+      LocalDate referenceDate,
+      PriceType priceType,
+      ReferenceData refData) {
+
+    if (isExpired(cdsIndex, ratesProvider)) {
+      return 0d;
```
jodastephen: This case should be in the Javadoc

modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaHomogenousCdsIndexProductPricer.java
```diff
@@ -265,6 +265,64 @@ public CurrencyAmount rpv01(
...
+   * @return the risky annuity sensitivity
+   */
+  public PointSensitivityBuilder riskyAnnuitySensitivity(
+      ResolvedCdsIndex cdsIndex,
+      CreditRatesProvider ratesProvider,
+      LocalDate referenceDate,
+      ReferenceData refData) {
+
+    if (isExpired(cdsIndex, ratesProvider)) {
+      return PointSensitivityBuilder.none();
```
jodastephen: This case should be in the Javadoc

modules/pricer/src/main/java/com/opengamma/strata/pricer/credit/IsdaCdsProductPricer.java
```diff
@@ -200,14 +201,14 @@ public CurrencyAmount presentValue(
    * @param refData  the reference data
    * @return the present value sensitivity
    */
-  public PointSensitivityBuilder presentValueSensitivity(
+  public PointSensitivities presentValueSensitivity(
```
jodastephen: Sorry for confusion, but this would be an incompatible change, so we can't do it. If all the other methods on the class return a builder already then that is fine and the new method should too.

PR #2095, opened 2019-10-11
Extend the CSV trade loader

jodastephen:
Allow the standard parsing to be overridden for a given product type

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvInfoResolver.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvInfoResolver.java
@@ -433,8 +433,28 @@ public interface TradeCsvInfoResolver {
     return CdsTradeCsvPlugin.parseCdsIndex(row, info, this);
   }
 
+  //-------------------------------------------------------------------------
   /**
-   * Parses any other kind of trade from CSV.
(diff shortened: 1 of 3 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/TradeCsvInfoResolver.java
```diff
@@ -433,8 +433,28 @@ public default CdsIndexTrade parseCdsIndexTrade(CsvRow row, TradeInfo info) {
...
 
+  //-------------------------------------------------------------------------
   /**
-   * Parses any other kind of trade from CSV.
+   * Parses any kind of trade from CSV before standard matching.
+   * <p>
+   * This is called before the standard matching on the 'Product Type' column.
+   * As such, it allows the standard parsing to be replaced for a given type.
+   * 
+   * @param typeUpper  the upper case product type column
```
akshaishah: Has 2 spaces after the parameter name always been a thing?

jodastephen: In Strata, yes

akshaishah: Got it 👍

PR #2096, opened 2019-10-11
Fix bond pricer dirty price after new issue

jodastephen:
Adjust by eventsPerYear

See [forum](https://forums.opengamma.com/t/yield-duration-diff-on-new-issue-gilts-vs-bloomberg/912) post.


modules/pricer/src/main/java/com/opengamma/strata/pricer/bond/DiscountingFixedCouponBondProductPricer.java
```diff
@@ -804,6 +804,9 @@ private double factorToNextCoupon(ResolvedFixedCouponBond bond, LocalDate settle
     int couponIndex = couponIndex(bond.getPeriodicPayments(), settlementDate);
     double factorSpot = accruedYearFraction(bond, settlementDate);
     double factorPeriod = bond.getPeriodicPayments().get(couponIndex).getYearFraction();
+    if (bond.getYieldConvention().equals(GB_BUMP_DMO)) {
```
jonfreedman: I think this change should apply for all bonds not just UK gilts

PR #2097, opened 2019-10-14
Add a new flag to control whether the presence of unsupported FpML elements triggers errors when parsing

ashleyheath:


modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocumentParser.java
```diff
@@ -178,7 +182,36 @@ public static FpmlDocumentParser of(
...
+   * <li>the {@linkplain ReferenceData reference data}
+   * </ul>
+   *
+   * @param ourPartySelector  the selector used to find "our" party within the set of parties in the FpML document
+   * @param tradeInfoParser  the trade info parser
+   * @param tradeParsers  the map of trade parsers, keyed by the FpML element name
+   * @param refData  the reference data to use
+   * @return the document parser
+   */
+  public static FpmlDocumentParser withLenientMode(
```
jodastephen: Would be better to have `public FpmlDocumentParser withLenientMode()` which takes the instance you already have and returns a new instance with the flag flipped.

PR #2098, opened 2019-10-15
Add CSV file header matching for multiple headers

jodastephen:
Provides the ability to match a set of headers.


jodastephen:
@cjkent The build failure was a CircleCI temporary failure

PR #2107, opened 2019-10-31
Enhance Money

jodastephen:
Allow `Money` to be used for calculations

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/currency/Money.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/currency/Money.java
@@ -47,6 +47,16 @@ public class Money
 
   //-------------------------------------------------------------------------
   /**
+   * Obtains a zero amount instance of {@code Money} for the specified currency.
+   *
+   * @param currency  the currency the amount is in
+   * @return the zero amount instance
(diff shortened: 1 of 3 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/currency/Money.java
```diff
@@ -142,6 +152,98 @@ public BigDecimal getAmount() {
...
+   * Returns a copy of this {@code Money} with the amount multiplied.
+   * <p>
+   * This takes this amount and multiplies it by the specified value.
+   * <p>
+   * This instance is immutable and unaffected by this method.
+   * 
+   * @param valueToMultiplyBy  the scalar amount to multiply by
+   * @return an amount based on this with the amount multiplied
+   */
+  public Money multipliedBy(long valueToMultiplyBy) {
```
ashleyheath: Is it worth having `public Money multipliedBy(BigDecimal valueToMultiplyBy)` too?

jodastephen: The problem is rounding. Anything other than an integer needs rounding.

PR #2112, opened 2019-11-07
LazyFxRateProvider

ashleyheath:
Add a LazyFxRateProvider for just-in-time construction of a delegate FxRateProvider.


modules/basics/src/main/java/com/opengamma/strata/basics/currency/LazyFxRateProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private final Supplier<FxRateProvider> target;
+
+  private LazyFxRateProvider(Supplier<FxRateProvider> target) {
+    ArgChecker.notNull(target, "target");
+    this.target = Suppliers.memoize(target::get);
+  }
+
+  /**
+   * Create a new LazyFxRateProvider.
```
jodastephen: We don't say "create" or "new" with factory methods in Strata. "Obtains" or "Returns" are suitable

modules/basics/src/main/java/com/opengamma/strata/basics/currency/LazyFxRateProvider.java
```diff
@@ -0,0 +1,59 @@
...
+ * until actually necessary.
+ * <p>
+ * This is typically useful where you <em>may</em> need a {@code MarketDataFxRateProvider}
+ * but want to delay loading market data to construct the provider until you are sure you actually do need it.
+ */
+public class LazyFxRateProvider implements FxRateProvider {
+
+  private final Supplier<FxRateProvider> target;
+
+  private LazyFxRateProvider(Supplier<FxRateProvider> target) {
```
jodastephen: In Strata even private fields/constructors need docs for top level classes, fields are generally Javadoc, but private methods can be `//` such as `// restricted constructor`

modules/basics/src/main/java/com/opengamma/strata/basics/currency/LazyFxRateProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    this.target = Suppliers.memoize(target::get);
+  }
+
+  /**
+   * Create a new LazyFxRateProvider.
+   *
+   * @param fxRateProviderSupplier a supplier of a rate provider
+   * @return the rate provider
+   */
+  public static LazyFxRateProvider of(Supplier<FxRateProvider> fxRateProviderSupplier) {
```
jodastephen: Return the interface, not the concrete type

ashleyheath: What's the reason for that? It doesn't seem consistent with the factories of other `FxRateProvider` implementations.

jodastephen: If the type returned is a bean, it should be returned. Otherwise, there is no need to refer to the type. In fact, looking at it again, it probably makes sense to make the new class package scoped and put the static factory method on the interface. We've done this a lot in Strata to minimise the exposed API.

PR #2116, opened 2019-11-27
Add name to byte source

jodastephen:
Access to the name of the underlying file is useful functionality


modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
```diff
@@ -149,21 +155,32 @@ public static ArrayByteSource ofUtf8(String str) {
...
    * 
    * @param other  the other byte source
    * @return the byte source
    * @throws UncheckedIOException if an IO error occurs
    */
   public static ArrayByteSource from(ByteSource other) {
-    if (other instanceof ArrayByteSource) {
-      return (ArrayByteSource) other;
+    if (other instanceof BeanByteSource) {
+      return ((BeanByteSource) other).load();
```
wjnicholson: This worries me - it looks like the default implementation calls this `from` method, so if somebody forgets to override it (or hasn't yet) then we get infinite recursion and a stack overflow

modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
```diff
@@ -221,6 +255,24 @@ public MetaBean metaBean() {
...
+   * Returns an instance with the file name updated.
+   * <p>
+   * If a path is passed in, only the file name is retained.
+   * 
+   * @param fileName the file name, an empty string can be used to remove the file name
+   * @return a source with the specified file name
+   */
+  public ArrayByteSource withFileName(String fileName) {
+    ArgChecker.notNull(fileName, "fileName");
+    return new ArrayByteSource(array, Paths.get(fileName).getFileName().toString());
```
wjnicholson: Uhhh I think this uses the local default filesystem to resolve that filename, which isn't necessarily what you want.

modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
wjnicholson: And why get rid of the fast path without copying?

modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
```diff
@@ -73,10 +75,14 @@
    * The byte array.
    */
   private final byte[] array;
+  /**
+   * The file name, null if not known.
+   */
+  private final String fileName;
```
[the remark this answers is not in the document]
jodastephen: Its a common use case to process based on the file name, such as extracting a date. Being able to carry the file name around proves to be very useful.

modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
```diff
@@ -149,21 +155,32 @@ public static ArrayByteSource ofUtf8(String str) {
...
    * @param other  the other byte source
    * @return the byte source
    * @throws UncheckedIOException if an IO error occurs
    */
   public static ArrayByteSource from(ByteSource other) {
     if (other instanceof ArrayByteSource) {
-      return (ArrayByteSource) other;
+      return ((ArrayByteSource) other);
+    }
+    String fileName = other instanceof BeanByteSource ? ((BeanByteSource) other).getFileName().orElse(null) : null;
```
wjnicholson: Probably want to nest this into the other `if/else` otherwise we're calling the `.toString()` when it isn't required

modules/collect/src/main/java/com/opengamma/strata/collect/io/ArrayByteSource.java
```diff
@@ -175,6 +175,8 @@ public static ArrayByteSource from(ByteSource other) {
       String path = str.substring(23, pos);
       int lastSlash = path.lastIndexOf('/');
       fileName = path.substring(lastSlash + 1);
+    } else if (other instanceof BeanByteSource) {
```
wjnicholson: I meant stick this first, since you don't need to call `toString()` if this is true, and nest the rest under an else.

PR #2120, opened 2019-12-04
Additional utilities for handling ZIP files

jodastephen:
Extend class to cover more zip/unzip use cases.

### DIFF
--- a/modules/collect/src/test/java/com/opengamma/strata/collect/io/ZipUtilsTest.java
+++ b/modules/collect/src/test/java/com/opengamma/strata/collect/io/ZipUtilsTest.java
@@ -10,20 +10,98 @@ import static org.assertj.core.api.Assertions.fail;
 import static org.assertj.core.api.Assertions.within;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


modules/collect/src/test/java/com/opengamma/strata/collect/io/ZipUtilsTest.java
```diff
@@ -47,6 +125,71 @@ public void test_zipInMemory() throws Exception {
...
+  }
+
+  //-------------------------------------------------------------------------
+  @Test
+  public void test_unzipInMemory_zip() {
+    ArrayByteSource source1 = ArrayByteSource.ofUtf8("Hello World").withFileName("TestFile1.txt");
+    ArrayByteSource source2 = ArrayByteSource.ofUtf8("Hello Planet").withFileName("TestFile2.txt");
+    ArrayByteSource zipFile = ZipUtils.zipInMemory(ImmutableList.of(source1, source2)).withFileName("Test.foo");
+
+    ZipUtils.unzipInMemory(zipFile, (name, extracted) -> {
```
martynassateika: If the `BiConsumer` wasn't invoked at all this test would still pass.

PR #2122, opened 2019-12-19
Add Attributes.withAttributes()

jodastephen:
Provide ability to bulk update attributes
Make `SimpleAttributes` public


modules/product/src/main/java/com/opengamma/strata/product/Attributes.java
```diff
@@ -113,4 +112,21 @@ public static Attributes empty() {
    */
   public abstract <T> Attributes withAttribute(AttributeType<T> type, T value);
 
+  /**
+   * Returns a copy of this instance with the attributes added.
+   * <p>
+   * This returns a new instance with the specified attribute added.
```
akshaishah: `This returns a new instance with the specified attributes added.`

modules/product/src/main/java/com/opengamma/strata/product/Attributes.java
```diff
@@ -113,4 +112,21 @@ public static Attributes empty() {
...
+   * The attributes are added using {@code Map.putAll(type, value)} semantics.
+   * 
+   * @param other  the other instance to copy from
```
akshaishah: this line has 2 spaces, whereas the line below has 1 space

jodastephen: The 2 space rule only applies to `@param`

akshaishah: Ah :) didn't realise, apologies!

PR #2131, opened 2020-02-05
MapStream factory method from collection only

headphonejack:
Caters to the common use-case of creating a MapStream from a collection and operating on the values while retaining the original object as a key.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
@@ -94,7 +94,18 @@ public final class MapStream<K, V>
    * @return a stream of map entries derived from the values in the collection
    */
   public static <V> MapStream<V, V> of(Collection<V> collection) {
-    return of(collection.stream(), key -> key);
+    return of(collection.stream());
+  }
+
+  /**
+   * Returns a stream of map entries where the keys and values are taken from a stream.
+   *
+   * @param <V>  the key and value type
+   * @param stream  the stream
+   * @return a stream of map entries derived from the values in the stream
+   */
+  public static <V> MapStream<V, V> of(Stream<V> stream) {
+    return of(stream, key -> key);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -100,6 +111,7 @@
     return of(collection.stream(), keyFunction);
   }
 
+
```
wjnicholson: Erroneous extra new line?

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -86,6 +86,17 @@
...
+  /**
+   * Returns a stream of map entries where the keys and values are taken from a collection.
+   *
+   * @param <V>  the key and value type
+   * @param collection  the collection
+   * @return a stream of map entries derived from the values in the collection
+   */
+  public static <V> MapStream<V, V> of(Collection<V> collection) {
+    return of(collection.stream(), key -> key);
+  }
```
wjnicholson: Might as well add the other option of
`MapStream.of(Stream<V> stream)`
and delegate to that.

modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
```diff
@@ -52,6 +53,16 @@ public void values() {
     assertThat(result).isEqualTo(ImmutableList.of(1, 2, 3, 4));
   }
 
+  public void keys_fromList() {
+    List<String> result = MapStream.of(list).keys().collect(toImmutableList());
+    assertThat(result).isEqualTo(list);
+  }
+
+  public void values_fromList() {
```
wjnicholson: Same for the stream version too!

modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -86,6 +86,28 @@
...
+
+  /**
+   * Returns a stream of map entries where the keys and values are taken from a stream.
+   *
+   * @param <V>  the key and value type
+   * @param stream  the stream
+   * @return a stream of map entries derived from the values in the stream
+   */
+  public static <V> MapStream<V, V> of(Stream<V> stream) {
+    return of(stream, key -> key);
```
wjnicholson: Looking at the current factory methods in MapStream, this should probably be inlined and have the identity function removed.

PR #2132, opened 2020-02-07
Add optional getters to IniFile and PropertySet

jodastephen:
Optional configuration turns out to be quite common

### DIFF
--- a/modules/collect/src/test/java/com/opengamma/strata/collect/io/PropertySetTest.java
+++ b/modules/collect/src/test/java/com/opengamma/strata/collect/io/PropertySetTest.java
@@ -5,6 +5,7 @@
  */
 package com.opengamma.strata.collect.io;
 
+import static com.opengamma.strata.collect.Guavate.entry;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


modules/collect/src/test/java/com/opengamma/strata/collect/io/PropertySetTest.java
```diff
@@ -32,6 +32,7 @@ public void test_empty() {
 
     assertThat(test.isEmpty()).isEqualTo(true);
     assertThat(test.contains("unknown")).isEqualTo(false);
+    assertThat(test.findValue("unknown")).isEmpty();
```
martynassateika: Could add this in `test_of_multimap`:

```java
assertThatIllegalArgumentException().isThrownBy(() -> test.findValue("a"));
```

ashleyheath:
> If a key has multiple values, the values will be returned as a comma separated list.

Isn't this exactly what `public ImmutableList<String> valueList(String key)` is for? Concatenating the values seems like it'll introduce more bugs than it'll fix...

jodastephen:
Having methods like `asMap()` throw an exception was causing issues anyway.

PR #2154, opened 2020-05-11
Generate holiday calendars as part of the build

jodastephen:
Avoid issues with forgetting to generate them


jodastephen:
Can't get this to work in both IDE and command line

PR #2161, opened 2020-05-20
Add Guavate.tryCatchToOptional

jodastephen:
Convert an exception throwing `getFoo()` method to an `Optional` returning `findFoo()` method

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
@@ -140,6 +140,25 @@ public final class Guavate {
 
   //-------------------------------------------------------------------------
   /**
+   * Wraps a try-catch block around an expression, avoiding exceptions.
+   * <p>
+   * This converts an exception throwing method into an optional returning one by discarding the exception.
+   * In most cases it is better to add a `findXxx()` method to the code you want to call.
+   * 
+   * @param <T> the type of the result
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -138,6 +138,25 @@ private Guavate() {
...
 
+  //-------------------------------------------------------------------------
+  /**
+   * Wraps a try-catch block around an expression, avoiding exceptions.
+   * <p>
+   * This converts an exception throwing method into an optional returning one by discarding the exception.
+   * In most cases it is better to add a `findXxx()` method to the code you want to call.
+   * 
+   * @param <T> the type of the result
+   * @param supplier the supplier that might throw an exception
```
wjnicholson, line 149: Spacing looks odd here - is that double spaced?

modules/collect/src/main/java/com/opengamma/strata/collect/Guavate.java
```diff
@@ -138,6 +138,25 @@ private Guavate() {
...
+   * @return the value wrapped in an exception, empty if the method returns null or an exception is thrown
```
wjnicholson, line 150:
```suggestion
   * @return the value wrapped in an optional, empty if the method returns null or an exception is thrown
```

PR #2162, opened 2020-05-22
Cache size call for DenseLocalDateDoubleTimeSeries

wjnicholson:
Current implementation has an O(N) lookup time.

Profiling revealed the `isEmpty()` check in `.get(LocalDate)` to be a hotspot as it could potentially traverse a reasonable amount of the array (density is required to be >0.7 to become dense, so potentially up to 0.3 of a large array could be traversed, but additionally this isn't preserved when doing head or tail of the timeseries)

Could use a `Suppliers.memoize()` call instead but decided against as in several cases we can preserve the size information. Could switch back though if that would be clearer.
Also could have cached the `isEmpty()` check instead of piggy-backing on size() and a full traversal but that feels like over optimization.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
@@ -200,8 +200,9 @@ final class DenseLocalDateDoubleTimeSeries
   /**
    * The size of the time series.
    * The amount of valid values.
+   * Offset by one to account for 0 being the unset value of an int, hence the actual size is size - 1. 
    */
-  private transient int size; // not a property, derived from points and cached using racy single check as is an O(n) lookup
+  private transient int size; // not a property, derived from points and cached as is an O(n) lookup
 
   /**
    * Package protected factory method intended to be called
@@ -222,7 +223,7 @@ final class DenseLocalDateDoubleTimeSeries
 
     double[] points = new double[dateCalculation.calculatePosition(startDate, endDate) + 1];
     Arrays.fill(points, Double.NaN);
-    int size = 0;
+    int size = 1;
     for (LocalDateDoublePoint pt : in(values)) {
       points[dateCalculation.calculatePosition(startDate, pt.getDate())] = pt.getValue();
       size++;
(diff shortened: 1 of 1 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
```diff
@@ -215,41 +222,53 @@ static LocalDateDoubleTimeSeries of(
 
     double[] points = new double[dateCalculation.calculatePosition(startDate, endDate) + 1];
     Arrays.fill(points, Double.NaN);
-    values.forEach(pt -> points[dateCalculation.calculatePosition(startDate, pt.getDate())] = pt.getValue());
-    return new DenseLocalDateDoubleTimeSeries(startDate, points, dateCalculation, true);
+    int size = 0;
+    for (LocalDateDoublePoint pt : Guavate.in(values)) {
```
jodastephen, line 226: Static import

modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
```diff
@@ -215,41 +222,53 @@ static LocalDateDoubleTimeSeries of(
...
   @Override
   public boolean isEmpty() {
-    return !validIndices().findFirst().isPresent();
+    return size() == 0;
   }
 
   @Override
   public int size() {
-    return (int) validIndices().count();
+    int s = size;
```
jodastephen, line 266: Add comment "threadsafe via racy single check" or similar

wjnicholson: @jodastephen Just thinking - is this correct if we're using `-1` as the unset value rather than `0`? I can't seem to find documentation on whether one will always see a "set" value for the field or whether it can see the "unset value" of 0 before we would set it to -1.

jodastephen: True. I've never seen a case with -1 as the default. 0 has special meaning in the JVM, so best to stick with 0 as the default and adjust the size to cope (add one or size 0 -> -1)

wjnicholson: Yeah that was my thought, just do an offset by 1!

PR #2163, opened 2020-05-22
Overriden daysBetween implementation for ImmutableHolidayCalendar performance

wjnicholson:
Noticed when calculating lots of PVs that there was a major hot spot calculating year fractions using Business252DayCount.
Other day counts calls have a relatively fixed cost - they don't need to check individual holidays.
Whereas this uses the holiday calendar `daysBetween` which streams over every date and checks whether it is a holiday, creating lots of intermediary objects and lookups.
We can use the bit-packed lookup in ImmutableHolidayCalendar and the JDK method `Integer.bitCount()` to count business days per month in a single call.

Using some very unscientific benchmarking of the `test_yearFraction` method on `Business252DayCountTest` for 300K year fraction calls between 1 and 1000 days in the future the total time went from 51.49 seconds locally to 256ms, or just over 200 times faster.

Limiting to just the next day (as in requesting the yearFraction between today and tomorrow) 100 million times the speed up was from 27 seconds to 8s, or roughly 3 times faster

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/date/ImmutableHolidayCalendar.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/date/ImmutableHolidayCalendar.java
@@ -626,7 +626,7 @@ public final class ImmutableHolidayCalendar
       // e.g 4th day of month - want holidays from index 3 inclusive
       int start = Integer.bitCount(lookup[startIndex] & (BIT_MASK_ALL_ONES << (startInclusive.getDayOfMonth() - 1)));
       // count of last month = ones before day of month exclusive == total for month - ones after end inclusive
-      int missingEnd = Integer.bitCount(lookup[endIndex] & (BIT_MASK_ALL_ONES << endExclusive.getDayOfMonth() - 1));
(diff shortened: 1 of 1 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/date/ImmutableHolidayCalendar.java
```diff
@@ -606,6 +613,49 @@ private LocalDate lastBusinessDayOfMonthOutOfRange(LocalDate date) {
...
+      // count of last month = ones before day of month exclusive == total for month - ones after end inclusive
+      int missingEnd = Integer.bitCount(lookup[endIndex] & (BIT_MASK_ALL_ONES << endExclusive.getDayOfMonth() - 1));
+      if (startIndex == endIndex) {
+        // same month - return holidays up to end exclusive 
+        return start - missingEnd;
+      }
+      
+      int end = Integer.bitCount(lookup[endIndex]) - missingEnd;
+      // otherwise add start and end month counts, and sum months between
+      long count = start + end;
```
jodastephen, line 637: My maths suggests `int` is sufficient here, as the maximum days in 10,000 years is within capacity

PR #2166, opened 2020-05-29
Cache isEmpty separately from size in DenseLocalDateDoubleTimeSeries

wjnicholson:
Realised my previous PR would add a lot of computation in the case that we rarely access a time series


modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/DenseLocalDateDoubleTimeSeries.java
```diff
@@ -259,7 +266,16 @@ private DenseLocalDateDoubleTimeSeries(
   //-------------------------------------------------------------------------
   @Override
   public boolean isEmpty() {
-    return size() == 0;
+    // threadsafe via racy single-check idiom
+    int e = isEmpty;
+    if (e == 0) {
+      // take cached size if already calculated
+      int s = size;
+      boolean any = s != 0 ? s > 1 : validIndices().findAny().isPresent();
```
jodastephen, line 274: I suspect if you are worried about performance, looping over the double[] and calling `isValidPoint` would be quicker than the more weighty stream system. Also, I'd expect that `size()` and the constructors should set `isEmpty`.

PR #2172, opened 2020-06-02
Comparable MarketDataId

jodastephen:
`MarketDataId` should be comparable.
It is possible that this will be incompatible if someone has implemented `Comparable` themselves.


jodastephen:
It turns out that lots of other classes are not comparable, so lets not make this a special case.

PR #2177, opened 2020-06-11
Lookup FX swap convention by currency pair

jodastephen:
Adds a useful way to lookup conventions
Also complete the FX swap conventions for the key four currencies

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConvention.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConvention.java
@@ -48,6 +48,21 @@ public interface FxSwapConvention
   }
 
   /**
+   * Obtains the standard convention for the specified currency pair.
+   * <p>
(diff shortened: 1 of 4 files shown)


modules/product/src/main/java/com/opengamma/strata/product/fx/type/FxSwapConvention.java
```diff
@@ -47,6 +47,21 @@ public static FxSwapConvention of(String uniqueName) {
     return extendedEnum().lookup(uniqueName);
   }
 
+  /**
+   * Obtains the standard convention for the specified currency pair.
+   * <p>
+   * The convention accessed will be for the conventional order of the currency pair.
+   * 
+   * @param currencyPair  the currency pair to get the convention for
```
akshaishah, line 55:
```suggestion
   * @param currencyPair the currency pair to get the convention for
```

jodastephen: In Strata, its two spaces :-)

PR #2184, opened 2020-06-25
Add MarketTenor to represent ON/TN/SN/SW concept

jodastephen:
Market tenors are distinct from tenors used in Strata

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/date/MarketTenor.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/date/MarketTenor.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2020 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package com.opengamma.strata.basics.date;
+
+import java.io.Serializable;
+import java.time.Period;
+
+import org.joda.convert.FromString;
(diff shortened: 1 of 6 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/date/MarketTenor.java
```diff
@@ -0,0 +1,271 @@
...
+   */
+  private final String code;
+  /**
+   * The tenor.
+   */
+  private final Tenor tenor;
+  /**
+   * The order.
+   */
+  private final int order;
```
brianweller89, line 90: The order concept feels a bit too 'magic'... I get that it's trying to solve for the problem of some market tenors having a known number of days whilst others use the default spot lag but it takes some getting your head around.

Could use a private SpotType enum with values ZERO_DAYS, ONE_DAY and MARKET_STANDARD? You'd need a bit more code in adjustSpotLag to map to the right number of days for the non standard one.

Alternatively you could keep the order concept but associate it with the private enum values (or just use ordinal()). Might give you the best of both worlds in terms of making it easier to read and keeping the code in the methods simple

jodastephen: An enum feels over the top, but I think a variable rename and a constant make it clearer.

PR #2192, opened 2020-07-28
fixed deserialisation for ArrayByteSource with filename

anand-uk:


wjnicholson:
Could add a test that would have failed previously?

Issue #1820, opened 2018-10-12
Curve calibration with overnight futures


jodastephen:
Strata can currently support rates curve calibration with Ibor futures. Now that `OvernightFuture` exists, the calibration should be extended.

PR #2203, opened 2020-08-20
Add Overnight Future contract specifications

jodastephen:
Add Overnight Future contract specifications

See #1820
Futures should be modelled as contract specifications, not conventions


marc-henrard (approved):
A couple of cosmetic comments. Approved as is, but you may want to modify for clarity.

PR #2204, opened 2020-08-20
Use Math.abs when checking inflation inputs

jodastephen:
Use absolute value and a smaller tolerance based on feedback


jonfreedman:
I ended up with the following but I think this looks fine also

```java
Math.abs(adjustmentType.applyShift(yFixing, seasonality.get(i)) / yFixing) - 1d < 1.0E-10
```

marc-henrard (approved):
Good for me.

PR #2210, opened 2020-08-27
Support complex date sequences

jodastephen:
Support "every three months with every month for the first 6 month".
Define the base sequence as "every three months".
and the full sequence as all known dates.
Add `SequenceDate` to capture the more complex state.

### DIFF
--- a/modules/basics/src/main/java/com/opengamma/strata/basics/date/StandardDateSequences.java
+++ b/modules/basics/src/main/java/com/opengamma/strata/basics/date/StandardDateSequences.java
@@ -53,9 +53,8 @@ enum StandardDateSequences implements DateSequence {
     }
 
     private LocalDate shift(LocalDate base, int sequenceNumber) {
-      int month = base.getMonthValue();
-      int offset = (month % 3 == 0 ? 0 : 3 - month % 3) + (sequenceNumber - 1) * 3;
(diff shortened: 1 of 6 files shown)


modules/basics/src/main/java/com/opengamma/strata/basics/date/StandardDateSequences.java
```diff
@@ -64,6 +64,114 @@ public LocalDate dateMatching(YearMonth yearMonth) {
...
+    }
+
+    private LocalDate shift(LocalDate base, int sequenceNumber) {
+      // first 4 serial can be expanded to first 6 serial by subsuming first two quarterly
+      if (sequenceNumber <= 6) {
+        return base.plusMonths(sequenceNumber - 1).with(THIRD_WEDNESDAY);
+      }
+      int effectiveQuarterlySequenceNumber = sequenceNumber - 4;
+      int month = base.getMonthValue();
+      int offset = (month % 3 == 0 ? 0 : 3 - month % 3) + (effectiveQuarterlySequenceNumber - 1) * 3;
```
wjnicholson, line 111: Finding it hard to read that first ternary operation with operator precedence, might be easier to read in separate statements?
That first part is the offset to the next quarterly, and the second is the remaining quarterlies?

wjnicholson (approved):

PR #2215, opened 2020-09-04
Add SGD-SORA index

jodastephen:
SORA is the alternative reference rate for Singapore
Compounded SORA calculated from historic SORA values has not been added
https://www.mas.gov.sg/monetary-policy/sora


marc-henrard (approved):
Good for me.

PR #2216, opened 2020-09-04
Add overnight futures to calibration

jodastephen:
Follow the existing coding pattern


modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/PresentValueCalibrationMeasure.java
```diff
@@ -66,7 +69,17 @@
           (trade, p) -> DiscountingIborFutureTradePricer.DEFAULT.presentValueSensitivity(trade, p));
 
   /**
-   * The calibrator for {@link SwapTrade} using par spread discounting.
+   * The calibrator for {@link OvernightFutureTrade} using present value discounting.
+   */
+  public static final PresentValueCalibrationMeasure<ResolvedOvernightFutureTrade> OVERNIGHT_FUTURE_PV =
+      PresentValueCalibrationMeasure.of(
+          "OvernightFutureParSpreadDiscounting",
```
wjnicholson, line 76: These constants still all have "ParSpread" in the name - is that meaningful or need to be changed?

wjnicholson (approved):

modules/pricer/src/main/java/com/opengamma/strata/pricer/curve/PresentValueCalibrationMeasure.java
jodastephen: Good spot

PR #2221, opened 2020-09-09
Correct bug in OvernightFutureCurveNode end date.

marc-henrard:
End date was on the end of the first ON period in the composition instead of the last ON period.


marc-henrard:
Improvements by Stephen are good for me.

jodastephen (approved):

PR #2226, opened 2020-09-16
Parse bracketed numbers as negative numbers

anand-uk:


modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -234,14 +234,22 @@ public static BigDecimal parseBigDecimal(String str) {
...
     } catch (NumberFormatException ex) {
       NumberFormatException nfex = new NumberFormatException("Unable to parse BigDecimal percentage from '" + str + "'");
       nfex.initCause(ex);
       throw nfex;
     }
   }
 
+  private static String normalizeIfBracketed(String value) {
+    if (value.startsWith("(") && value.endsWith(")")) {
+      return "-" + value.replace("(", "").replace(")", "");
```
jodastephen, line 247: If the input string is "(123(456)" this approach will strip the middle bracket and parse when it shouldn't. You'll need to use `substring()`

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -166,7 +166,7 @@ public static boolean parseBoolean(String str) {
    */
   public static int parseInteger(String str) {
     try {
-      return Integer.parseInt(str);
+      return Integer.parseInt(normalizeIfBracketed(str));
```
jodastephen, line 169: This change should be added to the Javadoc of each method.

jodastephen (changes requested):

modules/loader/src/test/java/com/opengamma/strata/loader/LoaderUtilsTest.java
```diff
@@ -69,6 +69,14 @@ public void test_parseBoolean() {
   @Test
   public void test_parseInteger() {
     assertThat(LoaderUtils.parseInteger("2")).isEqualTo(2);
+    assertThat(LoaderUtils.parseInteger("(2)")).isEqualTo(-2);
+    assertThat(LoaderUtils.parseInteger("(23)")).isEqualTo(-23);
+    assertThatIllegalArgumentException()
+        .isThrownBy(() -> LoaderUtils.parseInteger("()"))
+        .withMessage("Unable to parse integer from '()'");
```
jodastephen, line 76: With AssertJ you have to put the `withMessage` before the method that actually throws, otherwise the message is not picked up. Needs fixing in all the methods below.

jodastephen (changes requested):

modules/loader/src/test/java/com/opengamma/strata/loader/LoaderUtilsTest.java
jodastephen: My mistake. This is correctly testing the message that is thrown.

jodastephen (approved):

PR #2249, opened 2020-11-03
Don't set roll convention when creating accrual periods for CNY indices with 1W reset frequency.

daminuk:
* When considering CNY indices with a 1W reset frequency don't set the roll convention, this results in an appropriate weekly roll convention being selected for each sub-schedule.


modules/product/src/main/java/com/opengamma/strata/product/swap/IborRateCalculation.java
```diff
@@ -319,7 +322,9 @@ public void collectIndices(ImmutableSet.Builder<Index> builder) {
     // resolve against reference data once
     DateAdjuster fixingDateAdjuster = fixingDateOffset.resolve(refData);
     Function<SchedulePeriod, Schedule> resetScheduleFn =
-        getResetPeriods().map(rp -> rp.createSchedule(accrualSchedule.getRollConvention(), refData)).orElse(null);
+        getResetPeriods().map(rp -> getIndex().getCurrency().equals(CNY) && rp.getResetFrequency().equals(P1W) ?
```
jodastephen, line 325: I think we can be more general here. If `accrualSchedule.getFrequency().isMonthBased()` and `rp.getResetFrequency().isWeekBased()` then...

modules/product/src/test/java/com/opengamma/strata/product/swap/IborRateCalculationTest.java
```diff
@@ -889,6 +901,69 @@ public void test_expand_gearingSpreadEverythingElse() {
...
+  public void test_createAccrualPeriodsCny1wFixingStub() {
+    BusinessDayAdjustment businessDayAdjustment = BusinessDayAdjustment.of(FOLLOWING, SAT_SUN);
+    TenorAdjustment tenorAdjustment =
+        TenorAdjustment.of(Tenor.TENOR_1W, PeriodAdditionConventions.NONE, businessDayAdjustment);
+
+    ImmutableIborIndex index = ImmutableIborIndex.builder()
+        .currency(Currency.CNY)
+        .dayCount(ACT_365F)
+        .name("CNY-REPO-1W")
+        .fixingCalendar(SAT_SUN)
```
jodastephen, line 915: No idea what happens if there is a holiday. Maybe the test case should cover that?

jodastephen (approved):

PR #2252, opened 2020-11-13
Add exception throwing for curves with no nodes in RatesCalibrationCsvLoader

jfcliang:
In the `RatesCalibrationCsvLoader.java` of the `loader` module, if there is a curve in the nodes data `calibration.csv` but the same name doesn't appear in `settings.csv `, there will be an error message `"Missing settings for curve: {name}"`. However, if it's the other way around (i.e., nodes data with no curve setting in the `settings.csv`), only a null pointer exception will be thrown, thus making it not as easy to spot and fix the error.  

Added a check in the `RatesCalibrationCsvLoader` so that  if there is any curve in the `settings.csv` without a matching nodes definition, a similar exception will be thrown indicating there are curve nodes data missing. Also added a test case in the `RatesCalibrationCsvLoaderTest` and corresponding input file `calibration-no-nodes.csv` with missing nodes info for the "USD-Disc" curve.


jodastephen (approved):
Thanks!

PR #2254, opened 2020-11-20
Parse an OG-ETD identifier

jodastephen:
Sometimes it can be useful to parse the elements back out
This wasn't originally intended to be supported, but it is possible


wjnicholson (approved):

wjnicholson:
What's the driver behind adding the functionality? I thought the original reasoning was that you should be looking up the information from elsewhere, rather than parsing the ID. Could we not store this information as a fully-fledged object under an attribute rather than parsing an ID?

jodastephen:
We have code elsewhere that does this., and I'd prefer to centralise it. That doesn't mean the parsing is endorsed as a great idea.

PR #2255, opened 2020-11-24
Add the ability to combine schedule periods if requested.

daminuk:
* Add a field to PeriodicSchedule to provide the ability to remove duplicated adjusted dates.
* Use this in the `IborRateCalculation` when we consider week based reset frequencies (to deal with cases like China's Golden Week).


modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
```diff
@@ -109,9 +106,8 @@
  * When the unadjusted schedule has been determined, the appropriate business day adjustment
  * is applied to create a parallel schedule of "adjusted" dates.
  */
-@BeanDefinition
-public final class PeriodicSchedule
-    implements ImmutableBean, Serializable {
+@BeanDefinition(style = "minimal")
```
daminuk, line 109: Switched this over to minimal to stop us coming up against our checkstyle 2000 line rule.

modules/basics/src/main/java/com/opengamma/strata/basics/schedule/SchedulePeriod.java
```diff
@@ -227,19 +227,21 @@ public boolean contains(LocalDate date) {
...
    * @return the sub-schedule
    * @throws ScheduleException if the schedule cannot be created
    */
   public PeriodicSchedule subSchedule(
       Frequency frequency,
       RollConvention rollConvention,
       StubConvention stubConvention,
-      BusinessDayAdjustment adjustment) {
+      BusinessDayAdjustment adjustment,
+      boolean combinePeriodIfNecessary) {
```
jodastephen, line 244: Strata needs backwards compatibility in places like this. You'll have to overload. I'm also not clear why the flag needs to be in here.

modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
```diff
@@ -424,6 +425,17 @@ public Schedule createSchedule(ReferenceData refData) {
     List<LocalDate> adj = applyBusinessDayAdjustment(unadj, refData);
     List<SchedulePeriod> periods = new ArrayList<>();
     try {
+      // Remove any duplicate adjusted dates if requested
+      if (combinePeriodsIfNecessary) {
```
jodastephen, line 429: Looking at how the code turned out, it seems my analysis was wrong, sorry. I don't think it is right to add `combinePeriodsIfNecessary` as an extra property in `PeriodicSchedule` since it can be handled just as well as an extra parameter to `createSchedule()` (with a backwards compatibility overload).

This has the side effect that `PeriodicSchedule` can stay as a normal bean, removing another incompatibility.

jodastephen (changes requested):

modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java
daminuk: No sweat, does work out cleaner with the overloaded method. Though wasn't a significant saving in terms of number of new lines, so I've tried to making savings elsewhere sits at 1999 lines :flushed:

jodastephen (approved):
The compressed comments are less readable, but the only way I can find to turn off the checkstyle check is to put `CSOFF: ALL` as the first line, above the header, which isn't ideal. As such, I think we go with the compressed version for now!

PR #2260, opened 2020-12-07
Add CMS market data lookup

benhamilton-brown:


modules/measure/src/main/java/com/opengamma/strata/measure/swaption/DefaultSwaptionMarketDataLookup.java
```diff
@@ -192,11 +222,7 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    StringBuilder buf = new StringBuilder(64);
-    buf.append("DefaultSwaptionMarketDataLookup{");
-    buf.append("volatilityIds").append('=').append(JodaBeanUtils.toString(volatilityIds));
-    buf.append('}');
-    return buf.toString();
+    return "DefaultSwaptionMarketDataLookup{" + "volatilityIds" + '=' + JodaBeanUtils.toString(volatilityIds) + '}';
```
benhamilton-brown, line 225: Intellij warned me about replacing string builder with string. Wasn't too sure about it though, but if there is no performance issues it odes look more readable.

benhamilton-brown:
Also committed merging implementation separately so can revert if need be

jodastephen:
Closing this at present

PR #2270, opened 2020-12-22
Date fix for calculating a running OIS coupon when valuation date falls on a holiday

ChristianOG:
Fixes issues arising when trying to value a running OIS coupon (i.e., with past fixings) when the valuation date happens to fall on a holiday. Usually when an OIS fixing is just before the valuation date, the publication date will fall on the valuation date: in this case Strata is able to determine the the next fixing date. However, in the case where the valuation date lies on a holiday, the publication date will be the next good business day after the valuation date, not the valuation date itself. As Strata currently only checks for equality of the publication date and the valuation date, it interprets this edge case as the publication date lying before the valuation date, so does not calculate either the current overnight rate nor the next fixing date.

This fix checks for the edge case when the valuation date lies between the current fixing date and the publication date, in which case it is still able to calculate the current overnight rate (which effectively straddles the valuation date) as well as the next fixing date.


akshaishah:
@jodastephen @ChristianOG I've tested this change for the scenario we encountered previously and can confirm this fixes the issue

jodastephen (approved):

PR #2271, opened 2020-12-29
update LoaderUtils to parse numbers with commas

anand-uk:


modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -265,6 +272,27 @@ private static String normalizeIfBracketed(String value) {
...
+      decimalPart = "";
+    } else {
+      integerPart = value.substring(0, decimalIndex);
+      decimalPart = value.substring(decimalIndex);
+    }
+
+    // only remove commas in the integer part, if those commas are followed by pattern "###," or ends with pattern "###"
+    // E.g. "1,234,567" => "123456" or "1,234" => "1234"
+    // Should ignore incorrectly placed commas, e.g. "1,234,57" => "1234,57" or "1,234," => "1234,"
+    String normalizedIntegerPart = integerPart.replaceAll("(?<=\\d),(?=\\d{3}(,|$))", "");
```
jodastephen, line 291: If a regex is being used, it should be a static constant, but regexes tend to be quite slow. Simply removing commas from the input string would be simpler and probably good enough.

jodastephen (changes requested):

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
anand-uk: in your view would it acceptable to parse `1,234,56` as `123456` or should it fail?

jodastephen: Seems like the simplest is 123456. Seems fine so long as it is specified in the method Javadoc

modules/loader/src/main/java/com/opengamma/strata/loader/LoaderUtils.java
```diff
@@ -265,6 +292,17 @@ private static String normalizeIfBracketed(String value) {
     }
   }
 
+  private static String normalizeIfCommaSeparated(String value) {
+    if (!value.startsWith(",") && !value.startsWith("-,") && !value.endsWith(",") && !value.contains(",,")) {
+      String[] parts = value.split("\\.");
```
jodastephen, line 297: Setup a private static constant for `Splitter.on('.')` and use it here to avoid creating a regex each time

jodastephen (changes requested):

jodastephen (approved):

PR #2278, opened 2021-01-14
Add isPaymentFixedAt for SwapPaymentEvents

benhamilton-brown:


jodastephen (approved):
Test case names should be updated too

PR #2282, opened 2021-01-25
Allow any curve to be used in PriceIndexValues

jodastephen:
Originally it was restricted to just NodalCurve
That blocked parallel shifts

This change does introduce a source and binary incompatibility, but most users will be using the interface, not the concrete class.


marc-henrard (approved):
Good for me

PR #2285, opened 2021-02-01
Add a plugin for Position CSV parsing

jodastephen:
Allow application code to plugin their own products
Approach is more standard than the resolver used up to now

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/EtdFuturePositionCsvPlugin.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/EtdFuturePositionCsvPlugin.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2021 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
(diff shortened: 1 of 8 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/csv/EtdFuturePositionCsvPlugin.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.common.collect.ImmutableSet;
+import com.opengamma.strata.collect.io.CsvRow;
+import com.opengamma.strata.product.Position;
+import com.opengamma.strata.product.PositionInfo;
+import com.opengamma.strata.product.SecurityPosition;
+import com.opengamma.strata.product.etd.EtdFuturePosition;
+
+/**
+ * Handles the CSV file format for CDS index trades.
```
skitini, line 19: Very minor but the JavaDoc for the different position csv plugin classes all refer to CDS index trades

skitini (changes requested):
Just the JavaDoc changes for the four plugin classes

skitini (approved):

PR #2290, opened 2021-02-08
New barrier type. Barrier constant on different periods.

marc-henrard:
Pieceiwse constant barrier. Barrier level can change on a given list of dates.


marc-henrard:
PR in another repository

PR #2293, opened 2021-02-19
Update japanese calendar

marcwrobel:
Due to the postponement of the Olympics Marine Day, Sports Day and Mountain Day has been moved again in 2021.


jodastephen (approved):
Thanks

PR #2295, opened 2021-02-23
Implementation of wrapper class CurrencyParameterSensitivitiesScenarioArray

ChristianOG:
Allows for use of CalculationRunner when calculating vega for scenarios.
Updated Vega measure calculations to return this type for scenario calculations.


ChristianOG:
Wrapper class needed to calculate scenario trade vegas with CalculationRunner. Would have put it with the other ScenarioArray wrapper classes in strata-data, but including CurrencyParameterSensitivities class causes a cyclic dependency between strata-market and strata-data.

jodastephen:
Adopting a rule that `ScenarioArray` should only have subclasses where the subclass provides optimised data storage. Opened #2298 instead of this one.

PR #2299, opened 2021-03-05
Replace BiFunction<K, V, Boolean> with BiPredicate<K, V> in MapStream

wjnicholson:


modules/collect/src/main/java/com/opengamma/strata/collect/MapStream.java
```diff
@@ -277,8 +277,8 @@ private MapStream(Stream<Map.Entry<K, V>> underlying) {
    * @param predicate  a predicate function applied to each key and value in the stream
    * @return a stream including the entries for which the predicate function returned true
    */
-  public MapStream<K, V> filter(BiFunction<? super K, ? super V, Boolean> predicate) {
-    return wrap(underlying.filter(e -> predicate.apply(e.getKey(), e.getValue())));
+  public MapStream<K, V> filter(BiPredicate<? super K, ? super V> predicate) {
```
cjkent, line 280: Is this binary backwards compatible? @jodastephen?

wjnicholson: Presumably not

cjkent: I presume so to. The question is whether we're happy to break binary backwards compatibility for this

jodastephen: Its not compatible. Does it cause any problems if left as is?

wjnicholson: No - I was just attempting to clean up and remove the boxing. 
Would it be detrimental if I added an overload with the changed signature?

jodastephen: It is a well known bad idea to overload methods that take a lambda (or at the very least, it needs great care). I think this change could be made, but we would need to bump to the next minor version (not patch version). Maybe leave this open and we try and merge it when we are next bumping that version

jodastephen (approved):

jodastephen (approved):

PR #2301, opened 2021-03-09
Provides access to parameters used in the pricer.

marc-henrard:
Swap pricer and parameters underlying the CMS pricer were not accessible from outside. Create getters.

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/swap/FxResetCalculation.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/swap/FxResetCalculation.java
@@ -146,7 +146,7 @@ public final class FxResetCalculation
    * @throws ReferenceDataNotFoundException if an identifier cannot be resolved in the reference data
    * @throws RuntimeException if the calculation is invalid
    */
-  BiFunction<Integer, SchedulePeriod, Optional<FxReset>> resolve(ReferenceData refData) {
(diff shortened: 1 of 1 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/cms/SabrExtrapolationReplicationCmsPeriodPricer.java
```diff
@@ -144,7 +144,35 @@ public static SabrExtrapolationReplicationCmsPeriodPricer of(
   public static SabrExtrapolationReplicationCmsPeriodPricer of(double cutOffStrike, double mu) {
     return of(DiscountingSwapProductPricer.DEFAULT, cutOffStrike, mu);
   }
+  
+  /**
+   * Returns the underlying swap pricer.
+   * 
+   * @return the pricer
+   */
+  public DiscountingSwapProductPricer swapPricer() {
```
jodastephen, line 153: To be consistent with other pricers, these methods should be `getXxx`()` and be located after the constructor

jodastephen (changes requested):

modules/pricer/src/main/java/com/opengamma/strata/pricer/cms/SabrExtrapolationReplicationCmsPeriodPricer.java
marc-henrard: Done

PR #2310, opened 2021-04-13
Fix lastBusinessDayOfMonth() @return javadoc

ebourg:
The javadoc of `lastBusinessDayOfMonth()` was probably copy/pasted from `isLastBusinessDayOfMonth()` but the description of the returned value was not modified.


jodastephen (approved):
Thanks

PR #2317, opened 2021-06-11
Add Fx Option Trade and Fx Option Product

santorog:
The notion of FxOptionTrade as a marker interface would be useful, hence this merge request.

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOption.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOption.java
@@ -6,6 +6,7 @@
 package com.opengamma.strata.product.fxopt;
 
 import java.io.Serializable;
+import java.time.ZonedDateTime;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


jodastephen (changes requested):
There also needs to be an `FxOptionProduct` interface that extends `FxProduct` and adds a `getExpiry() method. `FxOptionProduct` would then override `getProduct()` to return `FxOptionProduct`

modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOption.java
```diff
@@ -115,14 +116,22 @@ private void validate() {
...
     return underlyingOption.getCurrencyPair();
   }
 
+  //-----------------------------------------------------------------------
+  /**
+   * Gets the underlying Fx vanilla option's expiry.
+   * @return the expiry
+   */
+  @Override
+  public ZonedDateTime getExpiry() { return underlyingOption.getExpiry(); }
```
jodastephen, line 133: This formatting is wrong.

modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOption.java
```diff
@@ -59,15 +60,15 @@
...
+   * The barrier level stored in this field must be represented based on the direction of the currency pair in the
    * underlying FX transaction.
    * <p>
    * For example, if the underlying option is an option on EUR/GBP, the barrier should be a certain level of EUR/GBP rate.
    */
   @PropertyDefinition(validate = "notNull")
   private final Barrier barrier;
   /**
-   * For a 'out' option, the amount is paid when the barrier is reached; 
+   * For a 'out' option, the amount is paid when the barrier is reached;
```
jodastephen, line 71: Your IDE should not be configured to automatically remove whitespace like this. (No need to undo the whitespace parts)

modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxVanillaOption.java
```diff
@@ -49,7 +49,7 @@
  */
 @BeanDefinition
 public final class FxVanillaOption
-    implements FxProduct, Resolvable<ResolvedFxVanillaOption>, ImmutableBean, Serializable {
+    implements FxOptionProduct, Resolvable<ResolvedFxVanillaOption>, ImmutableBean, Serializable {
```
jodastephen, line 52: Need to add `@Override` to `getExpiry()`

jodastephen (changes requested):

modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOption.java
santorog: I followed the dev guide to configured my IDE and just double-checked I did that correctly earlier, I am not sure what is happening.

jodastephen (approved):

PR #2318, opened 2021-06-24
Create CapFloor parser and writer

skitini:


modules/loader/src/main/java/com/opengamma/strata/loader/csv/IborCapFloorTradeCsvPlugin.java
```diff
@@ -0,0 +1,330 @@
...
+
+    if (requiredJavaType.isAssignableFrom(IborCapFloorTrade.class)) {
+      return Optional.of(resolver.parseCapFloorTrade(baseRow, info));
+    }
+    return Optional.empty();
+  }
+
+  @Override
+  public String getName() {
+    return "CapFloor";
```
jodastephen, line 131: I think this should be `IborCapFloor`. The name is generally the trade class minus "Trade"

modules/loader/src/main/java/com/opengamma/strata/loader/csv/IborCapFloorTradeCsvPlugin.java
```diff
@@ -0,0 +1,330 @@
...
+  // parse the upfront premium amount
+  private static Optional<AdjustablePayment> parsePremium(CsvRow row) {
+    Optional<Double> premiumAmountOpt = row.findValue(PREMIUM_AMOUNT_FIELD)
+        .map(LoaderUtils::parseDouble);
+    if (premiumAmountOpt.isPresent()) {
+      Currency premiumCurrency = row.getValue(PREMIUM_CURRENCY_FIELD, LoaderUtils::parseCurrency);
+      PayReceive premiumDirection = row.getValue(PREMIUM_DIRECTION_FIELD, LoaderUtils::parsePayReceive);
+      LocalDate premiumDate = row.getValue(PREMIUM_DATE_FIELD, LoaderUtils::parseDate);
+      CurrencyAmount premiumAmount = CurrencyAmount.of(premiumCurrency, premiumAmountOpt.get());
+      AdjustablePayment payment = premiumDirection.isPay() ? AdjustablePayment.ofPay(premiumAmount, premiumDate)
```
jodastephen, line 252: When breaking a tertiary onto multiple lines there should be a new line after the ? and after the :

modules/loader/src/main/java/com/opengamma/strata/loader/csv/IborCapFloorTradeCsvPlugin.java
```diff
@@ -0,0 +1,330 @@
...
+    IborCapFloorLeg.Builder capFloorLegBuilder = IborCapFloorLeg.builder()
+        .payReceive(payReceive)
+        .paymentSchedule(paymentSchedule)
+        .currency(currency)
+        .notional(ValueSchedule.of(notional))
+        .calculation(IborRateCalculation.of(iborIndex));
+    String capFloorType = row.getValue(CAP_FLOOR_FIELD);
+    if (capFloorType.toUpperCase(Locale.ENGLISH).equals("CAP")) {
+      capFloorLegBuilder.capSchedule(strike);
+    } else if (capFloorType.toUpperCase(Locale.ENGLISH).equals("FLOOR")) {
```
jodastephen, line 182: Looking at the API, I would assume that a leg can have both a cap and a floor. There does seem to be validation against it though. Might need to investigate changing things to a single schedule and a type in another PR.

jodastephen (changes requested):

jodastephen (approved):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/IborCapFloorTradeCsvPlugin.java
brianweller89: A cap/floor would always only have one or the other; the payoff is (Rate fixing - Strike) or (Strike - Rate Fixing)

skitini: The `IborCapFloorLeg` only lets you have one, but I think the point is that it has a cap schedule and floor schedule as optional properties (even though there is validation that won't allow it). It might be better having a `CapFloor` property and then just a single schedule, and the payoff direction can be determined by the `CapFloor` type, which also aligns with how the parser works/ is represented in CSV format.

PR #2319, opened 2021-06-25
Create FX NDF parser and writer

skitini:


modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
```diff
@@ -0,0 +1,159 @@
...
+  // parses the row to a trade
+  private static FxNdfTrade parseRow(CsvRow row, TradeInfo info) {
+    CurrencyAmount settlementCurrencyNotional = CsvLoaderUtils.parseCurrencyAmountWithDirection(
+        row, SETTLEMENT_CURRENCY_FIELD, SETTLEMENT_CURRENCY_NOTIONAL_FIELD, SETTLEMENT_CURRENCY_DIRECTION_FIELD);
+    LocalDate paymentDate = row.getField(PAYMENT_DATE_FIELD, LoaderUtils::parseDate);
+    Currency settlementCurrency = row.getField(SETTLEMENT_CURRENCY_FIELD, LoaderUtils::parseCurrency);
+    Currency nonDeliverableCurrency = row.getField(NON_DELIVERABLE_CURRENCY_FIELD, LoaderUtils::parseCurrency);
+    CurrencyPair currencyPair = CurrencyPair.of(settlementCurrency, nonDeliverableCurrency).toConventional();
+    FxRate agreedFxRate = FxRate.of(currencyPair, row.getField(FX_RATE_FIELD, LoaderUtils::parseDouble));
+    FxIndex.extendedEnum();
```
jodastephen, line 110: Does this line do anything?

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
```diff
@@ -0,0 +1,159 @@
...
+    return FxNdfTrade.of(info, fxNdf);
+  }
+
+  // parses the FX Index from a currency pair
+  private static Optional<FxIndex> parseFxIndex(CurrencyPair currencyPair) {
+    ImmutableMap<CurrencyPair, FxIndex> ccyFxIndexMap = FxIndex.extendedEnum().lookupAll().values().stream()
+        .collect(toImmutableMap(
+            fxIndex -> fxIndex.getCurrencyPair().toConventional(),
+            fxIndex -> fxIndex,
+            (fxIndex1, fxIndex2) -> fxIndex1)); //arbitrarily chose when we have multiple indices for the same pair
```
jodastephen, line 130: The FX index name should be in the csv format. Potentially it should be in the convention column. If this is done, there is probably no need for the separate settlement and ND currency columns.

There is also a good case for re-using the Buy/Sell column, instead of adding the new Pay/Receive one

modules/loader/src/test/resources/com/opengamma/strata/loader/csv/fxtrades.csv
```diff
@@ -1,8 +1,10 @@
...
-FX,CME,OG,tradeId1,06/12/2016,,,USD/INR,Sell,,USD,3850000,67.4,08/12/2016,Following,USNY,,,,,,,,,,,,
-FX,CME,OG,tradeId2,22/12/2016,,,EUR/CZK,Buy,,EUR,1920000,25.62,24/12/2016,,,,,,,,,,,,,,
-FxSwap,CME,OG,tradeId11,06/12/2016,,,USD/CAD,Buy,,USD,120000,1.31,08/12/2016,,,1.34,08/01/2017,,,,,,,,,,
-Fx Swap,CME,OG,tradeId12,06/12/2016,,,USD/CAD,Sell,,CAD,-160000,1.32,08/12/2016,,,1.34,08/01/2017,,,,,,,,,,
-Bullet,,OG,tradeId21,06/12/2016,,,,,Receive,USD,25000,,08/12/2016,Following,USNY,,,,,,,,,,,,
-BulletPayment,,OG,tradeId22,22/12/2016,,,,,Pay,GBP,35000,,24/12/2016,,,,,,,,,,,,,,
-FxVanillaOption,CME,OG,tradeId31,06/12/2016,,,USD/CAD,Buy,,USD,30000,1.31,10/01/2017,,,,,Long,08/01/2017,11:00,Europe/London,Pay,GBP,2000,08/12/2016,,
+Strata Trade Type,CCP,Id Scheme,Id,Trade Date,Trade Time,Trade Zone,Convention,Buy Sell,Direction,Currency,Notional,FX Rate,Payment Date,Payment Date Convention,Payment Date Calendar,Far FX Rate,Far Payment Date,Long Short,Expiry Date,Expiry Time,Expiry Zone,Premium Direction,Premium Currency,Premium Amount,Premium Date,Premium Date Convention,Premium Date Calendar,Settlement Currency,Settlement Currency Direction,Settlement Currency Notional,ND Currency
```
jodastephen, line 1: You've got two "Notional" columns

jodastephen (changes requested):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
brianweller89: The way we always end up parsing these trades is to take two currencies and work out the index ourselves; typically the  specific index is not something that clients will have available in their trade capture systems (it will be part of trade confirms). 

Whilst technically Strata is correct in terms of modelling the trade it is not really practical to force all users to map to the index names that we recognise.

We could put this in a private repo instead?

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
skitini: removed

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
jodastephen: What is needed is a way to allow a currency pair to be entered in the convention column and mapped to a suitable FX index (ideally just a generic one that doesn't provide any more detail). When writing it out, and for round tripping, the full index name must be used.

skitini: Perhaps we try to search the `FxIndexData.csv` and if not present, we create using a `ImmutableFxRate` with suitable defaults?

jodastephen: Yes, `FxIndexCsvLookup` can probably be extended to support creating indices on the fly for plain currency-pairs

modules/basics/src/main/java/com/opengamma/strata/basics/index/FxIndex.java
```diff
@@ -47,7 +49,47 @@
   @FromString
   public static FxIndex of(String uniqueName) {
     ArgChecker.notNull(uniqueName, "uniqueName");
-    return extendedEnum().lookup(uniqueName);
+    try {
+      return extendedEnum().lookup(uniqueName);
```
jodastephen, line 53: Use `find(uniqueName)` to avoid catching the exception

modules/basics/src/main/java/com/opengamma/strata/basics/index/FxIndex.java
```diff
@@ -47,7 +49,47 @@
...
+    } catch (IllegalArgumentException ex) {
+      CurrencyPair currencyPair = CurrencyPair.parse(uniqueName);
```
jodastephen, line 55: Needs to be a try-catch around this line. The idea is that the error message should refer to "index not found", not "unable to parse currency pair".

modules/basics/src/main/java/com/opengamma/strata/basics/index/FxIndex.java
```diff
@@ -47,7 +49,47 @@
...
+   * If a currency pair does not have an implementation, an FX index will be created.
+   *
+   * @param currencyPair the currency pair
+   * @return the index
+   */
+  public static FxIndex of(CurrencyPair currencyPair) {
+    ArgChecker.notNull(currencyPair, "currencyPair");
+    return extendedEnum().lookupAll().values().stream()
+        .filter(index -> index.getCurrencyPair().equals(currencyPair))
+        .findFirst()
```
jodastephen, line 72: Without a sort finding the first is essentially random. Maybe alpha sort by index name first (eg. there are two EUR/USD indices defined)

modules/loader/src/test/resources/com/opengamma/strata/loader/csv/fxtrades.csv
```diff
@@ -1,8 +1,11 @@
...
-FX,CME,OG,tradeId2,22/12/2016,,,EUR/CZK,Buy,,EUR,1920000,25.62,24/12/2016,,,,,,,,,,,,,,
-FxSwap,CME,OG,tradeId11,06/12/2016,,,USD/CAD,Buy,,USD,120000,1.31,08/12/2016,,,1.34,08/01/2017,,,,,,,,,,
-Fx Swap,CME,OG,tradeId12,06/12/2016,,,USD/CAD,Sell,,CAD,-160000,1.32,08/12/2016,,,1.34,08/01/2017,,,,,,,,,,
-Bullet,,OG,tradeId21,06/12/2016,,,,,Receive,USD,25000,,08/12/2016,Following,USNY,,,,,,,,,,,,
-BulletPayment,,OG,tradeId22,22/12/2016,,,,,Pay,GBP,35000,,24/12/2016,,,,,,,,,,,,,,
-FxVanillaOption,CME,OG,tradeId31,06/12/2016,,,USD/CAD,Buy,,USD,30000,1.31,10/01/2017,,,,,Long,08/01/2017,11:00,Europe/London,Pay,GBP,2000,08/12/2016,,
+Strata Trade Type,CCP,Id Scheme,Id,Trade Date,Trade Time,Trade Zone,Convention,Buy Sell,Direction,Currency,Notional,FX Rate,Payment Date,Payment Date Convention,Payment Date Calendar,Far FX Rate,Far Payment Date,Long Short,Expiry Date,Expiry Time,Expiry Zone,Premium Direction,Premium Currency,Premium Amount,Premium Date,Premium Date Convention,Premium Date Calendar,Settlement Currency,Settlement Currency Direction,Settlement Currency Notional,ND Currency,FX Index
```
jodastephen, line 1: "FX index" column should just go in the "Convention" column.

Rather than "Settlement Currency" and "Settlement Currency Notional" just re-use "Notional" and "Currency" columns. Direction too. Minimising the number of columns is important.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
```diff
@@ -0,0 +1,157 @@
...
+        .paymentDate(paymentDate)
+        .build();
+    return FxNdfTrade.of(info, fxNdf);
+  }
+
+  // parses the FX Index
+  private static FxIndex parseFxIndex(CsvRow row, CurrencyPair currencyPair) {
+    // prioritize value in fx index field if present
+    try {
+      return FxIndex.of(row.getValue(FX_INDEX_FIELD));
```
jodastephen, line 125: Avoid the try-catch where possible

`FxIndex.of(row.findValue(FX_INDEX_FIELD).orElse(currencyPair.toString()))`

jodastephen (changes requested):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxNdfTradeCsvPlugin.java
```diff
@@ -102,10 +102,10 @@ static FxNdfTrade parse(CsvRow row, TradeInfo info, TradeCsvInfoResolver resolve
   // parses the row to a trade
   private static FxNdfTrade parseRow(CsvRow row, TradeInfo info) {
     CurrencyAmount settlementCurrencyNotional = CsvLoaderUtils.parseCurrencyAmountWithDirection(
-        row, SETTLEMENT_CURRENCY_FIELD, SETTLEMENT_CURRENCY_NOTIONAL_FIELD, SETTLEMENT_CURRENCY_DIRECTION_FIELD);
+        row, LEG_1_CURRENCY_FIELD, LEG_1_NOTIONAL_FIELD, LEG_1_DIRECTION_FIELD);
```
jodastephen, line 105: So, my intent was that "leg 1" did not imply "settlement". ie. that a user could enter these three fields as leg 1 or leg 2, and that the parser would figure out which was the settlement currency based on the presence/absence of notional.

jodastephen (approved):

PR #2326, opened 2021-07-08
Update CDS Fpml parser to use the indexId element

beajeanm:
indexId, if provided, is a better option to indentify the CDS Index than
the free form index name.
We re-use the index scheme if provided, but default to
StandardSchemes.RED9_SCHEME if not. (This is probably what the id is, as
it's the de facto standard for CDS indices.)

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/impl/fpml/CdsFpmlParserPlugin.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/impl/fpml/CdsFpmlParserPlugin.java
@@ -202,8 +202,10 @@ final class CdsFpmlParserPlugin
     if (idOpt.isPresent()) {
       XmlElement idElement = idOpt.get();
       String id = idElement.getContent();
-      String indexIdScheme = idElement.findAttribute("indexIdScheme").orElse(StandardSchemes.RED9_SCHEME);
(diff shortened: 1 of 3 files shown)


modules/loader/src/main/java/com/opengamma/strata/loader/impl/fpml/CdsFpmlParserPlugin.java
```diff
@@ -194,6 +197,17 @@ Trade parseCds(FpmlDocument document, XmlElement tradeEl, TradeInfoBuilder trade
     throw new FpmlParseException("FpML CDS must be single name or index");
   }
 
+  private Optional<StandardId> tryParseIndexId(XmlElement indexInformation) {
+    Optional<XmlElement> idOpt = indexInformation.findChild("indexId");
+    if (idOpt.isPresent()) {
+      XmlElement idElement = idOpt.get();
+      String id = idElement.getContent();
+      String indexIdScheme = idElement.findAttribute("indexIdScheme").orElse(StandardSchemes.RED9_SCHEME);
```
jodastephen, line 205: I think this logic makes more sense:
* Is it length 9, then set as RED9
* Is the indexIdScheme attribute set, then use it
* Else return `Optional.empty`

jodastephen (changes requested):

jodastephen (approved):

PR #2327, opened 2021-07-08
Make premium optional for Fx options.

santorog:


modules/product/src/main/java/com/opengamma/strata/product/fxopt/FxSingleBarrierOptionTrade.java
```diff
@@ -66,7 +67,7 @@
   * The premium sign should be compatible with the product Long/Short flag.
   * This means that the premium is negative for long and positive for short.
   */
-  @PropertyDefinition(validate = "notNull")
+  @PropertyDefinition(get = "optional")
```
santorog, line 70: @jodastephen would need your review especially here.
I think it is the change needed.

santorog (commented):
Self review done

PR #2329, opened 2021-07-08
Adapt the CSV plugin to support empty premium fields.

santorog:


modules/loader/src/main/java/com/opengamma/strata/loader/csv/FxSingleBarrierOptionTradeCsvPlugin.java
```diff
@@ -97,7 +97,8 @@ private FxSingleBarrierOptionTrade parse(CsvRow row, TradeInfo info, TradeCsvInf
 
     return FxSingleBarrierOptionTrade.builder()
         .product(productBuilder.build())
-        .premium(premium)
+        .premium(CsvLoaderUtils.tryParsePremiumFromDefaultFields(row)
+            .orElse(AdjustablePayment.of(USD, 0d, vanillaOptionTrade.getProduct().getExpiryDate())))
```
jodastephen, line 101: This currency should be one of the two in the FX. I suggest it is `currencyPair.toConventional().getBaseCurrency()` (Hard coding USD may force additional market data to be loaded that we don't want.)

jodastephen (changes requested):

jodastephen (approved):

PR #2330, opened 2021-07-08
Fix incorrect methods for ZonedDateTime

santorog:
Zone and Date field have been inverted in the write method.

### DIFF
--- a/modules/loader/src/main/java/com/opengamma/strata/loader/csv/CsvWriterUtils.java
+++ b/modules/loader/src/main/java/com/opengamma/strata/loader/csv/CsvWriterUtils.java
@@ -22,7 +22,6 @@ import static com.opengamma.strata.loader.csv.CsvLoaderColumns.SECURITY_ID_SCHEM
 import static com.opengamma.strata.loader.csv.CsvLoaderColumns.TRADE_TYPE_FIELD;
 
 import java.time.LocalDate;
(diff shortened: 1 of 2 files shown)


santorog (commented):
Self review done.

modules/loader/src/main/java/com/opengamma/strata/loader/csv/CsvWriterUtils.java
```diff
@@ -131,11 +131,11 @@ public static void writeZonedDateTime(
       String timeField,
       String zoneField) {
 
-    csv.writeCell(dateField, zonedDateTime.getZone());
-    csv.writeCell(timeField, LocalTime.of(zonedDateTime.getHour(), zonedDateTime.getMinute()));
     csv.writeCell(
-        zoneField,
+        dateField,
```
jodastephen, line 135: Should be `zonedDateTime.toLocalDate()`

jodastephen (changes requested):

jodastephen (approved):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/CsvWriterUtils.java
santorog: Good spot. Thanks.

PR #2332, opened 2021-07-18
Proposal to Fix two Flaky Tests in ExtendedScenarioMarketDataTest

hwang-pku:
Two flaky tests in the class ExtendedScenarioMarket DataTest are found in this project and this PR proposes to fix them.

1. How to reproduce the flaky tests
- The project is built and tested under `javaJDK8`.
- The tool used for the flaky test detection is [NonDex](https://github.com/TestingResearchIllinois/NonDex)
- I ran `mvn -pl modules/data edu.illinois:nondex-maven-plugin:1.1.2:debug -Dtest=com.opengamma.strata.data.scenario.ExtendedScenarioMarketDataTest`, which produces
`java.lang.AssertionError: 
Actual and expected have the same elements but not in the same order, at index 0 actual element was: 2
whereas expected element was: 1
at com.opengamma.strata.data.scenario.ExtendedScenarioMarketDataTest.of_addition(ExtendedScenarioMarketDataTest.java:67)`
and
`java.lang.AssertionError: 
Actual and expected have the same elements but not in the same order, at index 0 actual element was: 2
whereas expected element was: 1
at com.opengamma.strata.data.scenario.ExtendedScenarioMarketDataTest.of_override(ExtendedScenarioMarketDataTest.java:85)`.

2. Why the tests failed
- In the implementation of the method `com.opengamma.strata.data.scenario.ExtendedScenarioMarketData.getIds()` in [`modules/data/src/main/java/com/opengamma/strata/data/scenario/ExtendedScenarioMarketData.java`](https://github.com/OpenGamma/Strata/blob/main/modules/data/src/main/java/com/opengamma/strata/data/scenario/ExtendedScenarioMarketData.java), a set is returned, which does not guarantee the iteration order.

3. The changes made
- Here we change the `containExactly` to `containExactlyInAnyOrder` in the corresponding two test methods in the test class `ExtendedScenarioMarketDataTest`.

### DIFF
--- a/modules/data/src/test/java/com/opengamma/strata/data/scenario/ExtendedScenarioMarketDataTest.java
+++ b/modules/data/src/test/java/com/opengamma/strata/data/scenario/ExtendedScenarioMarketDataTest.java
@@ -64,7 +64,7 @@ public class ExtendedScenarioMarketDataTest {
     assertThat(test.findValue(ID2)).isEqualTo(Optional.of(VAL2));
     assertThat(test.findValue(ID3)).isEqualTo(Optional.of(VAL3));
     assertThat(test.findValue(ID4)).isEqualTo(Optional.empty());
-    assertThat(test.getIds()).containsExactly(ID1, ID2, ID3);
+    assertThat(test.getIds()).containsExactlyInAnyOrder(ID1, ID2, ID3);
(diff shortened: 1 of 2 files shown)


jodastephen (approved):
I doubt that there is a real problem with the code because we use `ImmutableSet` extensively, and that guarantees order. Nevertheless, there is no harm in merging this change. Thanks

PR #2339, opened 2021-08-04
Force XAU to always be the base currency

akshaishah:


jodastephen (commented):
Obviously Gold sits above USD, but does it sit above EUR? Can't find any sources on the issue.

jodastephen (approved):

PR #2342, opened 2021-08-09
Add Cap/Floor parsing and related methods.

santorog:
Adding the CapFloor enum in strata, mirroring the BuySell enum.

Also adding some parsing utilities and CSV headers.


modules/loader/src/main/java/com/opengamma/strata/loader/csv/CsvLoaderColumns.java
```diff
@@ -46,24 +46,28 @@
   /** CSV header. */
   public static final String DIRECTION_FIELD = "Direction";
   /** CSV header. */
-  static final String LEG_1_DIRECTION_FIELD = "Leg 1 " + DIRECTION_FIELD;
+  public static final String LEG_1_DIRECTION_FIELD = "Leg 1 " + DIRECTION_FIELD;
```
santorog, line 49: I forgot to set these as public when I added the field in my previous merge request, a couple weeks ago.

benhamilton-brown (approved):
Looks fine. Might be worth finalising with Brian on header columns though, as he's trying to minimise the amount of new fields we have.

PR #2355, opened 2021-09-06
Fix time-series merge to correctly handle weekends

jodastephen:
The merge method did not set the `containsWeekends` flag
This caused an exception on building


modules/collect/src/main/java/com/opengamma/strata/collect/timeseries/LocalDateDoubleTimeSeriesBuilder.java
```diff
@@ -156,9 +155,14 @@ public LocalDateDoubleTimeSeriesBuilder merge(LocalDate date, double value, Doub
    * @return this builder
    */
   public LocalDateDoubleTimeSeriesBuilder merge(LocalDateDoublePoint point, DoubleBinaryOperator operator) {
-    ArgChecker.notNull(point, "point");
```
wjnicholson, line 159: Did you not want to leave the `notNull` check here? 
The equivalent `put(LocalDateDoublePoint point)` has a similar check before delegating

wjnicholson (approved):

wjnicholson (approved):

uuilee:
Thanks for fixing!

PR #2356, opened 2021-09-09
Enhance IborCapFloorTrade pricer

skitini:
- Add implied volatility and forward rate calculations to IBOR cap/floor trade, product, leg and period pricers.
- For the trade, product and leg pricers, a map is returned of the implied volatility / forward rate per individual caplet/floorlet.


benhamilton-brown (commented):
All looks fine, is it worth adding a few tests for some floor/floorlets?

benhamilton-brown (approved):

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/VolatilityIborCapFloorLegPricer.java
```diff
@@ -227,6 +230,44 @@ public CurrencyAmount currentCash(
...
 
+  //-------------------------------------------------------------------------
+  /**
+   * Calculates the forward rates for each caplet/floorlet of the Ibor cap/floor leg.
+   *
+   * @param capFloorLeg  the Ibor cap/floor leg
+   * @param ratesProvider  the rates provider
+   * @return the forward rates
+   */
+  public Map<IborCapletFloorletPeriod, Double> forwardRates(
```
jodastephen, line 241: Maybe there should be a bean to represent this map?

skitini: Would you recommend two beans, `IborCapletFloorletPeriodForwardRates` and `IborCapletFloorletPeriodVolatilities`, or just one bean that handles both, as they both just a map of period to double?

jodastephen: We tend to share classes in results, eg `CurrencyAmount`. So I think `IborCapletFloorletPeriodAmounts`

jodastephen (approved):
I suspect it may be possible to add methods to the new class, eg. ones that return `double` instead of `Double`, but experience can guide us with this.

PR #2361, opened 2021-09-17
More IborCapFloorTrade pricer enhancements

skitini:
- Change forward rate calculations to filter caplet/floorlet periods which have expired, consistent with implied volatility calculations
- Add method to get the present value for the caplet/floorlet periods. This method ignores any present value calculations for the product pay leg or the trade premium.
- Add currency amount map to the `IborCapletFloorletPeriodAmounts` bean, plus some additional methods and validation


modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/IborCapletFloorletPeriodAmounts.java
```diff
@@ -15,28 +15,101 @@
...
    */
   @PropertyDefinition(validate = "notNull")
-  private final ImmutableMap<IborCapletFloorletPeriod, Double> periodAmounts;
+  private final ImmutableMap<IborCapletFloorletPeriod, Double> periodDoubleAmounts;
+
+  /**
+   * The map of Ibor caplet/floorlet periods to the currency amount.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final ImmutableMap<IborCapletFloorletPeriod, CurrencyAmount> periodCurrencyAmounts;
```
jodastephen, line 48: This should be two separate beans, not one.

jodastephen (changes requested):

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/VolatilityIborCapFloorLegPricer.java
```diff
@@ -268,7 +294,7 @@ public IborCapletFloorletPeriodAmounts impliedVolatilities(
         .filterKeys(period -> volatilities.relativeTime(period.getFixingDateTime()) >= 0)
```
benhamilton-brown: maybe use rates provider here to check for expiration to keep consistent with forward rates

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/IborCapletFloorletPeriodCurrencyAmounts.java
```diff
@@ -0,0 +1,318 @@
...
+import com.opengamma.strata.basics.currency.CurrencyAmount;
+import com.opengamma.strata.product.capfloor.IborCapletFloorletPeriod;
+
+/**
+ * A map of currency amounts keyed by Ibor caplet/floorlet periods.
+ * <p>
+ * The currency amounts should represent the same metric.
+ */
+@BeanDefinition
+public final class IborCapletFloorletPeriodCurrencyAmounts implements ImmutableBean, Serializable {
```
benhamilton-brown, line 34: Is this only used for present values now? Might be more representative as IborCapletFloorletPeriodPresentValue?

jodastephen (approved):

benhamilton-brown (changes requested):
It may also be worth having some normal volatility tests of a similar ilk. I know it's not used in SIMM currently, but technically if added to genericvolatilitypricer normal should work if someone tried to use it

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/VolatilityIborCapFloorLegPricer.java
skitini: The filter here is reusing the same logic contained in the original implied vol calculation in the period pricer, and purely used to stop exceptions being thrown by the period pricer's ArgChecker when calling in the stream.

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/IborCapletFloorletPeriodCurrencyAmounts.java
skitini: Right now it is only used for present values, but I think it's more consistent this way with `IborCapletFloorletPeriodAmounts` which has doubles, but can be forward rates/implied vols. By keeping this name, it allows any currency amount to be stored against the periods.

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/VolatilityIborCapFloorLegPricer.java
skitini: Additionally, the `validate()` method checks that the rates provider and volatilities have the same valuation date, so it wouldn't offer any functionality.

benhamilton-brown: In strata a client could potentially pass an IborCapletFloorletVolatilities that is time dependent, so it can handle intraday. In this case volatilities.relativeTime(period.getFixingDateTime()) >= 0) might not equate to ratesProvider.getValuationDate().isAfter(period.getFixingDate()). This means we could have a case where implied volatility assumes trade hasn't expired where as for forward rates trade is presumed expired

benhamilton-brown (changes requested):
left comment on older comment

modules/pricer/src/main/java/com/opengamma/strata/pricer/capfloor/VolatilityIborCapFloorLegPricer.java
skitini: The check using the valuation date from the rates provider `forwardRates()` method is basically if the rates provider date is before or on - if a time were available then we'd check that too but the rates provider doesn't provide. If I check using the logic from the `forwardRates()` method instead of checking against the volatilities date and time, then if the volatilities are intraday, the method could throw an exception in the period pricer if we are on the fixing date but after the vol expiry.

benhamilton-brown (approved):

PR #2372, opened 2021-11-01
Correct bug for negative forward price in NormalFormulaRepository

marc-henrard:
Change starting point and maxChange in implied volatility computation to deal with negative forwards.

### DIFF
--- a/modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/option/NormalFormulaRepository.java
+++ b/modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/option/NormalFormulaRepository.java
@@ -279,15 +279,15 @@ public final class NormalFormulaRepository {
     if (Double.doubleToLongBits(optionPrice) == Double.doubleToLongBits(intrinsicPrice)) {
       return 0d;
     }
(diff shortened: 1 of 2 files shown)


modules/pricer/src/main/java/com/opengamma/strata/pricer/impl/option/NormalFormulaRepository.java
```diff
@@ -279,15 +279,15 @@ public static double impliedVolatility(
     if (Double.doubleToLongBits(optionPrice) == Double.doubleToLongBits(intrinsicPrice)) {
       return 0d;
     }
-    double sigma = (Math.abs(initialNormalVol) < 1e-10 ? 0.3 * forward : initialNormalVol);
-    double maxChange = 0.5 * forward;
+    double sigma = (Math.abs(initialNormalVol) < 1e-10 ? 1e-10 : initialNormalVol); // to start with a non zero vol
```
benhamilton-brown, line 282: can we use Math.min here?

benhamilton-brown (changes requested):

benhamilton-brown (approved):

PR #2375, opened 2021-11-11
Enhance MapStream

jodastephen:
- Add ability to get inverse
- Add ability to map key and value together

### DIFF
--- a/modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
+++ b/modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
@@ -57,24 +57,35 @@ public class MapStreamTest {
     assertThat(result).isEqualTo(ImmutableList.of(1, 2, 3, 4));
   }
 
+  @Test
   public void keys_fromList() {
     List<String> result = MapStream.of(list).keys().collect(toImmutableList());
     assertThat(result).isEqualTo(list);
   }
 
+  @Test
   public void values_fromList() {
     List<String> result = MapStream.of(list).values().collect(toImmutableList());
     assertThat(result).isEqualTo(list);
   }
 
+  @Test
   public void keys_fromStream() {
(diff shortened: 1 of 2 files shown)


modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
```diff
@@ -141,6 +152,14 @@ public void mapKeysAndValuesToValues() {
     assertThat(result).isEqualTo(expected);
   }
 
+  @Test
+  public void mapBoth() {
+    Map<String, Short> expected =
+        ImmutableMap.of("one - 1", (short) 0, "two - 1", (short) 1, "three - 1", (short) 2, "four - 1", (short) 3);
+    Map<String, Short> result = MapStream.of(map).mapBoth((k, v) -> entry(k + " - 1", (short) (v - 1))).toMap();
```
wjnicholson, line 159: Why all the (short) casting?

wjnicholson (approved):

modules/collect/src/test/java/com/opengamma/strata/collect/MapStreamTest.java
jodastephen: To demonstrate that the type of the result can be changed

wjnicholson: What type is being changed? You're going from `Map<String, Short> -> Map<String, Short>`

jodastephen: You can write `short s = 1` to convert the literal `1`. But you cannot convert that way to a capital S `Short` - the `1` is interpreted as an `int`, hence the cast.

wjnicholson: Ahah, got you. Isn't that due to `Guavate.entry` doing the boxing for you?

PR #2381, opened 2021-12-09
Allow Setting and Loading of SettlementType for Physically Settled Options

jordanmorrison476:
This code allows us to tag Physically Settled Options with a settlement type of 'Physical' and then gives us the option to filter these positions from the Eurex Time to Expiry Adjustment Calc.


jordanmorrison476:
Hi @jodastephen, would you mind reviewing this PR please? We need to update the EtdVariant + CsvUtilsLoader class in Strata to allow us to tag Eurex Government Bond Options with a settlement type of Physical so we can filter them out from the Eurex TEA calculation. 

More info on the problem is here for context: https://opengamma.atlassian.net/browse/PROD-11341

jodastephen (approved):
Adjusted code to be `ofMonthly(EtdSettlementType)`. Fixed generated ETD code.

PR #2382, opened 2021-12-09
Extend SwapIndex to FixedOvernight swaps

marc-henrard:
Extend the SwapIndex to support Fixed v Overnight swap as underlyings.
The new swap rates to support are
* GBP and USD: ICE Swap Rate https://www.theice.com/iba/ice-swap-rate
* JPY: Refinitif Tokyo Swap Rate https://www.refinitiv.com/en/financial-data/financial-benchmarks/tokyo-swap-rate
Only the GBP and USD are added in the "OvernightSwapIndices" as I don't know if the JPY ones are heavily used.

NOTE: This change introduces source and binary incompatibility, but results in the best outcome for the long term.


modules/measure/src/main/java/com/opengamma/strata/measure/cms/CmsMeasureCalculations.java
```diff
@@ -260,7 +260,7 @@ MultiCurrencyAmount currentCash(
 
   //-------------------------------------------------------------------------
   // returns the Ibor index or the CMS leg
-  static IborIndex cmsLegIborIndex(ResolvedCmsTrade trade) {
+  static RateIndex cmsLegIborIndex(ResolvedCmsTrade trade) {
```
jodastephen, line 263: This method should be renamed,

jodastephen (approved):

modules/measure/src/main/java/com/opengamma/strata/measure/cms/CmsMeasureCalculations.java
marc-henrard: Done. Method used only locally, no impact outside the class.

jodastephen (approved):

PR #2385, opened 2021-12-15
Avoid NullPointer exception when forward is close to zero in CmsIntegrantProvider

qed-:
g() references the g0 vector that's defined below if forward is below EPS


jodastephen (approved):

jodastephen:
thanks

PR #2426, opened 2022-02-11
Handle combined holiday calendars correctly in HolidaySafeReferenceData

skitini:
- Fix JavaDocs for `HolidayCalendar.combinedWith()` and `HolidayCalendar.linkedWith()`
- If a combined/linked holiday calendar ID is not present in the reference data used for `HolidaySafeReferenceData`, don't default to creating an `ImmutableHolidayCalender` when `queryValueOrNull()` and `tryDefaultValue()` are called. If null is returned, then `HolidayCalendarId.create()` will handle splitting the IDs appropriately and try to get the calendar using the individual calendar IDs


modules/basics/src/main/java/com/opengamma/strata/basics/date/HolidayCalendar.java
```diff
@@ -280,7 +280,7 @@ public default int daysBetween(LocalDate startInclusive, LocalDate endExclusive)
    * Combines this holiday calendar with another.
    * <p>
    * The resulting calendar will declare a day as a business day if it is a
-   * business day in both source calendars.
+   * business day in either source calendar.
```
jodastephen, line 283: I don't think this correction is correct. combinedWith is where a business day only happens if both calendars agree that it is a business day.

modules/basics/src/test/java/com/opengamma/strata/basics/date/HolidaySafeReferenceDataTest.java
```diff
@@ -0,0 +1,117 @@
...
+  void test_nonCalendarId() {
+    HolidaySafeReferenceData test = new HolidaySafeReferenceData(ReferenceData.empty());
+
+    assertThat(test.queryValueOrNull(ID1)).isNull();
+  }
+
+  //-------------------------------------------------------------------------
+  @Test
+  public void coverage() {
+    Map<ReferenceDataId<?>, Object> dataMap = ImmutableMap.of(ID1, VAL1);
```
jodastephen, line 102: Is this actually testing `HolidaySafeReferenceData`?

modules/basics/src/test/java/com/opengamma/strata/basics/date/HolidaySafeReferenceDataTest.java
```diff
@@ -0,0 +1,117 @@
...
+  private static final TestingReferenceDataId ID1 = new TestingReferenceDataId("1");
+  private static final TestingReferenceDataId ID2 = new TestingReferenceDataId("2");
+  private static final Number VAL1 = 1;
+  private static final Number VAL2 = 2;
+
+  private static ImmutableHolidayCalendar createCalendar(HolidayCalendarId id) {
+    return ImmutableHolidayCalendar.of(id, ImmutableList.of(), WEEKEND_DAYS);
+  }
+
+  @Test
```
jodastephen, line 46: I would like to see tests using `getValue` or `findValue` in the super-interface as that is the main API that needs to be satisfied. eg. searching for GBLO+EUTA where GBLO is in reference data but EUTA is not should default EUTA and then combine it.

jodastephen (changes requested):

modules/basics/src/main/java/com/opengamma/strata/basics/date/HolidayCalendar.java
skitini: The `CombinedHolidayCalendar.isHoliday()` method is `calendar1.isHoliday(date) || calendar2.isHoliday(date)` and the `LinkedHolidayCalendar.isHoliday()` method is `calendar1.isHoliday(date) && calendar2.isHoliday(date)` so I don't think this is correct? The changes here also line up with the existing JavaDoc for both CombinedHolidayCalendar and LinkedHolidayCalendar.

jodastephen: The code refers to holidays, the Javadoc to business days.

skitini: Ah okay, I did misread that, shall revert.

jodastephen (approved):

PR #2429, opened 2022-02-22
Be able to add a RateIndex to the ImmutableRatesProviderBuilder.

santorog:


jodastephen (commented):
Doesn't `indexCurve()` do this?

santorog:
> Doesn't `indexCurve()` do this?
Well - you're right 😆

PR #2434, opened 2022-03-21
SABR parameters sensitiviites

marc-henrard:
Add a method to convert swaption sensitivity to volatility (at a given strike) converted to SABR sensitivities to the different model parameters (alpha, beta, rho, nu).


ChristianOG (approved):
Looks good.

PR #2438, opened 2022-04-11
Trade model for OTC bond options

marc-henrard:
Create descriptions for bond options: product and its resolved version.


modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBondOption.java
```diff
@@ -0,0 +1,800 @@
...
+  /**
+   * The <i>clean</i> price at which the option can be exercised, in decimal form.
+   * <p>
+   * The "clean" price excludes any accrued interest.
+   * <p>
+   * Strata uses <i>decimal prices</i> for bonds in the trade model, pricers and market data.
+   * For example, a price of 99.32% is represented in Strata by 0.9932.
+   */
+  @PropertyDefinition(validate = "ArgChecker.notNegative")
+  private final double strike;
```
benhamilton-brown, line 101: worth naming this cleanStrikePrice? It would be clearer when dealing with the object higher up in the stack.

modules/product/src/main/java/com/opengamma/strata/product/bond/ResolvedFixedCouponBondOption.java
```diff
@@ -0,0 +1,550 @@
...
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final ZonedDateTime expiry;
+  /**
+   * The quantity that was traded.
+   * <p>
+   * This will be positive if buying (call) and negative if selling (put).
+   */
+  @PropertyDefinition
+  private final double quantity;
```
benhamilton-brown, line 67: do we need quantity on both the underlying and the resolved version. we can add a getter on resolved to access the underlying quantity

marc-henrard: There is a "notional" on the FixedCouponBond, but this is intended for "tradable size". It most cases it is simply 1, but for some bond, they can be traded only in multiples of 1,000 or 1,000,000. The FixedCouponBondTrade have also a similar "quantity".

modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBondOption.java
marc-henrard: name changed

benhamilton-brown (approved):

PR #2448, opened 2022-05-04
Enhance failure management

jodastephen:
* New exception types `FailureItemException` and `ParseFailureException`
* Alter Result to capture failure exception types
* Alter ValueWithFailures to capture failure exception types
* Provide ability to recreate the message template

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
@@ -152,8 +152,8 @@ public final class Messages {
    * not throw another exception, which might hide the intended exception.
    * <p>
    * If the template contains a named placeholder, then the output will contain a populated attribute map
-   * for all placeholders including those without names.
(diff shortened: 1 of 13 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -143,12 +150,19 @@ public static String format(String messageTemplate, Object... args) {
    * <p>
    * This method is null tolerant to ensure that use in exception construction will
    * not throw another exception, which might hide the intended exception.
+   * <p>
+   * If the template contains a named placeholder, then the output will contain a populated attribute map
+   * for all placeholders including those without names.
+   * including a 'templateLocation' attribute identifying the location of the named placeholders.
```
ivayloVarbanov, line 156: is 156 supposed to be a separate sentence? Starts with a lower case letter

ivayloVarbanov (changes requested):

ivayloVarbanov (approved):

PR #2449, opened 2022-05-05
Enhance failure exceptions

jodastephen:
* Ensure that `FailureItem` captures underlying failures
* This avoids losing failure information within wrapped exceptions

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
@@ -277,7 +277,7 @@ public final class Messages {
   /**
    * Merges two template locations.
    * <p>
-   * This takes two template locations and the legth of the first message and returns the combined template.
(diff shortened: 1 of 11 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/Messages.java
```diff
@@ -272,4 +274,60 @@ public static String recreateTemplate(String message, String templateLocation) {
     }
   }
 
+  /**
+   * Merges two template locations.
+   * <p>
+   * This takes two template locations and the legth of the first message and returns the combined template.
```
anand-uk, line 280:
```suggestion
   * This takes two template locations and the length of the first message and returns the combined template.
```

anand-uk (changes requested):

anand-uk (approved):
As discussed offline, the failure attributes might get overridden when combining two or more failures items. This can be addressed in the subsequent PRs if required.

anand-uk (approved):

PR #2450, opened 2022-05-06
Cms spread expiry time

benhamilton-brown:


marc-henrard:
Solved on the CMS spread side.

PR #2453, opened 2022-05-09
More lenient payment schedule

jodastephen:
* Consider a single day accrual period with unadjusted start date
* When converted to a payment schedule this fails
* But really only the payment date needs adjusting
* Alter the code to be more lenient with the schedule start ad end date


jodastephen:
@Mergifyio refresh

PR #2456, opened 2022-05-10
Enhance parse exceptions

jodastephen:
* Use ParseFailureException
* Capture more detail on failures


anand-uk (changes requested):
`modules/loader/src/main/java/com/opengamma/strata/loader/fpml/FpmlDocument.java` 
In the above class, the docs for the public methods needs updating ... either throws `FpmlParseException` or `ParseFailureException`

PR #2471, opened 2022-05-19
Add support for THB THOR index, also move THB calendar to strata

benhamilton-brown:


benhamilton-brown:
Hey Stephen. The THB calendar is already in granite, so plan is to copy over calendar details from granite ini file to Strata. However the static variable AdditionalHolidayCalendarIds.THBA is still in granite, plan was to leave it in granite as would need changing variable downstream in a lot of places to move over to HolidayCalendarIds.

skitini:
If adding the index, should add the FloatingRateName also -> add to `FloatingRateNames` and `FloatingRateNameData.ini`. Should also add a test for the index in `OvernightIndexTest`.

benhamilton-brown:
> Hey Stephen. The THB calendar is already in granite, so plan is to copy over calendar details from granite ini file to Strata. However the static variable AdditionalHolidayCalendarIds.THBA is still in granite, plan was to leave it in granite as would need changing variable downstream in a lot of places to move over to HolidayCalendarIds.

modules/basics/src/main/resources/META-INF/com/opengamma/strata/config/base/FloatingRateNameData.ini
```diff
@@ -214,6 +214,7 @@ SGD-SOR-VWAP = SGD-SOR
 SGD-SOR-VWAP-Reference Banks = SGD-SOR
 
 # "THB-THBFIX" is the ISDA 2021 code
+THB-THOR = THB-THOR
```
skitini, line 217: This should be in its own "section", it's not part of the THBFIX rate.

modules/basics/src/main/resources/META-INF/com/opengamma/strata/config/base/FloatingRateNameData.ini
```diff
@@ -213,6 +213,9 @@ SGD-SOR-Telerate = SGD-SOR
 SGD-SOR-VWAP = SGD-SOR
 SGD-SOR-VWAP-Reference Banks = SGD-SOR
 
+# "THB-THOR/THB-THOR-OIS COMPOUND" is the ISDA 2021 code
```
skitini, line 216: Are these both the standard codes, or is it just the one?

jodastephen (approved):

modules/basics/src/main/resources/META-INF/com/opengamma/strata/config/base/FloatingRateNameData.ini
benhamilton-brown: both are

PR #2488, opened 2022-07-21
fix discount factors for historical dates

benhamilton-brown:


yukiiwashita:
This change to DiscountFactors involves a potential risk - we never get a value estimated from a curve for negative year fractions. If you really need discountFactor=1 for negative year fractions, is this something we can control in the underlying curve, e.g., 0 for negative x values for a zero rate curve, etc.

modules/pricer/src/main/java/com/opengamma/strata/pricer/ZeroRateDiscountFactors.java
```diff
@@ -168,11 +170,17 @@ public double relativeYearFraction(LocalDate date) {
   @Override
   public double discountFactor(double yearFraction) {
     // convert zero rate to discount factor
+    if (yearFraction <= EFFECTIVE_ZERO) {
```
yukiiwashita, line 173: If `discountFactor` is modified, `zeroRate` needs to be also modified for consistency. By the current change, we have discountFactor = 1, but nonzero yearFraction*zeroRate, for negative year fractions.

modules/pricer/src/main/java/com/opengamma/strata/pricer/ZeroRatePeriodicDiscountFactors.java
```diff
@@ -194,14 +200,18 @@ public double discountFactorTimeDerivative(double yearFraction) {
 
   @Override
   public double zeroRate(double yearFraction) {
-    double ratePeriod = curve.yValue(yearFraction);
+    double yearFractionMod = Math.max(EFFECTIVE_ZERO, yearFraction);
+    double ratePeriod = curve.yValue(yearFractionMod);
```
yukiiwashita, line 204: Is this consitent with `discountFactor` and `zeroRatePointSensitivity`?

yukiiwashita (changes requested):

modules/pricer/src/main/java/com/opengamma/strata/pricer/ZeroRatePeriodicDiscountFactors.java
benhamilton-brown: I have replaced the above and done the same in ZeroRateDiscountFactors.java to ensure we always return 0 if year faction is not greater than 0.

yukiiwashita (approved):

PR #2498, opened 2022-08-18
Undoing revert of #2381 as it did not solve problem as expected

MichaelRol:


jordanmorrison476 (commented):
Hey @MichaelRol , I think this can still be reverted as we never actually ended up using this logic in the end. The root cause issue was due to dodgy Eurex Market data, so we put these changes in place to override their screw up. Eurex corrected their market data so we won't need these changes.

PR #2512, opened 2022-09-12
Normal SABR swaption volatilities

marc-henrard:
Add the SABR swaption with normal SABR formula.
Similar to the existing normal SABR formula for caplet/floorlet.


modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/NormalSabrParametersSwaptionVolatilities.java
```diff
@@ -0,0 +1,1004 @@
...
+  private final ZonedDateTime valuationDateTime;
+  /**
+   * The SABR model parameters.
+   * <p>
+   * Each model parameter of SABR model is a surface.
+   * The x-value of the surface is the expiry, as a year fraction.
+   * The y-value of the surface is the swap tenor, as a year fraction rounded to the month.
+   */
+  @PropertyDefinition(validate = "notNull")
+  private final SabrInterestRateParameters parameters;
```
yukiiwashita, line 84: `ValueType` for `NormalSabrSwaptionVolatilities#getVolatilityType` does not necessarily match that for `SabrVolatilityFormula#getVolatilityType` in `SabrInterestRateParameters`. Do we need to add ArgChecker or is this intended?

yukiiwashita (changes requested):

modules/pricer/src/main/java/com/opengamma/strata/pricer/swaption/NormalSabrParametersSwaptionVolatilities.java
marc-henrard: The check is indeed missing. Adding it and updating the tests.

yukiiwashita (approved):

PR #2522, opened 2022-10-03
Make series and version optional when parsing CDS Index Trades

skitini:
- Removing the need to have series and version present when parsing a CDS index ID, replacing with a plain STD (for standard) stem to the ID. If the series and version are still present, we keep existing logic.
- Added a check that the  present RED code is a RED9 code - should not be parsing RED6 type codes for CDS indices.


santorog (approved):
Looks good to me

jodastephen (approved):

PR #2524, opened 2022-10-04
Expose forward rate in Ibor / Overnight future pricers

santorog:


modules/pricer/src/main/java/com/opengamma/strata/pricer/index/DiscountingOvernightFutureTradePricer.java
```diff
@@ -201,4 +201,15 @@ public PointSensitivities parSpreadSensitivity(ResolvedOvernightFutureTrade trad
     return productPricer.priceSensitivity(trade.getProduct(), ratesProvider);
   }
 
+  /**
+   * Returns the forwardRate rate.
+   *
+   * @param trade  the overnight future trade
```
robertszostakowskiOG, line 207: a resolved overnight future trade? Maybe it is worth mentioning that this must be the resolved version

robertszostakowskiOG (approved):

modules/pricer/src/main/java/com/opengamma/strata/pricer/index/DiscountingOvernightFutureTradePricer.java
jodastephen: Strata doesn't normally refer to the resolved form in Javadoc explicitly

jodastephen (approved):

PR #2525, opened 2022-10-07
adding commonly used convenience methods to Decimal and BigMoney

anand-uk:


modules/basics/src/main/java/com/opengamma/strata/basics/currency/BigMoney.java
```diff
@@ -278,6 +278,70 @@ public BigMoney mapAmount(UnaryOperator<BigDecimal> mapper) {
...
 
   //-------------------------------------------------------------------------
+  /**
+   * Checks if this amount is greater than the other amount.
+   *
+   * @param otherAmount the other amount
+   * @return true if this amount is greater than the other amount
+   * @throws IllegalArgumentException if the currencies are not equal
+   */
+  public boolean isGreaterThan(BigMoney otherAmount) {
```
jodastephen, line 288: These new methods need tests

modules/collect/src/main/java/com/opengamma/strata/collect/Decimal.java
```diff
@@ -906,6 +906,46 @@ public String format(int decimalPlaces, RoundingMode roundingMode) {
...
+  }
+
+  /**
+   * States if this decimal is less than the other decimal.
+   *
+   * @param other the other decimal
+   * @return true if this is less than the other
+   */
+  public boolean isLessThan(Decimal other) {
+    return !isGreaterThanEqualTo(other);
```
jodastephen, line 936: It would be more consistent to always use `compareTo`

jodastephen (changes requested):

jodastephen (approved):

PR #2532, opened 2022-11-28
Fix FxSwapConvention for sell

jodastephen:
Add FX methods that take Buy/Sell currency to clarify direction Retain Buy/Sell in an attribute
Update CSV parser for FX single/swap


modules/product/src/main/java/com/opengamma/strata/product/fx/type/ImmutableFxSwapConvention.java
```diff
@@ -157,19 +160,23 @@ public FxSwapTrade toTrade(
...
 
+    if (!currencyPair.contains(buySellCurrency)) {
+      throw new IllegalArgFailureException("Buy/Sell currency {value} does not match convention {options}", buySellCurrency, currencyPair);
+    }
     Optional<LocalDate> tradeDate = tradeInfo.getTradeDate();
     if (tradeDate.isPresent()) {
       ArgChecker.inOrderOrEqual(tradeDate.get(), startDate, "tradeDate", "startDate");
     }
-    double amount1 = BuySell.BUY.normalize(notional);
+    double amount1 = buySell.normalize(notional);
```
jodastephen, line 175: This is the key fix. It means the result takes account of the Buy/Sell flag where previously the flag was ignored.

yukiiwashita (approved):

PR #2544, opened 2023-01-03
Add Frequency to InterpolatedNodalCurveDefinition

yukiiwashita:
Added Frequency to InterpolatedNodalCurveDefinition as an optional field so that the curve calibrator can handle periodically compouneded zero rate curves.


marc-henrard (approved):
Good for me.

Issue #2296, opened 2021-02-25
test errors on AArch64, Fedora 33 


LutzWeischerFujitsu:
... 
[INFO] Running com.opengamma.strata.pricer.DiscountingPaymentPricerTest
[INFO] Tests run: 25, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.016 s - in com.opengamma.strata.pricer.DiscountingPaymentPricerTest
[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   BlackIborCapletFloorletPeriodPricerTest.regression_pv:658
Expecting:
  <3.4403901240887222>
to be close to:
  <3.4403901240887094>
by less than <1.0E-14> but difference was <1.2878587085651816E-14>.
(a difference of exactly <1.0E-14> being considered valid)
[ERROR]   SabrExtrapolationReplicationCmsLegPricerTest.test_explainPresentValue:313
Expecting:
 <39728.51321029527>
to be equal to:
 <39728.51321029542>
but was not.
[ERROR]   BlackBarrierPriceFormulaRepositoryTest.adjointPriceRegression:301->assertRelative:308
Expecting:
  <1.5563229354864467>
to be close to:
  <1.55632293548648>
by less than <1.55632293548648E-14> but difference was <3.3306690738754696E-14>.
(a difference of exactly <1.55632293548648E-14> being considered valid)
[ERROR]   ImpliedTrinomialTreeLocalVolatilityCalculatorTest.flatVolTest:80
Expecting:
 <1L>
to be equal to:
 <0L>
but was not.
[ERROR]   SabrModelFitterTest.testNoisyFitWithFixedBeta:179
Expecting value to be true but was false
[ERROR]   DiscountingSwapProductPricerTest.three_leg_swap:1191
Expecting:
 <[EUR 1.862645149230957E-9]>
to be equal to:
 <[EUR 0]>
but was not.
[INFO]
[ERROR] Tests run: 3030, Failures: 6, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Strata-Root 2.9.2-SNAPSHOT:
[INFO]
[INFO] Strata-Parent ...................................... SUCCESS [  7.469 s]
[INFO] Strata-Collect ..................................... SUCCESS [01:25 min]
[INFO] Strata-Basics ...................................... SUCCESS [01:29 min]
[...]

PR #2556, opened 2023-02-21
Fix tests for ARM

mbsDonatasZ:
Fixes https://github.com/OpenGamma/Strata/issues/2296

Due to ARM and X86 floating point calculation differences it was required to increase tolerance on few tests to successfully build on AARCH64.


jodastephen (approved):

jodastephen:
Thanks

PR #2559, opened 2023-03-02
Making CombinedReferenceData public

MichaelRol:


MichaelRol:
I need CombinedReferenceData to be public so that QueriableReferenceData in client data will handle CombinedReferenceData with an underlying LazyReferenceData. 

https://github.com/OpenGamma/ClientData/pull/9031

MichaelRol:
Another failed attempt at getting my first Strata commit 😂

PR #2581, opened 2023-05-29
Yield and Price Derivatives for Inflation Bonds

yukiiwashita:
* Added derivative calculations for price-yield conversions for inflation bond
* Added FieldName
* Added isRealPrice method to the yield conventions


marc-henrard (changes requested):
Only cosmetic suggestions.

marc-henrard (approved):
Good for me.

PR #2591, opened 2023-07-05
Deactive Libor indices after Libor transition

johannesOG:


modules/basics/src/test/java/com/opengamma/strata/basics/index/FloatingRateIndexTest.java
```diff
@@ -46,9 +39,6 @@ public void test_parse_withTenor() {
 
   @Test
   public void test_tryParse_noTenor() {
-    assertThat(FloatingRateIndex.tryParse("GBP-LIBOR")).isEqualTo(Optional.of(IborIndices.GBP_LIBOR_3M));
```
brianweller89, line 49: We can't really do this as it will break calculations for old valuation dates where LIBOR swaps may still be present.

Does setting the index to `inactive` mean that it doesn't work in the `tryParse` method anymore?

johannesOG: I was deleting too much here, it is the 
`    assertThat(FloatingRateIndex.parse("GBP-LIBOR", Tenor.TENOR_6M)).isEqualTo(IborIndices.GBP_LIBOR_6M);
`that actually fails
These two statements still work
```
    assertThat(FloatingRateIndex.parse("GBP-LIBOR-1M", Tenor.TENOR_6M)).isEqualTo(IborIndices.GBP_LIBOR_1M);
    assertThat(FloatingRateIndex.parse("GBP-LIBOR-3M", Tenor.TENOR_6M)).isEqualTo(IborIndices.GBP_LIBOR_3M);
```

brianweller89: @jodastephen What do you think on this? Initially our intention here was to deactivate the discontinued LIBOR curves so they no longer showed up in the index dropdowns in our UI.

But that seems to be causing Johannes quite a bit of pain with various tests breaking

jodastephen: It would be good to deactivate them as this PR does, but we would need some way to continue having `assertThat(FloatingRateIndex.parse("GBP-LIBOR")).isEqualTo(IborIndices.GBP_LIBOR_3M);` pass.

Maybe it just needs an alias adding in `FloatingRateName.ini`?

```
[alternates]
GBP-LIBOR = GBP-LIBOR-3M
```

PR #2599, opened 2023-08-28
Add ShiftType.FIXED

Mistuus:
Adding a new shift type that replaces the base value with the shift amount


jodastephen (approved):
A bit weird, but OK

brianweller89:
I think part of the reason it looks weird is that the enum is called `ShiftType`, whereas `ShockType` would be more appropriate. Typically this enum is used for defining the scenarios which get executed as part of a VaR calculation. And defining scenarios in terms of fixed prices is a pretty reasonable approach

Issue #2597, opened 2023-08-11
Periodic schedule deduplication does not work when there are same elements in more than 2 count


lfathimakn:
When creating a schedule from a periodic schedule, there is a deduplication logic if we pass "combinePeriodsIfNecessary" flag as true:
https://github.com/OpenGamma/Strata/blob/main/modules/basics/src/main/java/com/opengamma/strata/basics/schedule/PeriodicSchedule.java#L453-L462

But if there are more than 2 duplicates of same element, the deduplication is only removing first instance of duplicate. 
Eg:
adj = ["2020-01-01","2020-01-02","2020-01-03","2020-01-06","2020-01-06","2020-01-06","2020-01-07"]
After deuplcation:
adj = ["2020-01-01","2020-01-02","2020-01-03","2020-01-06","2020-01-06","2020-01-07"]

PR #2603, opened 2023-09-17
Fix for duplicate dates removal if combinePeriodsIfNecessary is set to true

lfathimakn:
Fix for duplicate dates removal if combinePeriodsIfNecessary is set to true
Issue was reported before in below bug:
https://github.com/OpenGamma/Strata/issues/2597

### DIFF
--- a/modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBond.java
+++ b/modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBond.java
@@ -170,19 +170,7 @@ public final class FixedCouponBond
    */
   @Override
   public ResolvedFixedCouponBond resolve(ReferenceData refData) {
-    return resolve(refData, false);
-  }
-
-  /**
-   * Resolves fixed coupon bond using specified reference data.
-   * @param refData  the reference data to use when resolving
-   * @param combinePeriodsIfNecessary flag to indicate that duplicate periods can be combined
-   * @return the resolved instance
-   * @throws ReferenceDataNotFoundException if an identifier cannot be resolved in the reference data
-   * @throws RuntimeException if unable to resolve due to an invalid definition
-   */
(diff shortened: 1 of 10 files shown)


modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBond.java
```diff
@@ -161,10 +161,28 @@ private void validate() {
...
+   * Resolves fixed coupon bond using specified reference data.
+   * @param refData  the reference data to use when resolving
+   * @return the resolved instance
+   * @throws ReferenceDataNotFoundException if an identifier cannot be resolved in the reference data
+   * @throws RuntimeException if unable to resolve due to an invalid definition
+   */
   @Override
   public ResolvedFixedCouponBond resolve(ReferenceData refData) {
-    Schedule adjustedSchedule = accrualSchedule.createSchedule(refData);
+    return resolve(refData, false);
```
jodastephen, line 173: My feeling is that we can just change this to use true instead of false. This avoids adding the extra method. It seems that combining periods is the normal thing most users would expect.

jodastephen (commented):
Thanks for the PR. I think we can simplify the bond part.

modules/product/src/main/java/com/opengamma/strata/product/bond/FixedCouponBond.java
lfathimakn: sure, let me update the MR

lfathimakn: @jodastephen can you please check the latest commit. I have removed the method and put true as default

jodastephen (approved):
Thanks

lfathimakn:
Thank you @jodastephen . When will the next version be released which has the fix?

PR #2615, opened 2023-11-23
Return all countries from getAvailableCountries

MichaelRol:


brianweller89:
@jodastephen - any concerns with this? We want a more complete list of contracts to expose as a dropdown in the UI

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
```diff
@@ -318,7 +321,6 @@ public static Country parse(String countryCode) {
   public static Country of3Char(String countryCode) {
     ArgChecker.matches(CODE_MATCHER, 3, 3, countryCode, "countryCode", "[A-Z][A-Z]");
```
jodastephen, line 322:
```suggestion
    ArgChecker.matches(CODE_MATCHER, 3, 3, countryCode, "countryCode", "[A-Z][A-Z][A-Z]");
```

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
```diff
@@ -41,10 +42,12 @@ public final class Country
...
    */
-  private static final Supplier<ImmutableBiMap<String, String>> COUNTRY_CODES = Suppliers.memoize(() ->
-      ImmutableBiMap.copyOf(PropertiesFile.of(
-          ResourceLocator.ofClasspath(Country.class, "country.properties").getCharSource())
-          .getProperties().asMap()));
+  private static final Supplier<ImmutableBiMap<String, Country>> COUNTRY_CODES = Suppliers.memoize(() ->
+      ImmutableBiMap.copyOf(MapStream.of(PropertiesFile.of(
+                  ResourceLocator.ofClasspath(Country.class, "country.properties").getCharSource())
+              .getProperties().asMap())
+          .mapValues(Country::of)
```
jodastephen, line 49: I believe this will cause all instances in the config file to be added to CACHE, thus I don't think you need to use `concatToSet` with the CACHE in `getAvailableCountries`

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
```diff
@@ -256,11 +259,11 @@ public final class Country
    * <p>
    * This contains all the countries that have been defined at the point
    * that the method is called.
-   * 
+   *
    * @return an immutable set containing all registered countries
    */
   public static Set<Country> getAvailableCountries() {
-    return ImmutableSet.copyOf(CACHE.values());
+    return Guavate.concatToSet(COUNTRY_CODES.get().values(), CACHE.values());
```
jodastephen, line 266: This should really also be sorted, so use `ConcurrentSkipListMap` as the CACHE implementation.

Would be good to have the test ensure that more instances are returned by this method

jodastephen (changes requested):

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
MichaelRol: This is correct, but EU is not in the config file, and by continuing to concat with the CACHE it means that any new countries that a user adds with `Countries::of` will also become available in `getAvailableCountries`. 

I can't just return the CACHE either, as that would rely on the Supplier having been called at least once already.

jodastephen: Write a test, but it looks to me that all instances are constructed via `of(String)`, where `computeIfAbsent()` will always call the supplier. Thus the CACHE will always contain all instances.

MichaelRol: That's correct, but the ones coming from the config file don't run until the `Suppliers.memoize()` thing is called for the first time. So to ensure that we get all countries, we need to make sure that `COUNTRY_CODES.get()` has been called. 
```
  @Test
  public void test_whether_Cache_is_always_fully_populated() {
    Set<Country> available = Country.getAvailableCountries(); //.size() == 46
    COUNTRY_CODES.get();
    Set<Country> more = Country.getAvailableCountries(); //.size() == 252
    assertThat(available.size()).isEqualTo(more.size()); // fails
  }
```

jodastephen: OK then, so the `concatToSet` is fine, but the result does need sorting

MichaelRol: `concatToSet` is not really needed now I think about it, as long as we call `COUNTRY_CODES.get()` the CACHE will contain the complete list. I'm then taking the values from the CACHE, and creating a SortedSet as the result.

modules/basics/src/test/java/com/opengamma/strata/basics/location/CountryTest.java
```diff
@@ -87,6 +87,14 @@ public void test_getAvailable() {
     assertThat(available.contains(Country.CA)).isTrue();
   }
 
+  @Test
+  public void test_new_Country_included_in_getAvailable() {
+    Set<Country> available = Country.getAvailableCountries();
```
jodastephen, line 92: assert available.size() > 0

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
```diff
@@ -256,11 +259,12 @@ public final class Country
...
    * This contains all the countries that have been defined at the point
    * that the method is called.
-   * 
+   *
    * @return an immutable set containing all registered countries
    */
   public static Set<Country> getAvailableCountries() {
-    return ImmutableSet.copyOf(CACHE.values());
+    COUNTRY_CODES.get();
+    return ImmutableSortedSet.copyOf(CACHE.values());
```
jodastephen, line 267: I'm pretty sure this is enough because of the ConcurrentSkipListMap. Worth checking though

```suggestion
    return CACHE.values();
```

jodastephen (changes requested):

modules/basics/src/main/java/com/opengamma/strata/basics/location/Country.java
MichaelRol: `CACHE.values()` returns a`Collection` so I used the `ImmutableSortedList` to avoid updating the return type.

jodastephen (approved):

PR #2622, opened 2023-12-18
Use defaults instead of null in IborRateSwapLegConvention

milenOG:
Invoking the getters in the toLeg() method will use defaults if the property is null.
E.g. instead of stub rate being null it will be defaulted to smart initial


jodastephen (approved):
It is quite late in the life of Strata to change this, but hopefully not harmful

milenOG:
I actually changed the approach. The defaulting is now only for the stub convention.
This standardises the behaviour between the parser and toLeg methods.
I.e in FullSwapTradeCsvPlugin.parseAccrualSchedule only the stub convention is defaulted, the adjustment dates and roll conventions are not.

vaibhavkrishnakumar (approved):

PR #2639, opened 2024-03-18
Remove redundant check in CurrencyAmount.negated()

brianweller89:
This is already handled in the constructor


modules/basics/src/main/java/com/opengamma/strata/basics/currency/CurrencyAmount.java
```diff
@@ -315,7 +315,7 @@ public boolean isNegative() {
    */
   public CurrencyAmount negated() {
     // Zero is treated as a special case to avoid creating -0.0 which produces surprising equality behaviour
-    return new CurrencyAmount(currency, amount == 0d ? 0d : -amount);
+    return new CurrencyAmount(currency, -amount);
```
jodastephen, line 318: Need to remove the comment as wlell

jodastephen (changes requested):

jodastephen (approved):

PR #2643, opened 2024-04-24
Added converted to result method on FxConvertible

James-OpenGamma:


modules/basics/src/test/java/com/opengamma/strata/basics/currency/CurrencyAmountTest.java
```diff
@@ -239,6 +245,19 @@ public void test_convertedTo_rateProvider() {
     assertThat(CCY_AMOUNT.convertedTo(CCY1, provider)).isEqualTo(CCY_AMOUNT);
   }
 
+  @Test
+  public void test_convertedToResult_failure() {
```
skitini, line 249: Can you add a test for the success case

modules/basics/src/main/java/com/opengamma/strata/basics/currency/FxConvertible.java
```diff
@@ -35,4 +37,20 @@ public interface FxConvertible<R> {
...
+   * Any FX conversion that is required will use rates from the provider.
+   * <p>
+   * If no FX rate is found the returned result will be a failure.
+   *
+   * @param resultCurrency  the currency of the result
+   * @param rateProvider  the provider of FX rates
+   * @return the result of the converted instance, which should be expressed in the specified currency
+   */
+  public default Result<R> convertedToResult(Currency resultCurrency, FxRateProvider rateProvider) {
+    return Result.of(() -> convertedTo(resultCurrency, rateProvider));
```
jodastephen, line 53: Given this implementation, I'm struggling to justify adding the method,. Bear in mind that the method becomes public on a large number of implementation classes.

James-OpenGamma: We have `Result.of(() -> ...convertedTo)` littered around most of the models, as we want to avoid any hard errors in a calculation, this method would help clean things up a lot.

PR #2645, opened 2024-04-24
Added Result of optional method

James-OpenGamma:
## Summary by CodeRabbit

- **New Features**
  - Introduced methods to handle optional values with custom failure messages in the `Result` class.

- **Tests**
  - Added tests to verify the behaviour of `Result` with optional values, including success and failure cases.

### DIFF
--- a/modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
+++ b/modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
@@ -460,6 +460,49 @@ public final class Result<T>
     return ofNullable(value, FailureReason.MISSING_DATA, "Found null where a value was expected");
   }
 
+  /**
+   * Returns a success result containing the value is present, else returns a failure result
+   * with the specified reason and message.
+   * <p>
+   * The message is produced using a template that contains zero to many "{}" placeholders.
+   * Each placeholder is replaced by the next available argument.
+   * If there are too few arguments, then the message will be left with placeholders.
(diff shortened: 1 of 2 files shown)


modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
```diff
@@ -460,6 +460,49 @@ public static <R> Result<R> ofNullable(R value) {
     return ofNullable(value, FailureReason.MISSING_DATA, "Found null where a value was expected");
   }
 
+  /**
+   * Returns a success result containing the value is present, else returns a failure result
```
jodastephen, line 464:
```suggestion
   * Returns a success result containing the value if present, else returns a failure result
```

modules/collect/src/main/java/com/opengamma/strata/collect/result/Result.java
```diff
@@ -460,6 +460,49 @@ public static <R> Result<R> ofNullable(R value) {
...
+   * with a reason of {@link FailureReason#MISSING_DATA} and message to say an unexpected empty value was found.
+   *
+   * @param <R> the expected type of the result
+   * @param value  the potentially null value
+   * @return a success result if the value is non-null, else a failure result
+   */
+  public static <R> Result<R> ofOptional(
+      Optional<R> value) {
+
```
jodastephen, line 502:
```suggestion
  public static <R> Result<R> ofOptional(Optional<R> value) {
```

jodastephen:
@coderabbitai review

Issue #2592, opened 2023-07-19
Yield calculates incorrectly on bonds with short/long last coupon


ShaunParkYieldX:
Here is an example bond.
Cusip: 52107QAG0
ISIN: US52107QAG01
Dated Date: 02/13/2015
First Coupon Date: 09/01/2015
Maturity Date: 02/13/2015
Coupon: 3.75
Coupon Frequency: Semi-Annual
Day Count Convention: 30/360

This bond has long first coupon (02/13/2015 to 09/01/2015) and short last coupon (09/01/2024 to 02/13/2025)
In DiscountingFixedCouponBondProductPricer, inside method dirtyPriceFromYieldStandard,  the last coupon and principal are discounted back to the first coupon date. It uses variable called pvAtFirstCoupon.
The problem appears to be with Math.pow(factorOnPeriod, (double)pow), where pow is assumed to be an integer. When a bond has short/long last coupon, that last period will not be 1. For a short last coupon, the last period is less than one. For a long last coupon, the last period is greater than one.

This incorrect discounting of the last coupon and principal results in the yield calculation being incorrect for bonds with short/long last coupon.›

ritonglue:
Same problem with this Bond
ISIN : XS2352739184
Dated Date : 2021-06-30
First coupon date : 2021-10-15
Maturity Date : 2026-06-30
Coupon : 0,0425 EUR
Nominal : 1,00 EUR
Coupon Frequency Semi-Annual
Day Count Convention: 30/360

Buy example
Quantity : 1 000 000
SettlementDate : 2023-06-06
Clean Price : 100.09 %
Results from DiscountingFixedCouponBondProductPricer.DEFAULT are :
Accrued : EUR 12041.67 (correct)

The yieldFromDirtyPrice method computes : 7.8034173382765015 %
This is far to below from 8.5 % with a such clean price.

I changed the last period from 1 to 2*0.208333 and the yield is now 8.46 % which is closer to 8.5 %

ritonglue:
Unfortunately, my PR breaks existing result. For example if one uses ACT_ACT_ISDA.

ritonglue:
Discussion about this problem in the forum : https://forums.opengamma.com/t/yield-calculates-incorrectly-on-bonds-with-short-long-last-coupon/1248

ritonglue:
The new PR doesn't break results if one uses ACT_ACT_ISDA.

jodastephen:
PR #2694

[...]

PR #2654, opened 2024-06-21
Pricer-2592 fix yield on bonds with short/long last coupon

ritonglue:
Fixes issue OpenGamma/Strata#2592


ritonglue:
Unfortunately, it will break results if ACT_ACT_ISDA is used

PR #2655, opened 2024-07-02
Pricer 2592 fix irregular dates yield

ritonglue:
fixes OpenGamma/Strata/#2592


ritonglue:
@jodastephen could you please have a look at this PR ?

PR #2659, opened 2024-08-05
Ensure toOnly collector supplier returns a new instance

James-OpenGamma:


jodastephen (approved):
A `Collector` Supplier must always return a new instance

PR #2682, opened 2024-10-24
Adding Inauguration Day to MXMC

Andras1022:


jodastephen (changes requested):
No. The original location (not module/basics) is the right location

MichaelRol:
@jodastephen Are you referring to the .bin file itself or the `Path` in the script?

jodastephen:
Both should be in the location they were before the recent change

MichaelRol:
Yes, so this PR corrects the location of the update .bin file - the change of the Path is because the old path caused the issue where the binary was incorrectly generated outside of the /modules/basics package. Both @Andras1022 and myself have tested with the updated Path and it correctly updates the binary file inside /modules/basics

jodastephen:
This appears to be an issue with IntelliJ, which uses a poor choice of Working Directory by default when running code. The tool is supposed to be run with a Working Directory of *strata-basics*, therefore there the extra `module/basics` is incorrect.

(Logically, once the source is compiled it has no knowledge of the original source code structure, so "module/basics" has no actual meaning at runtime. Eclipse's choice of working directory therefore makes much more sense here.)

Andras1022:
@jodastephen I have removed the extra "module/basics" from the path, but in the original PR the binary file was an addition, now it's replacing the previous version.

jodastephen (approved):

PR #2694, opened 2025-01-05
Bond irregular dates

jodastephen:
Updated code from #2655. Applies a fix to `dirtyPriceFromYieldStandard` for `FixedCouponBond`.

The main change is the use of `Math.pow(factorOnPeriod, yearFraction * eventsPerYear)` instead of `factorOnPeriod` when the period is not regular.


ritonglue:
The year 2025 starts very nicely. Thank you for reviewing my modifications.

PR #2727, opened 2025-06-26
Adding Delta Measure and Methods to pricers and calc functions

SaponeRolando:


modules/measure/src/main/java/com/opengamma/strata/measure/StandardMeasures.java
```diff
@@ -34,6 +34,8 @@ final class StandardMeasures {
   public static final Measure CASH_FLOWS = ImmutableMeasure.of("CashFlows");
   // currency exposure, with no currency conversion
   public static final Measure CURRENCY_EXPOSURE = ImmutableMeasure.of("CurrencyExposure", false);
+  // delta
+  public static final Measure DELTA = ImmutableMeasure.of("Delta");
```
yukiiwashita, line 38: Given that there are various measures people call delta, would OPTION_DELTA be a better name?

modules/pricer/src/main/java/com/opengamma/strata/pricer/fxopt/VannaVolgaFxVanillaOptionProductPricer.java
```diff
@@ -282,6 +282,48 @@ public MultiCurrencyAmount currencyExposure(
...
 
+  /**
+   * Calculates the delta of the foreign exchange vanilla option product.
+   *
+   * @param option  the option product
+   * @param ratesProvider  the rates provider
+   * @param volatilities  the Black volatility provider
+   * @return the delta
+   */
+  public double delta(
```
yukiiwashita, line 293: It would be good to add a unit test to cover this delta method as well.

SaponeRolando:
@yukiiwashita agreed with Alexis to not add these to the pricer which we do not actually use, i.e. Implied Trinomial for Single barrier option and Vanna Volga for Vanilla options

yukiiwashita (dismissed):

skitini (commented):
Two minor styling points

PR #2777, opened 2026-05-18
Auto compute missing BRL FVN

skitini:


modules/loader/src/main/java/com/opengamma/strata/loader/csv/SwapTradeCsvPlugin.java
```diff
@@ -121,12 +125,56 @@ static SwapTrade parse(CsvRow row, List<CsvRow> variableRows, TradeInfo info, Tr
...
 
+  // enhances BRL-CDI swaps by adding FutureValueNotional to the fixed leg
+  static SwapTrade enhanceBrlCdiSwap(SwapTrade trade) {
+    if (!hasBrlCdiFloatingLeg(trade)) {
+      return trade;
+    }
+
+    ImmutableList<SwapLeg> enhancedLegs = trade.getProduct().getLegs().stream()
+        .map(SwapTradeCsvPlugin::addFvnToFixedLegIfNeeded)
+        .collect(ImmutableList.toImmutableList());
```
James-OpenGamma: Guavate here, and List<>

skitini: Can't use List<> as `replaceLegs()` expect implementation, and would rather not go changing existing methods

James-OpenGamma (dismissed):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/SwapTradeCsvPlugin.java
```diff
@@ -121,12 +126,56 @@ static SwapTrade parse(CsvRow row, List<CsvRow> variableRows, TradeInfo info, Tr
...
+    return replaceLegs(trade, enhancedLegs);
+  }
+
+  // checks if the swap has a BRL-CDI floating leg
+  private static boolean hasBrlCdiFloatingLeg(SwapTrade trade) {
+    return trade.getProduct().getLegs().stream()
+        .flatMap(filtering(RateCalculationSwapLeg.class))
+        .map(RateCalculationSwapLeg::getCalculation)
+        .flatMap(filtering(OvernightRateCalculation.class))
+        .anyMatch(calc -> calc.getIndex().equals(OvernightIndices.BRL_CDI));
```
yukiiwashita: Is the FVN defined only for `OvernightAccrualMethod.OVERNIGHT_COMPOUNDED_ANNUAL_RATE`? That is, should the accrual method of the fixed leg be checked rather than the index of the floating leg?

James-OpenGamma (approved):

yukiiwashita (approved):

PR #2781, opened 2026-07-02
Enable Claude PR review workflow

OG-Stan:
## Summary
Add `.github/workflows/pr-review.yml` to enable the Claude PR review workflow, calling the shared reusable workflow `OpenGamma/og-github-actions/.github/workflows/claude-pr-review.yml@v1` with `secrets: inherit`.

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -1,0 +1,14 @@
+name: Claude PR Review
+
+on:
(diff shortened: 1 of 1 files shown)


vaibhavkrishnakumar (changes requested):
Don’t think we want this on the open source repo? It won’t/shouldn’t have access to og-github-actions I’d think

PR #2787, opened 2026-07-31
Add Convention for EURIBOR-12M Swap

yukiiwashita:


SaponeRolando (commented):
I see there is already a `EUR_FIXED_1Y_EURIBOR_12M` convention in `AdditionalFixedIborSwapConventions` class in Granite. Any reason why that one wouldn't work for our purposes?

yukiiwashita:
> I see there is already a `EUR_FIXED_1Y_EURIBOR_12M` convention in `AdditionalFixedIborSwapConventions` class in Granite. Any reason why that one wouldn't work for our purposes?

`Could not find fixed Ibor Swap convention for EUR-EURIBOR-12M` is thrown during the sensitivity calculation. I will investigate this issue.

PR #2799, opened 2026-09-22
Support step sequence variable notionals during csv writing

brianweller89:
Previous PR missed this use case; the variable notional can either be defined as a list of steps or as a sequence. Previous implementation only handled a list of steps

Verified that this change fixes the original IM round trip mismatch that causeed us to notice this problem

We need to resolve the step sequence to get the exact values, which requires `ReferenceData`. I have hardcoded the standard Strata reference data containing holiday calendars; it does not feel worthwhile to refactor the entire CSV writer infrastructure to pass custom reference data around.


beajeanm (approved):

modules/loader/src/main/java/com/opengamma/strata/loader/csv/FullSwapTradeCsvPlugin.java
```diff
@@ -1114,15 +1118,26 @@ private void writeNotionalSchedule(
...
+      Schedule accrualSchedule = leg.getAccrualSchedule().createSchedule(ReferenceData.standard());
+      List<SchedulePeriod> periods = accrualSchedule.getPeriods();
+      DoubleArray periodNotionals = notionalSchedule.resolveValues(accrualSchedule);
+
+      double currentNotional = notionalSchedule.getInitialValue();
+      for (int i = 0; i < periods.size(); i++) {
+        SchedulePeriod schedulePeriod = periods.get(i);
+        double periodNotional = periodNotionals.get(i);
+
+        // if notional has not changed then no need to include as a step
```
vaibhavkrishnakumar, line 1134: Does this preserve across round trip? E.g. if there was a step without a notional change read in, that will be lost here and merged with the step prior, is that fine?

vaibhavkrishnakumar (approved):

PR #2800, opened 2026-09-22
PROD-57270: Result.ofOptional lazy create the failure

beajeanm:
Creating a failure is expensive since we need to create a stackTrace for
it. Result.ofOptional currently always allocate a failure even if the
option is not empty.
Switch to lazily creating the failure instead


Mistuus (approved):
nice

